PR #8, opened 2015-09-12
Scons: Improved Python2 detection on Windows.

AlinGlow:
Now searches in HKEY_CURRENT_USER registry branch as well.
Caused by "Install for current user only" checkbox in windows installer.


kayhayen:
That's great. I never knew what installing for user or everybody meant. That will help some people a lot.

I am going to merge this into the next hotfix, but merge for now here.

PR #18, opened 2017-05-17
Tests: Fix sys.version comparisons

jwilk:
`sys.version` is a string, so it doesn't make sense to compare it with a tuple.

Found using [pydiatra](https://github.com/jwilk/pydiatra).


kayhayen:
Hello there,

is this all pydiatra found in Nuitka? Thanks for reporting.

PR #36, opened 2017-11-05
Move the main script into the module

andrewleech:
This allows running nuitka directly from the module, eg `python -m nuitka`
It also makes it much easier to run from other python scripts as they don't need to search for the platform/virtualenv script folder.

The everyday runnable executables in script / bin are now created automatically with setuptools entry-points.

This simplifies setup.py and leverages built in mechanism of setuptools the create executable in the correct location on all supported platforms.


kayhayen:
I really like the idea of having the target Python version provided by calling it, instead of Nuitka messing around with it. So "python2.7 -m nuitka" is much better than "python2.7 `which nuitka`" or "nuitka --python-version=2.7", by far. So this is very welcome.

However, be aware that Nuitka want's to re-execute the Python that it ran as, under a couple of conditions, one being random hashing being enabled. Since this makes dictionary ordering change, Nuitka hates that, and the "site" module must not be read, as it patches and changes the interpreter in ways that make it hard to tell what it did.

The tests may fail due to that being needed. I didn't look at your Travis run. I noticed you do. Feel free to ask me questions, I will however try and not take away from this. Looks like a very nice contribution to be made for usability of Nuitka you are doing there. I am totally in favor of the runners to become minimal or even automatically generated, maybe the batch scripts on Windows could die too then, which would be good too.

My only concern there would be start up time, which I would love to keep as low as possible. But maybe that can be addressed with patching up the resulting script afterwards.

andrewleech:
I'll try to follow through some tests of it re-executing itself, hopefully there wont be any major issues.

The auto-generated scripts/executables are generally quite small and yes, they will definitely remove the need for the batch files. I had meant to delete them already.

That being said, the launcher scripts use pkg_resources to query the package definitions to find the entrypoint (by name) as defined in setup.py, then runs it. There is certainly going to be some overhead here and while I've never noticed it in practice, I haven't profiled it to check how long this takes. 

Personally, in other projects, the convenience and zero-maintenance of the entry points has always won for me, but I agree this needs to be balanced with any performance penalties.

The unit tests failing on travis is a really strange one. All the tests pass for me on my local pc's but on travis I'm getting failures to import builtins like `runpy` and even `threading`. 

I though this might be bug on travis itself because the python files for these modules appear to simply not exist on travis filesystem. I'm following it up with them here https://github.com/travis-ci/travis-ci/issues/8709
I see however that other builds from develop and master has passed since then so it probably is related to my code for some reason. Not sure why it all works fine locally but not in travis still.

andrewleech:
Ok, so after a bit more reading of the code, the travis failure is clearly going to be due to the not-importing-site. 

I'm guessing that on travis python installations have a decent bunch of their libraries hiding in a zip or similar that needs its .pth file read to use.

Could you elaborate a little more on why site package needs to be excluded? 

Would it be enough to just exclude user-site instead (-s instead of -S)
Or is it only needed in some situations and the exclusion could be disabled with an argument for times where it's not needed?

That being said, as this site exclusion is only when running nuitka command line app, I can certainly change my launcher scripts as needed to not use any of the modules not working on travis, although that won't guarantee that there might be other times in the future this fails for other users with different things needing pth. For example I know the official Windows embeddable zip file distributions have everything in a zip file so these definitely wont work.... however I also doubt many people will try to use these for running nuitka itself.

kayhayen:
Hello Andrew,

indeed, not finding the "site-packages" will be causes by not having "site" module parsed, and many other bad things. Also the ".pth" file scan, which executes arbitrary code, and e.g. creates pseudo built-in modules like "gtk" in every Python process on your system is an issue.

Nuitka want's to run in a pristine environment not changed by "site.py", so it knows that "help" is not really a built-in function. It also want's to disable random hashing, or else code generation will not be stable, and defeat caching with ccache and comparison tests.

On the other hand, some virtualenv won't load practically anything without the site module. I recall that even "pickle" had issues for some setups.

The way this is handled is that Nuitka once starts with the site module enable, then preserves its effects on "sys.path" and namespace packages in environment variables, to make use of it, when it starts again without site module. A loader would always have to check this:

```
if "NUITKA_PYTHONPATH" in os.environ:
    # Restore the PYTHONPATH gained from the site module, that we chose not
    # to have imported. pylint: disable=eval-used
    sys.path = eval(os.environ["NUITKA_PYTHONPATH"])
    del os.environ["NUITKA_PYTHONPATH"]

```

This will be needed for "import nuitka" to work. Now that I see it, I kind of think, using eval there is not too safe, but that's another issue. Simply setting "PYTHONPATH" wouldn't be good, as the "--run" passes this on, as well as changing the Python version is harmed by that.

Injecting that part into the loader might be necessary for things to work. In setup.py you will find a hack for easy install and the "develop" target, which kind of needed this too. Don't use that generally though, easy install is breaking python 3.6 at least. But I am optimistic, hacking "NuitkaInstallScripts" to do those kinds of changes wouldn't be too bad.

BTW: One solution I was thinking of, was to make all "computations" in a slave process. So that e.g. the list of built-in names, etc. and even "7+2" is not done in the main process, but in that slave process, and that wouldn't ever "import nuitka" but merely provide a way to make a computation.  For cross compilation, that could really be the way to have "ctypes" work on the native platform and be able to tell us things, or "sys.maxint", that kind of stuff. But that kind of is an idea that is going to be a lot of work, not maybe never happening.

I hope this explains things a bit. Let me know where I was unclear. I will also try and extract into the Developer Manual after we finish the issue.

andrewleech:
Hi @kayhayen,
I believe my changes are stable and working well since the last update, I've been using it successfully in a few configurations.

Would it be possible to get this considered for the next release?

kayhayen:
Yes, going to include it for next release

andrewleech:
Great thanks. My other bdist_nuitka branch is hopefully quite close too, I should be doing some more in depth testing tomorrow.

kayhayen:
Nice, I was trying to merge, but the merge conflicts confused me deeply :)

andrewleech:
No worries, I already had them merged together locally, just the other way around ;-)

kayhayen:
I noticed a few instances of replacing "bin", "nuitka" in quality tools being wrong, there they just referred on a list of directories to work on. I am going to amend that away I think. I hope you don't mind me doing that.

Another thing is the discussed runners situation. In "bin" there remain a few runners, which do the hack of inserting "sys.path" before importing the "__main__" from nuitka.tools sub packages.

I kind of wish those runners back in the git checkout, as they allow me by PATH to pick the Nuitka, without messing around with PYTHONPATH. I will simply add them back if you do not object, and then "setup.py install" ignores those.

Then we have 2 ways to run nuitka, from PYTHONPATH, from PATH.

Also, I think the "LIBDIR" tricks were moved to "nuitka.__main__" but nothing in setup.py works with anymore, you made that obsolete, which is great, and now adding the "nuitka" parent package directory, no more has a point, as it was imported already. I think we can discard that too, going to do that.

Ah, and of course, I am going to add you to the contributors list. Fantastic contribution of yours, I am really excited about this structural improvement. Andrew. BTW: Your commits use 2 email addresses, are you aware of that?

Yours,
Kay

andrewleech:
Thanks for picking up mistakes in the renaming, I clearly didn't look at them all closely enough. 

Absolutely go ahead and change/amend any commits as suits, very happy to take any suggestion to meet any needs. 

And yes I know there's a couple of emails... some of the changes have been done at work as I'm using nuitka in combination with briefcase on a project, the rest has been done at home.

kayhayen:
I was looking further into the "__main__" things, and I noticed weird stuff and had realizations.

First of all, it was adding the directory of the "__main__" to the "sys.path", which adds a competing set of modules there, "distutils", "utils", all of that bound to conflict with programs. But it's easy enough to remove that with a filtering list contraction.

Second, I noticed that on re-execution, the "nuitka" package was not found anymore if the CPython version is changed, because of course, sys.path is not preserved in that case. So in that "-m nuitka" mode, we should disable that ability at least. And while my attempts at repairing it have succeeded, but I think we ought to remove it entirely. It only adds complexity, and makes people think that Nuitka needs such an option, which it clearly does not. I will therefore remove its visibility (help output presence) in one of the next releases, once I am sure it's not needed anymore.

For the runners, I was lost in "RuntimeError: Bad magic number in .pyc file", which strangely only happens if I come back via "__file__", but not if I come back via 'sys.modules["__main__"].__file__' not sure what is going on there. Maybe because I was adding "nuitka" as a "ModuleType()" to not have to add its parent directory to "sys.path" at all, which is generally not healthy for changed Python versions, but that seems to not work then.

Another topic is packaging. So far, Nuitka tucks away its "nuitka" package code not in "dist-packages" (elsewhere called "site-packages") on Debian, as it is nothing anybody else should see or use. That breaks "-m nuitka" of course and thus "bdist_nuitka" which currently works that way only. And Debian packages do not have any other way to switch Python version, so making Python3 and Python2 packages for it, kind of would be logical and like everything else does it, which feels right.

Re-execution with another Python version would however have to find the "nuitka", and then add the 2.7 main package path to say 3.5, and that's not healthy.

The real solution for this, is to stop doing that at all. No changing of the Python version at all anymore, which is ultimately the benefit of your work. Instead make people manually provide the Python version through dedicated runners (somebody made a "nuitka3" for RPM packages) and/or calling the final Python binary via "-m".

For now, the goal will be to get it out, which seems doable now. It works for me. I am going to push the necessary corrections later today, with TODOs added for me. On the packaging side, this is all going to be a lot of work, but a huge step towards normalizing nuitka.

andrewleech:
You make some great points. I certainly never tested the cross-version functionality as I've no personal need for it.

I know people who've preferred the approach of installing a utility once (system wide or in it's own venv) which can then be used for any python install; many people use `devpi` this way. 
Personally, with pip and pypi working so well these days I've always found it easier to install a utility/module in the virtualenv that's going to use it. It certainly potentially simplifies the nuitka code to not have to worry about which python environment it needs to work in.

Alternate system-specific packaging is another nightmare I've stayed away from. 
I have a personal interest in simplified, automated packaging so I can release changes with little to no work (here's another of my typical changes for this: https://github.com/andrewleech/Nuitka/commit/8791601576f2e2c7aa2a71e827d434bb35ceb7e0)
Maintaining alternate packaging schemes for debian/redhat/arch/etc is the opposite of hands-off. Again, I like pip and pypi specifically because it's cross platform and only needs to be done once. 
That being said a colleague who loves debian and deb packages has made a point that pip installed things can conflict with apt installed modules and virtualenv's are pretty unstructured and hands-on for end users to install new apps. `briefcase` is a package I'm using a lot these days to get closer to removing this issue (hasn't completely solved linux yet though).

PR #38, opened 2017-11-28
Updates to bdist_nuitka

andrewleech:
I'm really excited about this usage of nuitka, it provides a very clean and controlled output being able to produce fully compiled wheels ready to distribute and install.

I'd actually started making my own module to integrate nuitka into setuptools before I found what you already had built in.

I've been testing/using this on windows, however this PR doesn't work as is on windows due to `getExecutablePath("nuitka")` failing in my virtualenv. I've been using this branch rebased on #36 where it all works perfectly.

Contents of PR:

* My initial tests of bdist_nuitka resulted in broken wheels with incorrect paths inside the wheel package.
For example, on my pc the package I'm compiling is `C:\Users\anl\src\example_package\setup.py`
When I built the wheel and then open is as a zip there was a deep folder structure inside it like 
`zip:\Users\anl\src\example_package\.venv\lib\site-packages\example_package.pyd`.  
I found all the stuff starting at line 96 in the install phase was unnecessary once the wheel class was pre-informed of it being a binary module and `self._buildPackage` was called with an absolute path.

* I regularly find issues compiling packages with python files not being included due to them not being explicitly imported by other files in the package. 
The changes starting with `recurse_packages` is designed to get around this by compiling and including all python files in the wheel. The actual nuitka compiling is done against the files that have been staged for building the wheel so the normal methods of setuptools to include the correct files in the wheel are observed.
This doesn't yet do anything to work around the issues with other resources/files that would have been included in the wheel, these are still deleted along with the source. I'd like to resolve this but that will be a follow up pr.

* The final change in `setup.py` registers the `bdist_nuitka` command when nuitka is installed so you no longer need to explicitly reference the module when running.
old: `python setup.py --command-packages=nuitka.distutils clean -a bdist_nuitka`
new: `python setup.py bdist_nuitka`

### DIFF
--- a/nuitka/distutils/bdist_nuitka.py
+++ b/nuitka/distutils/bdist_nuitka.py
@@ -31,6 +31,15 @@ import wheel.bdist_wheel  # @UnresolvedImport
 # pylint: disable=C0103
 class bdist_nuitka(wheel.bdist_wheel.bdist_wheel):
 
+    def finalize_options(self):
+        # self.distribution.libraries = [[True]]
+        self.distribution.has_ext_modules = lambda: True
+
+        super(bdist_nuitka, self).finalize_options()
+        # Force module to be recognised as binary rather than pure python
+        self.root_is_pure = False
+        self.plat_name_supplied = self.plat_name is not None
+
     # Our wheel base class is very unpure, so are we
     # pylint: disable=attribute-defined-outside-init
     def run(self):
@@ -47,12 +56,11 @@ class bdist_nuitka(wheel.bdist_wheel.bdist_wheel):
         # Nuitka wants the main package by filename, probably we should stop
         # needing that.
         from nuitka.importing.Importing import findModule, setMainScriptDirectory
-        from nuitka.utils.Execution import getExecutablePath
 
         old_dir = os.getcwd()
         os.chdir(build_lib)
 
-        # Search in the build directory preferrably.
+        # Search in the build directory preferably.
         setMainScriptDirectory('.')
 
         package, main_filename, finding = findModule(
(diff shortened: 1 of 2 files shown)


kayhayen:
Great stuff. Regarding the recursion, I was going to add a --include-module/package option. Right now the --recurse options only allows to go there, but not force it, in fact only recurse-directory with --recurse-to combined does, which requires to find that directory manually. However I have not done anything to that end yet for real. Depending on how complex your package is, you might not find that necessary.

The others things read great. Going to look at this before next release definitely.

kayhayen:
Having looked at it, basically what you do, is you try to make it include every Python file that was copied there by distutils, rather than my idea of relying on the "PACKAGES" selector from the setup arguments, which as I pointed out, is not yet even implemented.

Having "distutils" do the heavy lifting seems very fit. I think we ought to have a better command line option to tell Nuitka those things, or else it will get very long.

I am going to try it out later this week. Thanks a lot and let me know when you think it is ready.

andrewleech:
I just added some fairly major changes to the bdist_nuitka handing.

`python setup.py bdist_nuitka` still creates a compiled wheel the same as previously, however now if there are any other files in the wheel (project assets/images/libraries) these are kept in the existing folder structure and the raw python files deleted. You get the compiled pyd at the top level and the original folder sitting next to it, all in the wheel.

I've also added a new setup.py option to use nuitka for all builds; `build_with_nuitka=True`
For eample, in your package you want compiled with nuitka, you add this to setup.py:
```
from setuptools import setup
setup(
    name='example',
    packages=['example'],
    description='test package',
    setup_requires=['nuitka'],
    build_with_nuitka=True
)
```
With this setting, any setuptools operation which should build your package will run it though nuitka, such as `python setup.py bdist_wheel`, `python setup.py install` or `pip install ./example`

ps. This allows easy integration with briefcase (https://github.com/pybee/briefcase) to allow your python application to be compiled by nuitka and packaged into a distributable installer/application all in one operation.

kayhayen:
Merging this for coming pre-release, expect rebases potentially

PR #42, opened 2017-12-17
Fix compiling with provided GCC through CC env var

pauliuszaleckas:
If system had too old (default) GCC found_gcc is set to False,
so if another compiler was provided with CC Nuitka used g++ instead.


kayhayen:
Thanks a lot. I checked that indeed, this is only done for "CC" and not for "CXX", which as a feature might even be missing. Not sure if it's worth the effort though.

PR #48, opened 2018-01-08
Faster DLL-Search

LarsK1:
Added faster DLL (or equivalent) search through parallelisation


LarsK1:
It will only work for **Python 3+**. Do you have an idea how to **fix** that ?

LarsK1:
@kayhayen
Is it so Okay? What do you think about?

kayhayen:
I was wondering if you are aware that Nuitka started caching these things, the output. For most users now, "depends.exe" should rarely run more than once. This should already be in develop branch. Do you think it still makes sense to optimise that first execution?

LarsK1:
Yes, it makes sense, because at my latest build it took about **20 hours** to find the dependencies. That's definitely too much!

LarsK1:
And so it took only 15 Minutes!

kayhayen:
So, that is quiet shocking to me. The 20hours, are you doing something very unusual? How many calls are that?

Obviously I will strongly consider to merge it. I just lacked the time to look at in enough.

Yours,
Kay

LarsK1:
I'm only programming somy usual with numpy and scipy (98 calls) for analysing some audio!

Lars

LarsK1:
It means you will merge it?

kayhayen:
Yes, I definitely will consider it. I was just going to check on my own if it works and didn't find the time yet.

xoviat:
The DLL analysis can be improved on Windows by adding a dependency on the `bindepend` package that I have published to pypi. It is not possible to add the code to Nuitka because it was originally released under the GPL. What's frustrating is that I know of three variants of this functionality across the Python ecosystem, which is unnecessarily duplicative IMO.

AnonymousDeveloper4201337:
I really hope this gets merged sometime :D anyway to make compilation faster sounds great to me :D

LarsK1:
@kayhayen So did you test it? Worked it well?

kayhayen:
How sure you are you that the long running you have seen is not caused by an infinite recursion. There is a fix for Linux and that in the new release. Maybe all it takes is to fix that. I have not merged it, because I would like to witness the slowness myself first.

Sorry for changing my mind.

LarsK1:
I'm 100% sure, because I timed it using a Windows PC, and there was no bug, or am i wrong?

kayhayen:
Hello Lars,

is there any way I could see it too? :-)

And yes, for Linux a endless recursion bug got fixed. Now I wonder if your scheduling code is not a really more complex than necessary way of solving that same issue for Windows. Because "depends.exe" really should be IO bound or called a lot less often than we currently do, because it recurses itself into things. I would want to look at these things and you seem to have the reproducer that makes it worthwhile.

Yours,
Kay

LarsK1:
Shure, you could test it on a Windows PC, too. If you havent such a PC, I could send you Screenshot's with a timer in front of it.
Lars
P.S. My PC:
  processor: Intel i7-6700
  ram: 32GB
  So it should be enough?!

kayhayen:
I was more thinking like example code that I compile myself that shows the slowness you encounter.

LarsK1:
OK,
try this one:
import numpy, scipy
numpy.Array()
scipy.signal.cubic(x)

LarsK1:
And replace x through any number

kayhayen:
Hello Lars,

this is actually not that bad:

Nuitka:INFO:Running depends.exe for Mini.dist\_bsddb.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_ctypes.pyd took 0.01 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_elementtree.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_hashlib.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_msi.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_multiprocessing.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_socket.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_sqlite3.pyd took 0.01 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_ssl.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_tkinter.pyd took 0.01 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\bz2.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\core\multiarray.pyd took 21.53 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\core\umath.pyd took 19.70 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\fft\fftpack_lite.pyd took 19.76 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\linalg\_umath_linalg.pyd took 19.28 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\linalg\lapack_lite.pyd took 20.25 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\random\mtrand.pyd took 20.81 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\pyexpat.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\scipy\_lib\_ccallback_c.pyd took 19.81 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\select.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\unicodedata.pyd took 0.00 seconds

The "0" numbers are from caching. The non-namespaced ones are cached from previous compilations of other things, the numpy and scipy ones are new, and relatively fast. I think they got faster, since I am using an SSD for Windows now.

I think your test case is not showing the issue. Maybe some more imports of scipy will trigger it?

Yours,
Kay

LarsK1:
What the hell, what kind of PC do you have? For me one dll took around 6 Minutes!
Ok, I send you the code later

kayhayen:
On a rotating disk it took me 1 minute worst case (single depends.exe call) for relatively normal compilations That convinced me that caching is worth while. However, there are only so many extension modules, next time it's automatically at 0 too.

If you say 6 minutes, and assuming your rotating disk is not 6 times slower, maybe CPU is also a factor indeed and we should compare numbers too. This is an 8 core 4Ghz AMD machine, not the Ryzen kind, but the generation before that.

kayhayen:
Look at this, doing it once, I take no time anymore:

Nuitka:INFO:Running depends.exe for Mini.dist\Mini.exe took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_bsddb.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_ctypes.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_elementtree.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_hashlib.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_msi.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_multiprocessing.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_socket.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_sqlite3.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_ssl.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\_tkinter.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\bz2.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\core\multiarray.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\core\umath.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\fft\fftpack_lite.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\linalg\_umath_linalg.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\linalg\lapack_lite.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\numpy\random\mtrand.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\pyexpat.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\scipy\_lib\_ccallback_c.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\select.pyd took 0.00 seconds
Nuitka:INFO:Running depends.exe for Mini.dist\unicodedata.pyd took 0.00 seconds

Which makes me wonder, if doing something to be done only once, is that worth it to be done with threads, if it doesn't take that long compared to the C compilation parts.

LarsK1:
Hey,
these are the real files.
[HueBeat.zip](https://github.com/kayhayen/Nuitka/files/1880037/HueBeat.zip)
Lars

My PC have 8 cores @ 3,4 GHz with 32GB RAm and an PCLe-SSD (System)!?

kayhayen:
More recently I have encountered the issue that the caching that I see as the solution (only once, why care) may in fact not be reliable. Things added later, things changed, they won't be noticed.

That is why I am considering the chance my stance and will re-consider this, or change the feature entirely so that e.g. users build a dependency package once, then re-use it, but still it could be done with a change like this.

I do not get why dependency walker is so slow. Maybe it is doing things, like profiling DLL load, that I do not think we need. The bindepend alternative seems stalled and unused at a 0.1 release, and not able to carry the load.

So don't be surprised if I merge this still. I kind of feel you must think I am stubborn, stupid, or both. And you might be right. ;-)

kayhayen:
Looking at the modest diff, I think one can merge this. Also there is a backport for at least Python 2.7 available, so concurrent.futures could be in-line copied. I would make a minimal stub in "nuitka.__past__" that avoids the need for version checking. I think I am not going to immediately go into adding an inline copy of that package, but we could consider it surely. I think Python 2.6 would have to get a single thread immediate worker stub in any case, so making it now, and only accelerating Python 3 seems OK.

For that, a "try/except ImportError:" would be good, to see if it is installed. We could suggest users to install it with appropriate pip commands instead of making yet another inline copy. Not sure how big it is, but it doesn't feel small.

Maybe I am not putting it in "nuitka.__past__" until anything else is using it, but in some utils module, so its import bloat doesn't strike on Linux too.

One observation, apparently we have been using an OrderedDict() to get a stable order. That seems out of the window naturally when workers complete in arbitrary order. Need to check if order is still available. The lists in that ordered dict will have become any order. I think some kind of sorting will have to be done in the end before returning.

All in all, this looks trivial to do. I will be working on this now. I am going to make a single commit out of your many. I was mistaken in thinking this was much more invasive than it actually turns out to be.

Thanks for your contribution. I expect the next pre-release to contain your work, I pushed it to factory just now, with a few minor adaptations (python_version vs. version_info, getCoreCount() for CPU taking, etc.). I will close it therefore now.

kayhayen:
I pushed the stubbing change to factory too, hopefully not breaking things too much.

I noticed the 3*cpu count in the code, and was thinking of job limit parameter. Probably there should be a limit adhered here. In my testing, I will check if dependency walker is CPU intensive. I think it might not be, bit overloading a machine must be possible to prevent. Job limit right now is "cpu count", which might be better at "cpu count +1". But definitely wouldn't be lower than 3 times it. Did you observe speed ups beyond cpu count? I imagine you did?

LarsK1:
@kayhayen yes I tested the code heavyly. And I noticed a great speed-up, also in the cpu usage. But your rigth i have only tested it with python 3.6.4. And I had already dowloaded depends.exe. So your rigth maybe it would be great not to multit**hread** it, rather than use multi**processing**, so that we could use the whole cpu with all it cores and also the whole IO capabilities!
Lars

PR #57, opened 2018-02-19
[WIP] Update multiprocessing plugin for Python 3.4

nmoehrle:
The multiprocessing module was subject to a larger refactoring in Python 3.4 (python/cpython@84ed9a68) -- the `multiprocessing.forking` module was split up into multiple modules `reduction, spawn, popen_fork*, popen_spawn_*` which broke the multiprocessing plugin. This PR is an attempt to fix this but is work in progress and should serve as an discussion platform.


kayhayen:
Hello,

are you still intend on working on this? This would be really good to see finished.

Yours,
Kay

kayhayen:
I cherry-picked your commits on to factory, and it compiled a multiprocessing test program fine, do you think it's still WIP, and why?

kayhayen:
Seems that multiprocessing does work on Linux and alike, but not on Windows and at least Python3, even before 3.4, as 3.3 also failed for me.

Are you intent on working for Windows still? Otherwise this seems good and I will merge at least this part, rebasing your branch ought to be trivial. Give a sign of life please. :-)

kayhayen:
Ok, stupid me, I forgot that in fact the plugin is only even used on Windows, so nothing new is really working, although I think Python 3.4 on Linux might have started to work now, but that can't be real.

kayhayen:
So I figured out the missing bits. The multiprocessing plugin so far only worked if you also provided --standalone or --recurse-to=multiprocessing, so it could monkey patch things while it loads them. That wasn't the case with my test program. And it shouldn't have to be. So now it works out of the box, if you enable the plugin, and doesn't hang anymore.

I think we can close this. Let me know if I am wrong.

PR #58, opened 2018-02-20
Remove obsolete nuitka binary search

andrewleech:
I found a merge-fail left over from my previous couple of merge requests, #38 and #36. 
Currently bdist_nuitka fails as it's searching for the binary which was removed in #36

This PR removed the check which is already no longer used.

### DIFF
--- a/nuitka/distutils/bdist_nuitka.py
+++ b/nuitka/distutils/bdist_nuitka.py
@@ -97,10 +97,6 @@ class build(distutils.command.build.build):
 
         output_dir = build_lib
 
-        nuitka_binary = getExecutablePath("nuitka")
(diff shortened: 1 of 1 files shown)


kayhayen:
I recall stumbling across this myself. Not sure if I ever did something about it, but I thought I did. Looks good to me though.

andrewleech:
Similar to other recent PR's, I don't think the travis failure is related to my change:
https://github.com/kayhayen/Nuitka/pull/59#issuecomment-368077042

Do you consider this change ready to accept?

kayhayen:
As usual, I am going to rebase the merge commit away later. Thanks for providing this. I am getting back to work on Nuitka with more time again.

PR #63, opened 2018-03-14
bdist_nuitka: Allow passing extra nuitka options from setup.py

andrewleech:
Adds the ability to pass any extra options to nuitka when used in bdist_nuitka.

Eg:
`setup.py`
```
from setuptools import setup

setup(
    name='mypackage',
    packages=['mypackage']
    options={
        'nuitka': {'recurse-not-to': ['mypackage.ui.assays'],
                   'plugin-enable': ['qt-plugins'],
                   'no-pyi-file': None}
    },
    build_with_nuitka=True
)
```

TODO: Documentation update is still required

### DIFF
--- a/nuitka/distutils/bdist_nuitka.py
+++ b/nuitka/distutils/bdist_nuitka.py
@@ -25,6 +25,8 @@ import os
 import shutil
 import subprocess
 import sys
+import collections
+from nuitka.__past__ import unicode
 
 import wheel.bdist_wheel  # @UnresolvedImport pylint: disable=I0021,import-error,no-name-in-module
 
@@ -110,6 +112,7 @@ class build(distutils.command.build.build):
             "--output-dir=%s" % output_dir,
             "--recurse-dir=%s" % self.main_package,
             "--recurse-to=%s" % self.main_package,
+            "--recurse-not-to=*.tests",
             "--show-modules",
             "--remove-output"
         ]
@@ -121,7 +124,8 @@ class build(distutils.command.build.build):
                 source, value = details
                 if value is None:
                     command.append(option)
-                elif hasattr(value, '__iter__'):
+                elif isinstance(value, collections.Iterable) and \
+                        not isinstance(value, (unicode, bytes, str)):
                     for val in value:
                         command.append('%s=%s' % (option, val))
                 else:
--- a/nuitka/distutils/bdist_nuitka.py
+++ b/nuitka/distutils/bdist_nuitka.py
@@ -110,14 +110,27 @@ class build(distutils.command.build.build):
             "--output-dir=%s" % output_dir,
             "--recurse-dir=%s" % self.main_package,
             "--recurse-to=%s" % self.main_package,
-            "--recurse-not-to=*.tests",
             "--show-modules",
-            "--remove-output",
-            main_filename,
+            "--remove-output"
         ]
 
+        # Process any extra options from setuptools
+        if 'nuitka' in self.distribution.command_options:
+            for option, details in self.distribution.command_options['nuitka'].items():
+                option = "--" + option.lstrip('-')
+                source, value = details
+                if value is None:
+                    command.append(option)
+                elif hasattr(value, '__iter__'):
+                    for val in value:
+                        command.append('%s=%s' % (option, val))
+                else:
+                    command.append('%s=%s' % (option, value))
+
+        command.append(main_filename)
+
         subprocess.check_call(
-            command
+            command, cwd=build_lib
         )
         os.chdir(old_dir)
 


kayhayen:
So I assume this is generic, so key, value -> --key=value unless value is None. Maybe it is worth, to reconsider how to handle booleans. Having --pyi-file and --no-pyi-file both in nuitka/Options and them working together, could allow to handle booleans like "py-file" : True in a generic way. For False we would prefix the name with "no" simply.

The one thing I really miss most about this is tests though. Like a hello world setuptools program and module, and we show the use cases. It will take creating a virtualenv, running setup.py with and without build_with_nuitka=True and having a minimal thing work, and a show off thing work. Like something that inject an option to the nuitka key in setup.py and then shows it has intended effect or not.

I am willing to work on adding that myself though, shouldn't be all that much effort to get that started, then once I am happy with the interface, I think we should document it.

nuitka/distutils/bdist_nuitka.py
```diff
@@ -105,14 +105,27 @@ def _buildPackage(self, build_lib):
...
         ]
 
+        # Process any extra options from setuptools
+        if 'nuitka' in self.distribution.command_options:
+            for option, details in self.distribution.command_options['nuitka'].items():
+                option = "--" + option.lstrip('-')
+                source, value = details
+                if value is None:
+                    command.append(option)
+                elif hasattr(value, '__iter__'):
```
kayhayen: Not sure if this ought to be: isinstance(value, collections.Iterable) and we need a string check first, there is something in nuitka.__past__  for that, as "basestring" is not available in Python3. Otherwise strings are iterables, and will fall into this.

nuitka/distutils/bdist_nuitka.py
```diff
@@ -105,14 +105,27 @@ def _buildPackage(self, build_lib):
...
+                    for val in value:
+                        command.append('%s=%s' % (option, val))
+                else:
+                    command.append('%s=%s' % (option, value))
+
+        command.append(main_filename)
+
         subprocess.check_call(
-            command
+            command, cwd=build_lib
```
kayhayen: This seems new, probably a fix of its own right?

andrewleech: Oh that's probably superfluous as there is a os.chdirearlier in the file and a restore just after this, however if the other things in this function can be fixed to not need the os.chdir this is probably safer (no chance of other threads os.chdir out from under you.... if there was ever other threads at play)

nuitka/distutils/bdist_nuitka.py
```diff
@@ -105,14 +105,27 @@ def _buildPackage(self, build_lib):
             "--output-dir=%s" % output_dir,
             "--recurse-dir=%s" % self.main_package,
             "--recurse-to=%s" % self.main_package,
-            "--recurse-not-to=*.tests",
```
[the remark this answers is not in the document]
andrewleech: Yeah you're probably right, I don't love having defaults that you can't override, but then I can't really think of a use case for keeping anything starting with test in a compiled piece of work

andrewleech:
Yes I'm pretty sure you're understanding the logic the way it's intended, key=val or boolean.

I'm not sure there's a problem with how booleans are handled already, either they're added for true or omitted for false. Having to write `None` in the options block to signify a (true) boolean value is a bit ugly though, I'd be open to a more positive readable format to use. 
Perhaps simply allowing the arg to be listed as True (instead of None) and still treat as boolean (as opposed to a string  to to treat as key/val). In this case if you set it to False the arg would be omitted.

You're absolutely right about the lack of tests.... sorry I'm terrible at rushing through functionality and not making the effort on tests if I'm not forced to.  I really do believe in the value of them, just not in the best of habits.
Stupid me often demonstrates the function working so don't get allowed the time to go and clean up / add tests.... again, bad habits.

Everything I've contributed recently has been like this, I should have taken this as the warning it was: https://github.com/kayhayen/Nuitka/pull/58

I should be able to make a first pass at a test today, perhaps we can build up coverage from there.

kayhayen:
Hello Andrew,

can you rebase this, it is in conflict with other changes I made.

Yours,
Kay

nuitka/distutils/bdist_nuitka.py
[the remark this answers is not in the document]
andrewleech: Would you prefer this brought back in before merging?

nuitka/distutils/bdist_nuitka.py
andrewleech: I haven't fixed this yet either sorry

kayhayen:
I don't know where now, not too easy to  oversee, but you had asked if I wanted the other changes as well, I am fine with rebasing only.

andrewleech:
Oh, I made the changes, should be all fixed and rebased.

andrewleech:
Seems I missed the release :-( oh well, I've rebased again (cherry picked)

kayhayen:
I am about to merge this, putting it through my tests. Hope you are OK if I squash the 2 commits of yours for noise reduction keeping your authorship intact though.

I will use this ones description to make another distutils test that indeed uses a plugin. The qt-plugin is a bad one, as it's not for compiled modules yet. I will try to find something more light weight, but actually I don't know what yet.

andrewleech:
That sounds great thanks.

Yeah sorry I didn't figure out the test system enough to add a suitable unit test for any of these bdist features. If you can get something in place I'll certainly aim to add coverage to it for any related features.

kayhayen:
If I am not mistaken, I actually merged this a while ago, sorry to keep it open for no reason.

No docs, no test though, I am lazy too.

andrewleech:
Yes , I did see it merged thanks. 
I've moved my ci away from git install to released version, working great!

kayhayen:
Do you happen to be a tox user?

andrewleech:
I'm actually not. As much as I love the idea the couple of times I tried it I ran into difficulties fitting it into my CI workflow. I use gitlab-ci for everything, with different python versions running in parallel jobs. Eg. https://gitlab.com/alelec/hello_briefcase/blob/master/.gitlab-ci.yml

PR #82, opened 2018-05-04
WIP: Don't link against dynamic python lib for modules on linux/mac

### LINKED ISSUE #81: Building in manylinux container
Hi, 
I'm trying `bdist_nuitka` inside a manylinux container to create linux binary wheels for distribution:
https://github.com/pypa/manylinux

Have you ever tried this at all?

I get failures at the final linking step (with debug enabled), 
```
g++ -o mypkg.so -export-dynamic -Wl,-R,'/opt/python/cp35-cp35m/lib' -Wl,--disable-new-dtags 
-Wl,-b -Wl,binary -Wl,/build/build/lib/mypkg.build/__constants.bin -Wl,-b -Wl,elf64-x86-64 
-Wl,-defsym -Wl,constant_bin=_binary__build_build_lib_mypkg_build___constants_bin_start 
-shared mypkg.build/module.mypkg.common.finder.os mypkg.build/module.mypkg.helpers.os 
<<SNIP FILE LIST>>
mypkg.build/fibers_x64.os mypkg.build/swapfiber.os -L/opt/python/cp35-cp35m/lib 
-lpython3.5m -ldl -lpthread -lutil -lm

/opt/rh/devtoolset-2/root/usr/libexec/gcc/x86_64-CentOS-linux/4.8.2/ld: cannot find -lpython3.5m
```
Because the dynamic library is not included in the environment (by design?)
https://github.com/pypa/manylinux/issues/69
https://github.com/pypa/manylinux/issues/85

> I believe the correct fix is for the OpenCV build process to allow the Python symbols to be undefined at link time

http://vtk.1045678.n5.nabble.com/Fwd-OSX-build-links-directly-with-Python-binary-avoidable-td5731734.html

https://gitlab.kitware.com/vtk/vtk/merge_requests/1511

To be honest I'm a little lost with regard to how it could possibly build without linking against the library, but I gather it would not want to be locked into the library in the build environment, so that it can use the end-user's library instead?

I've also failed to find where scons gets its reference to `-lpython3.5m` from.  I've tried manually deleting it from both python3-config and the related pkg-config files and scons still finds it.

Considering the manylinux environment is specifically designed for building compatible binary wheels, shouldn't this be possible, even recommended?

andrewleech:
Avoiding the library improves compatibility on other systems (and is required to build in manylinux container)

Closes #81

This has not yet been tested on an independent system.

### DIFF
--- a/nuitka/build/SingleExe.scons
+++ b/nuitka/build/SingleExe.scons
@@ -1133,13 +1133,14 @@ if win_target:
 
 else:
     # Debian and Ubuntu distinguish the system libraries like this.
-    if python_debug and \
-       python_prefix == "/usr" and \
-       not python_version.startswith('3') and \
-       platform.dist()[0].lower() in ("debian", "ubuntu"):
-        env.Append(LIBS = ["python" + python_abi_version + "_d"])
-    else:
-        env.Append(LIBS = ["python" + python_abi_version])
+    if not module_mode:
+        if python_debug and \
+           python_prefix == "/usr" and \
+           not python_version.startswith('3') and \
+           platform.dist()[0].lower() in ("debian", "ubuntu"):
+            env.Append(LIBS = ["python" + python_abi_version + "_d"])
+        else:
+            env.Append(LIBS = ["python" + python_abi_version])
 
     if python_prefix != "/usr" and "linux" in sys.platform:
         env.Append(


kayhayen:
I am closing this, as it solved differently, coming in the next pre-release.

PR #101, opened 2018-08-22
Use feature detection to reduce PyLint clutter

cclauss:
The Python porting best practice is to [use feature detection instead of version detection](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection).  The __try / except (ImportError, NameError)__ approach proposed in this PR __placates the linters PyLint and flake8__ while substantially reducing the comment clutter of disable directives.  The __assert__ statements at the end also probably not necessary anymore but this PR leaves them in place.

### PR Checklist
      are Travis tests that cover the most important things however, you are welcome
      to rely on those.
- [ ] Ideally new or changed features have documention updates.

### DIFF
--- a/nuitka/__past__.py
+++ b/nuitka/__past__.py
@@ -30,15 +30,15 @@ import sys
 # pylint: disable=I0021,invalid-name,redefined-builtin
 
 # Work around for CPython 3.x renaming "long" to "int".
-if str is bytes:
-    long = long  # @ReservedAssignment pylint: disable=I0021,undefined-variable
-else:
+try:
+    long = long  # @ReservedAssignment
+except NameError:
     long = int   # @ReservedAssignment
 
 # Work around for CPython 3.x renaming "unicode" to "str".
-if str is bytes:
-    unicode = unicode  # @ReservedAssignment pylint: disable=I0021,undefined-variable
-else:
+try:
+    unicode = unicode  # @ReservedAssignment
+except NameError:
     unicode = str      # @ReservedAssignment
 
 def iterItems(d):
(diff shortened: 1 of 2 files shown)


cclauss:
The Python 3.3 fail on the first round of test looked like a fluke but lets see on the second round.

cclauss:
;-)  Python 3.3. now passes -- it was a fluke

But PyLint 2.1.1 adds new tests : https://travis-ci.org/kayhayen/Nuitka/jobs/419162745#L742-L751

kayhayen:
At one time, in Nuitka there was an issue, where a user was making "import builtins" in his code and having that module name just on the wrong Python versions. Also back ports happen. That is why I made it a rule to not use try/except ImportError anymore. Otherwise user code can influence Nuitka in hard to debug ways. See also commit 4f97a09e7c9a23c1e2c734a2d95a2ca2bfdc066e where I did that.

Basically, if it's a standard library import, then it shouldn't be done. I was lazy for the 2.6 total_ordering, because functools is there, then only the name will not be there, and that's OK, nothing a user environment changes because of absolute imports. 

As for the "NameError", yes, that could be done, I don't really care, but I think it makes sense to keep the test used the same, "str is bytes" basically means Python3. If we have to be explicit for imported standard library modules, why not also be it for the names?

What do you think?

kayhayen:
@cclauss These PyLint messages are regressions of PyLint 2.0.1 (even 2.0.0 second go with non-dev astroid), and *all* false alarms that have been reported upstream, but not yet been fixed in any new release. I am waiting for fixes of those before updating makes sense.

cclauss:
Understood.  Closing.

kayhayen:
Please do not let this discourage you.You are welcome in Nuitka land 
👍

PR #114, opened 2018-09-03
nuitka: fixed output filename on Unix like platforms

AnythingTechPro:
Fixes output executable filenames on Unix like platforms, executables on Windows end with the .exe extension while executables on a majority of Unix like operating systems have no file extension...

### DIFF
--- a/nuitka/MainControl.py
+++ b/nuitka/MainControl.py
@@ -814,7 +814,10 @@ def main():
             sys.exit(0)
 
         if Options.isStandaloneMode():
-            binary_filename = options["result_name"] + ".exe"
+            if Utils.getOS() == "Windows":
+                binary_filename = options["result_name"] + ".exe"
+            else:
+                binary_filename = options["result_name"]
 
             standalone_entry_points.insert(
                 0,


kayhayen:
I am OK with this change, esp. for standalone mode, there is no collision chance at all.

There is tests for standalone, that will fail then however, due to hardcoded output file assumptions:

This is in tests/standalone/run_all.py

```
    # Second use "strace" on the result.
    loaded_filenames = getRuntimeTraceOfLoadedFiles(
        path = os.path.join(
            filename[:-3] + ".dist",
            filename[:-3] + ".exe"
        )
    )
```

Probably stripping the ".exe" works for all platforms, or maybe it does not. Not sure if "Popen" will search for suffixes to execute, or insists on a fully expanded path. Probably it ought to just check what exists.

And then there is still this code:

`        exe_filename += ".exe"

in "nuitka/tools/testing/compare_with_cpython/__main__.py" which appends that suffix to find the binary to use in testing.

For non-standalone, I would like your opinion. Say you compile "program" in the current directory, what is the expected output? Wouldn't it be "program" ? That is why I added a suffix in the first place I think. Maybe just refuse to do it without an output dir, but then I liked that overwriting is default. These are non-trivial issues, the suffix has avoided. However, they do not apply to standalone mode, that I agree with.

On travis, standalone tests are not done, because they take too much CPU cycles and barely are needed. For you to see the issues, run "./tests/standalone/run_all.py search" and you will find errors raised due the codes I highlighted.

Actually we ought to maybe just make Nuitka report the output file when it finished with log.info() and parse that part of the output, to get at filenames on the outside, but that is just an additional idea, it would also help users who add "--standalone" flag to be less confused about where the final output is.

Anyway, would like to see this finished. I did not realize from looking at the code how the test runner handles ".dist" output for filename to execute. Very likely only "--run" is working for it. And then "two step execution" is just not used with standalone, which is forced only for coverage mode, where currently standalone might not be done, which is an existing bug. So you would probably only run into that ".exe" suffix in the run time traces check.

Yours,
Kay
`

AnythingTechPro:
On Unix (Linux, Mac OS), files with no extension are considered an executable application. I think this would be a better way to go because I actually pack up the application for Apple’s App Store and the executable file needs that exact name with no extension. When I get the chance I’ll fix the other references to executable extension like you’ve mentioned above.

kayhayen:
You probably saw my poll on Twitter as @kayhayen , everybody wants to have a "-o" option it seems, everybody hates the ".exe" suffix.

Therefore I added the "-o" option with the follow behavior:

a) Not allowed for standalone, there what your code does is used. The platform convention, i.e. ".exe" on Windows, empty suffix elsewhere.
b) Not allowed for module mode, as the name is part of the interface and cannot be changed.
c) For accelerated mode, there is now a *default* of ".bin" on non-Windows, which can be overridden using the "-o" parameter.

The advantage is that I still do not have to deal with considerations if overwriting is OK or not. It always is with those choices. We own the ".dist" folder, and the ".bin" suffix is not your typical Python scrip, yet people will be dramatically less surprised when they see it.

I am torturing Travis on factory with it, as it obviously breaks a lot of testing code.

This kind of invalidates your PR, but I consider it a full success since it pushed me over the line to go out and ask people what they want, and since "-o" is beyond your scope, it felt like I have to take this on myself. I hope you continue to create more PRs.

Thanks for your contribution!

PR #122, opened 2018-09-07
Fix #103, generate .lib and .exp files on build directory

### LINKED ISSUE #103: Wheels created on Windows contain useless files
Windows bdist_nuitka wheels contain a .lib and .exp file, both of which are useless and should not be included.

The files are small, but add only noise. Unsure if they are there for MSVC and/or MinGW64.

All it takes is to find the proper spot in ./nuitka/distutils/bdist_nuitka.py to delete them.

brupelo:
This should be ready to go!


kayhayen:
I cannot reopen because you deleted the branch.

PR #125, opened 2018-09-10
Feature implib

brupelo:


kayhayen:
Hello there,

> Commit Summary
>
>    - Fix #103, generate .lib and .exp files on build directory
>    - no message
>
> File Changes
>
>    - *M* nuitka/build/SingleExe.scons
>    <https://github.com/kayhayen/Nuitka/pull/125/files#diff-0> (11)
>
>
From what I can tell, this patch breaks MinGW64. You either should do the
no_import_lib for MSVC, or provide the proper options. Most probably there
is a mingw.py in the tools directories, that should hint you at the syntax.

Yours,
Kay

brupelo:
From SCons docs:

> no_import_lib
> When set to non-zero, suppresses creation of a corresponding Windows static import lib by the SharedLibrary builder when used with MinGW, Microsoft Visual Studio or Metrowerks. This also suppresses creation of an export (.exp) file when using Microsoft Visual Studio.

As you can see I've initialized no_import_lib to False instead non-zero but I guess that'll be alright. Unless SCons makes a very explicit comparison, ie:

    >>> False==0
    True
    >>> False is 0
    False

kayhayen:
Looks good to me, so I merged as a squash with your fixups. Thanks for your contribution.

PR #152, opened 2018-09-27
Minor cleanup

brupelo:


kayhayen:
I am making this into the next hotfix, seems innocent enough.

PR #156, opened 2018-10-01
Developer manual: Fix assignment order in example

Mortal:
With multiple assignment targets, Python assigns left-to-right.
Fix the example in the documentation to be left-to-right instead of
right-to-left.

Note that Nuitka already handles this correctly as evidenced by running
the following example through nuitka-run:

	i = 0
	a = [0, 0]
	i = a[i] = 1
	print(a)

The example first assigns 1 to i and then 1 to a[1], resulting in the
correct output [0, 1], as opposed to [1, 0] which would be output in a
right-to-left-assigning language.

### DIFF
--- a/Developer_Manual.rst
+++ b/Developer_Manual.rst
@@ -1461,16 +1461,16 @@ multiple simple assignments instead.
 .. code-block:: python
 
    _tmp = c
-   b = _tmp
    a = _tmp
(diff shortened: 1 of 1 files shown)


kayhayen:
Very nice, also like your example code and will add it to Nuitka tests as well.

PR #157, opened 2018-10-01
Fix several typos all over the place.

pmeerw:
This fixes several typos in documentation and comments which I notices while getting familiar with the project.


kayhayen:
Very nice, thanks and you are very welcome.

PR #158, opened 2018-10-02
Fix misc. typos

luzpaz:
Found via `codespell -q 3 --skip="./nuitka/build/inline_copy,./tests"`


kayhayen:
Codespell seems very cool, will merge right after Travis finishes.

Issue #159, opened 2018-10-02
cloudpickle nuitka.NuitkaOptimizationError


jamesstidard:
Hi,

I'm having some issues building a project with a [distributed](https://github.com/dask/distributed) dependency. I'm getting the following error:

```sh
nuitka.Errors.NuitkaOptimizationError: 
This statement does raise but didn't annotate an exception exit.

    <node co_argcount="1" co_filename="/home/super/.local/share/virtualenvs/nuitka_playground-THapz-Nf/lib/python3.7/site-packages/cloudpickle/cloudpickle.py" co_has_closure="False" co_has_stardict="False" co_has_starlist="False" co_is_optimized="True" co_kind="Function" co_kwonlyargcount="0" co_lineno="105" co_name="inner" co_new_locals="True" co_varnames="value,cell" code_flags="" kind="StatementsFrameFunction" line="105">
      <role name="statements" type="list">
        <node kind="StatementExpressionOnly" line="106">
          <role name="expression">
            <node code_name="cloudpickle$cloudpickle$$$function_1__make_cell_set_template_code$$$function_1_inner$$$function_1_lambda" kind="ExpressionFunctionRef" line="106" />
          </role>
        </node>
        <node is_temp="False" kind="StatementAssignmentVariable" line="107" owner="cloudpickle$cloudpickle$$$function_1__make_cell_set_template_code$$$function_1_inner" variable_name="cell">
          <role name="source">
            <node kind="ExpressionVariableRef" line="107" owner="cloudpickle$cloudpickle$$$function_1__make_cell_set_template_code$$$function_1_inner" variable_name="value" />
          </role>
        </node>
        <node kind="StatementReturnNone" line="105" />
      </role>
    </node>
```

Here's the source:

```python
# main.py
import distributed

if __name__ == "__main__":
    client = distributed.Client()
    client.close()
```

and building with `python -m nuitka main.py --follow-imports`

*  Nuitka version, full Python version and Platform (Windows, OSX, Linux ...)

   0.6.0.1
   Python: 3.7.0 (default, Oct  2 2018, 15:47:42) 
   Executable: /home/super/.local/share/virtualenvs/nuitka_playground-THapz-Nf/bin/python
   OS: Linux
   Arch: x86_64

*  How did you install Nuitka and Python (pip, anaconda, deb, rpm, from source,
   what is a virtualenv ...), this is very important usually.

I used pipenv 2018.7.1 which makes a virtual environment.

Thanks for your time & the project.

[...]

PR #161, opened 2018-10-03
fix: getFullName double append

jamesstidard:
### What does this PR do?
Makes a ignorant attempt to fix a syntax error.

#159


kayhayen:
That was quick. That package case, I don't know when it triggers, I am putting out a fixup now, it's meant to the a tuple, I am pushing that right now as an amend so I give you credit for the commit. Thanks for your report!

PR #163, opened 2018-10-04
Add command line flag --user-plugin

leojay:
### What does this PR do?
This PR adds a new command line flag --user-plugin.
This command line flag is used to load user plugins that are not in the Nuitka source code.

Right now, all plugin source files have to be in the source code of Nuitka.
Adding the flag would allow Nuitka users to write their own plugins.
And the source code of the plugin can be put anywhere.
It's even possible to write per project user plugin.
All UserPluginBase subclasses in the specified file will be imported, unless it's explicitly disabled with flag --plugin-disable

The flag can be used multiple times.

Usage example:
python -m nuitka --module package_test --include-package=package_test --user-plugin my-plugin.py

### PR Checklist

### DIFF
--- a/nuitka/OptionParsing.py
+++ b/nuitka/OptionParsing.py
@@ -741,6 +741,13 @@ Show list of all available plugins and exit. Defaults to off."""
 
 parser.add_option_group(plugin_group)
 
+plugin_group.add_option(
+    "--user-plugin",
+    action  = "append",
+    dest    = "user_plugins",
+    default = [],
+    help    = "The file name of user plugin. Can be given multiple times. Default empty."
+)
 
 def parseOptions():
     # First, isolate the first non-option arguments.
--- a/nuitka/Options.py
+++ b/nuitka/Options.py
@@ -444,6 +444,13 @@ def getPluginsDisabled():
     return tuple(set(options.plugins_disabled))
 
 
+def getUserPlugins():
+    if not options:
+        return ()
+
+    return tuple(set(options.user_plugins))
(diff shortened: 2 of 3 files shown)


leojay:
The tests failed because I'm using different libraries to import files in different Python versions. Not all libraries exist in all python versions.

kayhayen:
Very good job. For PyLint it will take something like this:

`# @UnresolvedImport pylint: disable=I0021,import-error,no-name-in-module`

I can do that for you though, no big deal. But looks good to me, thanks for tackling to all versions problem, was very positively surprised to see that. And indeed, that's code I really want, but didn't take the time to do it yet. Now we need a better API documentation that explains it.

I need to get the Doxygen work finished, so we can point people to well documented methods in the plugin base, and well add good doc strings.

Curious, are you writing your own plugins then? :)

leojay:
Just added pylint code. Please take another look.

Yes, I have written a plugin to automatically save build information in the generated .so file.
Here is how I'm doing it.
I have a `__init__.py` file in my package. And it looks like this:
```python
__version__ = '__GIT_REVISION__'
__build_host__ = '__BUILD_HOST__'
__build_timestamp__ = '__BUILD_TIMESTAMP__'
__source_dir__ = '__SOURCE_DIR__'
```

My plugin overrides the `onModuleSourceCode` method, and replaces the placeholders with actual values. This information helps me understand what source code version the .so file is.

kayhayen:
Thanks for your contribution.

PR #184, opened 2018-10-21
Exclude common glibc libraries from standalone build

4O4:
All libraries which come with glibc must be in sync to prevent
compatibility issues when distributing the executables to other
machines.

This means we should either not exclude any glibc libs and ship them all
(which may cause even more problems) or just exclude all of them
since they are related to and highly dependent  each other.


kayhayen:
Thanks for the research effort you have put into this. Let me know when you consider this ready for merge.

Some library names, e.g. `libBrokenLocale.so` were somewhat surprising to me. These libc people have really crazy stuff. :)

Looking forward to it.

4O4:
Yeah, they were surprising to me as well. And such an inconsistency in naming... wonder what happened there.

This pull request is actually a result of half of the weekend spent on painful debugging of the outdated OS, which I didn't have a root access to and which lacked basic dev tools like gdb and strace and I had to compile them manually. 

But god, I love challenges, this was real fun and helped me learn a some new things and forced me to lurk into the CPython codebase. I love projects like yours which give me such an opportunity :smile: It's also clean and well docummented so I can easily contribute.

I'll create another issue later for one more problem with standalone builds which I discovered along the way (tip: hashlib).

I've added some new tests, minor improvements in the execution process and hopefully silenced the linter which didn't like the long URL. I consider it done now, so If you are good with the new changes feel free to merge :)

4O4:
One of 7 build jobs (Python 2.7) is failing on PyLint I have no idea how to fix it...

> Using PyLint version: 1.9.3
> ************* Module nuitka.freezer.Standalone
> nuitka/freezer/Standalone.py:633 I0021 useless-suppression  Useless suppression of 'line-too-long'

kayhayen:
It's OK, I will take it from there. It's indeed rather strange, maybe an encoding issue of some sorts, where it can become confused about new lines. Just please allow at least until friday, my private life will be demanding today/tomorrow.

4O4:
Sure thing, thanks

kayhayen:
I merged it, only removing the .mypy_cache files that I am adding a .gitignore for.

I hope you don't mind me adding the standard copyright to the new test in addition to yours, it's the ASF license I want all on of it, since that is done automatically by a script, and of course I will add you to contributors list.

4O4:
Great! Yeah I'm totally ok with that

kayhayen:
Hello there,

on OSB, this gives an error for some platforms:

https://build.opensuse.org/package/live_build_log/home:kayhayen/Nuitka-experimental/CentOS_7/x86_64

It says:
```

[  510s] Comparing output of 'SocketUsing.py' using '/usr/bin/python' with flags silent, expect_success, standalone, remove_output ...
[  523s] Should not access '/lib64/libattr.so.1'.
[  523s] Should not access '/lib64/libbz2.so.1'.
[  523s] Should not access '/lib64/libcap.so.2'.
[  523s] Should not access '/lib64/libdw.so.1'.
[  523s] Should not access '/lib64/libelf.so.1'.
[  523s] Should not access '/lib64/liblzma.so.5'.
```

For Fedora 24 there is similar errors, I tried that in a virtual machine, but had no problem. I added "env" output, to see what it does different maybe, but nothing noticeable in there.

Do you have any idea @4O4 what this can be caused by? These libraries appear very basic, e.g. "libelf.so" definitely is (could be libc maybe), but "liblzma" and "libbz2" definitely are more libpython material.

On some platforms, e.g. newer Fedora, OSC has no issue. I would appreciate hints if you have any.

4O4:
libattr, libelf etc. are not part of glibc, they are installed separately. I suspect that CPython tries to use them if they are installed on the system (some deep dependency maybe), and if they're not it just keeps rolling without them. But this is a blind guess really, I would have to dig into cpython code to learn more about this specific issue.

kayhayen:
I recently noticed something, when looking at anaconda and standalone, there is a hack to make "libuuid.so" available, but for some reason I cannot get my head around, it was always loading "libuuid.so.1" which is also there, but it had no business picking, at least I couldn't figure out, where the .so version came from.

Might a similar or same issue, I will check if Nuitka loads "libelf.so.1" but includes "libelf.so" only, that could totally be what happens, esp, during RPM builds, maybe some config makes it go more specific with these things.

Yours,
Kay

4O4:
@kayhayen Usually the `.so` without number is simply a symlink to some specific (most often the newest) version of the lib. Also there are often symlinks from more generic major or minor version number to the specific ones. Basically the `lib` dir on linux is full of symlinks, so that's most likely the reason of the name inconsistency.

PR #205, opened 2018-12-12
Initial support for PySide2 added.

taimurrabuske:

### PR Checklist

### DIFF
--- a/nuitka/plugins/standard/PySidePyQtPlugin.py
+++ b/nuitka/plugins/standard/PySidePyQtPlugin.py
@@ -48,11 +48,23 @@ class NuitkaPluginPyQtPySidePlugins(NuitkaPluginBase):
     def __init__(self):
         self.qt_dirs = {}
 
-    def getPyQtPluginDirs(self, qt_version):
+    def getPyQtPySide2PluginDirs(self, qt_version, is_pyside):
         if qt_version in self.qt_dirs:
             return self.qt_dirs[qt_version]
+        if is_pyside:
+            command = """\
+from __future__ import print_function
 
-        command = """\
+import PySide2.QtCore
+for v in PySide2.QtCore.QCoreApplication.libraryPaths():
+    print(v)
+import os
+guess_path = os.path.join(os.path.dirname(PySide2.__file__), "plugins")
+if os.path.exists(guess_path):
+    print("GUESS:", guess_path)
+"""
+        else:
+            command = """\
 from __future__ import print_function
 
 import PyQt%(qt_version)d.QtCore
@@ -63,8 +75,8 @@ guess_path = os.path.join(os.path.dirname(PyQt%(qt_version)d.__file__), "plugins
 if os.path.exists(guess_path):
     print("GUESS:", guess_path)
 """ % {
-           "qt_version" : qt_version
-        }
+               "qt_version" : qt_version
+            }
 
         output = Execution.check_output([sys.executable, "-c", command])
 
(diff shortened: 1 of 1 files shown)


taimurrabuske:
Initial support for PySide2.

kayhayen:
Can you try tests/standalone/PySideUsing.py with your changes and maybe the probably trivial migration to PySide2, and if it even works? I think slots fail to work with PySide and no patches, but maybe PySide2 is improved.

kayhayen:
No documentation changes needed.

kayhayen:
Also, I was wondering if you won't have to about `nuitka/plugins/standard/ImplicitImports.py` as well for PySide2. In `getImplicitImports` there is many implicit dependencies for PyQt4/5 and PySide1, and those might need to be added, or not if there are `.pyi` files for PySide2, which would be great.

Before you start doing that, let me know, if you think it will be necessary, because there will be an easy way to tell which ones to add, but I won't want to describe it if it's not necessary.

taimurrabuske:
> Can you try tests/standalone/PySideUsing.py with your changes and maybe the probably trivial migration to PySide2, and if it even works? I think slots fail to work with PySide and no patches, but maybe PySide2 is improved.
Hello.
You're right. Slots fail the test. I could not find the patch in the repository. Can you hint on where they are, and I may investigate further and try to adapt.

taimurrabuske:
> Also, I was wondering if you won't have to about `nuitka/plugins/standard/ImplicitImports.py` as well for PySide2. In `getImplicitImports` there is many implicit dependencies for PyQt4/5 and PySide1, and those might need to be added, or not if there are `.pyi` files for PySide2, which would be great.
> 
> Before you start doing that, let me know, if you think it will be necessary, because there will be an easy way to tell which ones to add, but I won't want to describe it if it's not necessary.

I am under the impression that it may not be required. Maybe you could help me figure it out. In the following hello world example: 
```
import sys
from PySide2.QtWidgets import QApplication, QLabel

app = QApplication(sys.argv)
#label = QLabel("Hello World!")
label = QLabel("<font color=red size=40>Hello World!</font>")
label.show()
app.exec_()
```
the following files are automatically included in the .dist directory: 
libQt5Core.so.5
libQt5DBus.so.5
libQt5EglFSDeviceIntegration.so.5
libQt5Gui.so.5
libQt5MultimediaGstTools.so.5
libQt5Multimedia.so.5
libQt5MultimediaWidgets.so.5
libQt5Network.so.5
libQt5OpenGL.so.5
libQt5PrintSupport.so.5
libQt5Qml.so.5
libQt5Quick.so.5
libQt5Sql.so.5
libQt5Svg.so.5
libQt5WaylandClient.so.5
libQt5WebSockets.so.5
libQt5Widgets.so.5
libQt5XcbQpa.so.5

while the following files are included in .dist/PySide2
QtCore.so
QtGui.so
QtNetwork.so
QtWidgets.so

plus the qt-plugins directory.

I assume that since we only imported QtWidgets, but QtCore, QtGui and QtNetwork (?) are automatically included, implicit imports are not required. Am I seeing it correctly?

kayhayen:
That's great, and a nice improvement of PySide2. I would have to check it, but where PySide2 is installed, there must be `.pyi` files, and Nuitka understands there for a while now, and apparently they are getting used more often.

Allow me to check it out over the weekend, then I will merge. Also the patch, I think it was in the Pyside issue tracker, here: https://bugreports.qt.io/browse/PYSIDE-198

taimurrabuske:
> That's great, and a nice improvement of PySide2. I would have to check it, but where PySide2 is installed, there must be `.pyi` files, and Nuitka understands there for a while now, and apparently they are getting used more often.
> 
> Allow me to check it out over the weekend, then I will merge. Also the patch, I think it was in the Pyside issue tracker, here: https://bugreports.qt.io/browse/PYSIDE-198

Yes, that is great indeed. To corroborate with this, I just finished the compilation of a much more complicated Qt app with Nuitka, and it successfully included all the deps:
QtCore.so
QtGui.so
QtHelp.so
QtNetwork.so
QtOpenGL.so
QtSvg.so
QtTest.so
QtWidgets.so

taimurrabuske:
I think it is better to hold on a little before applying the patch. I just figured that we may run into some issues with Abstract Models. While trying to run the applications, I get
NotImplementedError: pure virtual method 'QAbstractItemModel.rowCount()' not implemented.
which is similar to 
http://bugs.nuitka.net/issue267

The method is implemented, and the app runs normally in a python interpreter.

kayhayen:
Sorry, until PySide starts accepting our patches, we better leave this.

shpi:
What is actual development state? I've seen Qt mades adaptions for nuitka. I could get some parts running, but QAbstractListmodel rowCount function throws not implemented error.

Issue #204, opened 2018-12-12
WinSxS msvcr90.dll conflict with Windows 10 Microsoft Feature update (Build 1803)


vis-123:
It appears a Microsoft Windows 10 update is breaking Nuitka at compile time for me, as it is unable to find the msvcr90.dll in the expected WinSxS folder. The issue ONLY appears after the MS update; seems the update is moving or deleting the msvcr90.dll from the WinSxS folders that nuitka looks for its dependency. I've tested it with rollbacks on multiple devices, and can confirm the microsoft update is definitely the culprit.

Edit: After more research, the issue may be related to dll dependency caching, will try nuitka with cl arg ```--force-dll-dependency-cache-update``` to force an update.

The specific Windows10 affected version and build are:
```
Version: 1803
OS Build: 17134.471
```

My nuitka version info:
```
0.6.0.6
Python: 2.7.15 (v2.7.15:ca079a3ea3, Apr 30 2018, 16:30:26) [MSC v.1500 64 bit (AMD64)]
Executable: c:\python27\python.exe
OS: Windows
Arch: x86_64
```

Err output:
```
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\nuitka\__main__.py", line 180, in <module>
    main()
  File "C:\Python27\lib\site-packages\nuitka\__main__.py", line 174, in main
    MainControl.main()
  File "C:\Python27\lib\site-packages\nuitka\MainControl.py", line 844, in main
    standalone_entry_points = standalone_entry_points
  File "C:\Python27\lib\site-packages\nuitka\freezer\Standalone.py", line 1187, in copyUsedDLLs
    target_path
  File "C:\Python27\lib\shutil.py", line 96, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: 'c:\\windows\\winsxs\\amd64_microsoft.vc90.crt_1fc8b3b9a1e18e3b_9.0.30729.9247_none_08e394a1a83e212f\\msvcr90.dll'
```

kayhayen:
Can you checkout the cache options of Nuitka on Windows and or erase it in AppData. I suspect that `depends.exe` output points to a DLL that n longer exists.

I guess that ought to make Nuitka update the cache for that item automatically, or warn, or something, but it can become outdated easily with any system update.

If you are willing to wait and preserve the "Nuitka" folder in AppData, I can make the change and use you to test it, would that be OK for you?

Yours,
Kay

[...]

PR #206, opened 2018-12-13
add automatic forced dependency cache update on missing/moved cached dependency

vi-s:
Here is an idea to possibly force a dependency cache update, if we detect that a cached dependency has been moved or missing (meaning the cache is invalidated). Not sure if this is the best approach, but hope it can serve as an idea to cut out issues where cached dependencies are moved (ex. for instance after a system update). Here is the related issue I had opened:
https://github.com/Nuitka/Nuitka/issues/204

### PR Checklist
- [ ] Correct base branch selected? `develop` for new fetures and bug fixes too. If it's


vi-s:
Made changes according to the PR comments, and PR updated.

vi-s:
Fix lint errors by replacing outer scope var with optional fn param

PR #219, opened 2019-01-18
Existing cache dir fix

deajan:
### What does this PR do?

Fix a small issue when half of the dependency cache path exists, hence makedirs will try to create it all and may fail.

No relevant issues, just less annoying when playing with dependency cache files.

### PR Checklist
- [X] Correct base branch selected? `develop` for new fetures and bug fixes too. If it's
- [X] All tests still pass. Check the developer manual about ``Running the Tests``. There
- [X] Ideally new features or fixed regressions ought to be covered via new tests.
     Not applicable
- [X] Ideally new or changed features have documentation updates.
     Not applicable

### DIFF
--- a/nuitka/utils/FileOperations.py
+++ b/nuitka/utils/FileOperations.py
@@ -62,7 +62,10 @@ def makePath(path):
     """ Create a directory if it doesn'T exist."""
 
     if not os.path.isdir(path):
-        os.makedirs(path)
+        try:
+            os.makedirs(path)
+        except FileExistsError:
+            pass
 
 
 def listDir(path):
--- a/nuitka/utils/FileOperations.py
+++ b/nuitka/utils/FileOperations.py
@@ -64,7 +64,9 @@ def makePath(path):
     if not os.path.isdir(path):
         try:
             os.makedirs(path)
-        except FileExistsError:
+        except OSError as err:
+            if err.errno is not os.errno.EEXIST:
+                raise
             pass
 
 
--- a/nuitka/utils/FileOperations.py
+++ b/nuitka/utils/FileOperations.py
@@ -67,7 +67,6 @@ def makePath(path):
         except OSError as err:
             if err.errno is not os.errno.EEXIST:
                 raise
-            pass
 
 
 def listDir(path):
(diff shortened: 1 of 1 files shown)


kayhayen:
I am confused, is it not that `os.makedirs` promises exactly that it creates intermediate directories? Can you provide an `rm` command to reproduce? I noticed something in the past I think, hence the top level `isdir()` check. Are you saying it will create stuff, and then exit with `OSError` when parts exist?

And if that is the case, I think it might happen a lot with multiple cache directories used.

kayhayen:
```
from nuitka.utils.FileOperations import makePath

makePath("deep/down/there")
makePath("deep/near/there")
```

But this works, does it not? There is that `os.path.isdir()` check there. Which of course would not catch on for files, but that would be weird anyway. Did not try on Windows, is it different there?

Yours,
Kay

deajan:
Your example works the same on Windows.
Yet, my Nuitka compilations triggers OSError 17 randomly in that part.
I have a hard time trying to find the reason and a reproductible code.
Maybe some concurrent issues when cache dirs are created by threads at the same time ?

kayhayen:
You are very likely correct, the dependency walker code is running in threads, and 2 of them might find the need to create the cache directory first time.

However, we might do away with that. That threading is just there because dependency walker is so horribly slow, but pefile maybe just isn't? You can set thread count to 1 there I guess to remove that effect for your branch.

Of course, in the light of that, your fix totally makes sense to me now and these kind of stupid races totally make sense.

Some form of `_withLock()` around the call to `makePath` might be missing right now, and maybe a directory existance check, so it's not acquiring the lock if it already exists. That ought to solve it for good.

Yours,
Kay

kayhayen:
Looks good to me, will merge once Travis completed.

deajan:
Could test the whole thing again today, seems to work for me.

kayhayen:
Why did you add @contextmanager to some of the functions though, that should not be necessary, right? Or is it?

deajan:
I do think it's needed to decorate a function that has with / yield statements.
I can be wrong too, being more or less a Python n00b :)

phil65:
only _with_file_Lock() needs the decorator. :)

kayhayen:
I removed the others in a cleanup commit, what surprised me what that it has no effect, but that is actually to be expected. The context manager provides for `__enter__` and `__exit__`, but not not change `__call__` other than wrapping it through a class.

Thanks for clarifying, and sorry for the noise. :)

Issue #137, opened 2018-09-15
Not including tkinter for standalone exe


JorjMcKie:
Cannot create standalone exe for scripts using tkinter.

* nuitka pip installation
* Compiler installed VS 2017
* output of nuitka generation:

```
Microsoft Windows [Version 10.0.16299.665]                                          
                                                                                    
Jorj@ARRAKIS C:\Users\Jorj\Desktop\test                                             
$ python -m nuitka --portable test.py                                               
                                                                                    
Jorj@ARRAKIS C:\Users\Jorj\Desktop\test                                             
$ python -m nuitka --version                                                        
0.5.33                                                                              
Python: 3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:59:51) [MSC v.1914 64 bit (AMD64)]
Executable: C:\Users\Jorj\AppData\Local\Programs\Python\Python37\python.exe         
OS: Windows                                                                         
Arch: x86_64                                                                        
```

* Script:

```python
import tkinter
root = tkinter.Tk()
root.bbox()
root.bell()
```

Output when running test.exe:

```
Jorj@ARRAKIS C:\Users\Jorj\Desktop\test\test.dist                                                                                                                                                                                              
$ test.exe                                                                                                                                                                                                                                     
Traceback (most recent call last):                                                                                                                                                                                                             
 File "C:\Users\Jorj\Desktop\test\test.dist\test.py", line 2, in <module>                                                                                                                                                                     
  File "C:\Users\Jorj\Desktop\test\test.dist\tkinter\__init__.py", line 2020, in __init__                                                                                                                                                      
_tkinter.TclError: Can't find a usable init.tcl in the following directories:                                                                                                                                                                  
    {C:\Users\Jorj\AppData\Local\Programs\Python\Python37\tcl} C:/Users/Jorj/AppData/Local/Programs/Python/Python37/tcl8.6 C:/Users/Jorj/Desktop/test/lib/tcl8.6 C:/Users/Jorj/Desktop/test/lib/tcl8.6 C:/Users/Jorj/Desktop/lib/tcl8.6 C:/User
s/Jorj/Desktop/test/library C:/Users/Jorj/Desktop/library C:/Users/Jorj/Desktop/tcl8.6.8/library C:/Users/Jorj/tcl8.6.8/library                                                                                                                
                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                               
[...]

PR #228, opened 2019-01-25
include new user plugins TkinterPlugin, NumpyPlugin

JorjMcKie:

### What does this PR do?
Implements two new user plugins for full support of **standalone compiles** of scripts which make use of tkinter and/or numpy.

The **tkinter** plugin is required on Windows only and aims to resolve issue #137. If the plugin is erroneously used on other platforms, it will disable itself and issue an information message.

The **numpy** plugin takes care of compiling numpy scripts successfully in standalone mode. At least on platforms Windows and Linux, numpy standalone compiles will not lead to a usable executable: you will see an exception about a missing ``_dtype_ctypes`` module. In addition, specifically for ``numpy+MKL`` (Math Kernel Library) installations, binaries will not be copied (correctly) to the script's ``dist`` folder, because they are not detectable by ``Dependency Walker`` and similar tools. The plugin will resolve this by copying relevant binaries from ``numpy/core`` to the ``dist/numpy/core`` folder.

### PR Checklist

### DIFF
--- a/nuitka/plugins/user/NumpyPlugin.py
+++ b/nuitka/plugins/user/NumpyPlugin.py
@@ -26,19 +26,12 @@ because these are not detectable by dependency walkers.
 import os
 import shutil
 import sys
-import glob
 import pkgutil
 import re
 from logging import info
 
 from nuitka import Options
-from nuitka.plugins.PluginBase import UserPluginBase, pre_modules, post_modules
-from nuitka.utils import Execution, Utils
-from nuitka.utils.FileOperations import (
-    getFileList,
-    getSubDirectories,
-    removeDirectory
-)
+from nuitka.plugins.PluginBase import UserPluginBase, pre_modules
 
 #------------------------------------------------------------------------------
 # The following code is largely inspired by PyInstaller hook_numpy.core.py
@@ -120,7 +113,7 @@ def get_numpy_core_binaries():
         re_mkllib = re.compile(r'^(?:lib)?mkl\w+\.(?:dll|so|dylib)', re.IGNORECASE)
         dlls_mkl = [f for f in os.listdir(lib_dir) if re_mkllib.match(f)]
         if dlls_mkl:
-            print("MKL libraries found when importing numpy. Adding MKL to binaries")
+            info("MKL libraries found when importing numpy. Adding MKL to binaries")
             binaries += [(os.path.join(lib_dir, f), '.') for f in dlls_mkl]
 
     return binaries
--- a/nuitka/plugins/user/NumpyPlugin.py
+++ b/nuitka/plugins/user/NumpyPlugin.py
@@ -1,4 +1,4 @@
-#     Copyright 2018, Kay Hayen, mailto:kay.hayen@gmail.com
+#     Copyright 2019, Jorj McKie, mailto:lorj.x.mckie@outlook.de
 #
 #     Part of "Nuitka", an optimizing Python compiler that is compatible and
 #     integrates with CPython, but also works on its own.
@@ -15,10 +15,12 @@
 #     See the License for the specific language governing permissions and
 #     limitations under the License.
 #
-""" User plug-in to make tkinter scripts work well in standalone mode.
+"""
+Plug-in to ensure numpy scripts compile and work well in standalone mode.
 
-To run properly, scripts need copies of the TCL / TK libraries as sub-folders
-of the script's dist folder.
+In addition, if the numpy+MKL (Intel's Math Kernel Library) version is
+installed, this plugin will copy its binaries to the dist/numpy/core folder,
+because these are not detectable by dependency walkers.
 """
 
 import os
@@ -30,7 +32,7 @@ import re
 from logging import info
 
 from nuitka import Options
-from nuitka.plugins.PluginBase import NuitkaPluginBase, pre_modules, post_modules
+from nuitka.plugins.PluginBase import UserPluginBase, pre_modules, post_modules
 from nuitka.utils import Execution, Utils
 from nuitka.utils.FileOperations import (
     getFileList,
@@ -88,21 +90,28 @@ def get_package_paths(package):
     return pkg_base, pkg_dir
 
 def get_numpy_core_binaries():
+    """Return any binaries in numpy/core, whether or not actually used
+    by our script. This covers the special case of MKL binaries, which
+    cannot be detected by dependency managers.
+    """
+    # covers/unifies cases, where sys.base_prefix does not deliver
+    # everything we need and / or is not an abspath.
     base_prefix = getattr(sys, 'real_prefix',
                           getattr(sys, 'base_prefix', sys.prefix)
                          )
     base_prefix = os.path.abspath(base_prefix)
-    is_win = sys.platform == "win32"
+    is_win = os.name == "nt"
 
     binaries = []
 
     # look for libraries in numpy package path
+    # should already return MKLs in ordinary cases
     pkg_base, pkg_dir = get_package_paths('numpy.core')
     re_anylib = re.compile(r'\w+\.(?:dll|so|dylib)', re.IGNORECASE)
     dlls_pkg = [f for f in os.listdir(pkg_dir) if re_anylib.match(f)]
     binaries += [(os.path.join(pkg_dir, f), '.') for f in dlls_pkg]
 
-    # look for MKL libraries in pythons lib directory
+    # also look for MKL libraries in pythons lib directory if present
     if is_win:
         lib_dir = os.path.join(base_prefix, "Library", "bin")
     else:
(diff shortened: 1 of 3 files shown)


nuitka/plugins/user/NumpyPlugin.py
```diff
@@ -0,0 +1,195 @@
...
+    """ This is for plugging in numpy support correctly.
+    """
+
+    plugin_name = "numpy-plugin"
+
+    def __init__(self):
+        self.bin_copied = False        # indicate binaries copied yet
+
+    @staticmethod
+    def createPreModuleLoadCode(module):
```
kayhayen: In the `ImplicitImports` plug-in there is an easier way detailed.

JorjMcKie:
Uff - pylint is zufrieden.

@kayhayen thanks a lot for your feedback.
Followed your recommendations. Should have thought of implicit import myself ...

kayhayen:
I think closing prevents me from merging, I am re-opening it and going to merge later today/tomorrow.

kayhayen:
Thanks for your contribution @JorjMcKie and the plugin mechanism is not nearly documented enough, definitely my fault to not do so better.

I am somewhat surprised of the numpy plugin copying data files, I am not sure what it is for. I was wondering how to change `NumpyUsing.py` to exhibit those needs. Any suggestions?

JorjMcKie:
Thanks for the merge.
``NumpyUsing.py``? Do you mean ``NumpyPlugin.py``?
Anyway, I share your frowning reaction. I am not quite surre where to begin. Und ich will keine Eulen nach Athen tragen - also bitte um Vergebung, falls es zu ausführlich sein sollte.
I am a mathematician and have been using numpy and scipy a lot over the recent few years (linear algebra and non-linear function approximation mostly).
Bottom line is:
Numpy vanilla is okay for household, occasional use. As soon as it comes to increased performance needs, some of the advanced packages are just inevitable, which add highspeed drop-in replacements for numpy's own routines. Intel's MKL (optimized for their processors I understand) is just one example of yet other packages (OpenBlas, LAPACK, ...), which add internal multithreading (i.e. not using Python's) and other highly optimized routines. In any case, the package against which numpy is linked, is invisible on the Python programmer's API level.
Several packages (scipy, scikit-learn) require numpy+MKL and do not even work otherwise.
The problem with specifically MKL is that there is only one DLL detectable by dependency walkers: ``mkl_rt.dll``. This guy in turn invokes any of the other lot as required - but their names are not visible by these tools - although I can see there names enumerated in the binary code.
So I see no way but copying them all in standalone compile mode - thank God they are in the ``num/core`` folder and this is how the plugin works. BTW in exactly the same way as PyInstaller approaches this.
There actually is a numpy variable set somewhere which indicates that this is an MKL version. And dependency walker does find ``mkl_rt.dll``.
When testing the plugin in Linux, where I have vanilla numpy, the plugin finds nothing in ``numpy/core`` and accepts this with a shrug.

What I can do, is putting a (shorter version of) this comment in the plugin source code.

As I mentioned when chatting with @deajan:
The MKL is **large**: 560 MB (but well compressible by UPX, down to about 140 MB). So I doubt that many people will actually build standalone applications with their scripts based on numpy+MKL. But probably better to have something workable sitting on the shelf than having to announce, that numpy+MKL cannot be compiled standalone ...

FabsM:
Hi guys, 
first of all thank you @kayhayen for nuitka! And a huge thank you to @JorjMcKie for these plugins. I came across nuitka today and I finally had it work, the only problem was that I couldn't actually run my program because it kept crashing. I then figured out that the problem was the lack of support for tkinter and I don't know how I saw the issue #137, which led me to this. I then changed the nuitka code and here we go!! It works perfectly!! 
Thank you again for this amazing work.

Fabio

JorjMcKie:
@kayhayen

It's a pleasure contributing to this extremely interesting project, I assure you.

I am not sure I understand your last comment around "... plugin warning too often".
If you alluded to the fact, that literally **every** Python script will come up with that warning, then you are definitely right! Even compiling the one-liner ``print("hello, world")`` will issue it.

I think this is proof, that the plugin should defy issuing this warning. How can this be done? By returning always ``False`` from ``isRelevant()``?

FabsM:
> 
> 
> 
> I am testing it right now, and except for the detector part of the plugin warning too often, it's working like a charm it seems. I think we agreed on a source code snippet to look for, @JorjMcKie or is that too limited?

@kayhayen , yesterday when I tried, it didn't warn at all. The only things that it printed were:
' Now copying tkinter libraries.' and ' Finished copying tkinter libraries.'. Am I missing something?

kayhayen:
@FabsM This is about warning of not having a plugin enabled. Doing it for every compilation is too broad, @JorjMcKie it should be like in multiprocessing plugin, where user code triggers it. But that is probably incomplete.

Plugins ought to be able to see call or usage of a module import. In the Tk case, "tkinter.Application" seems  fit. I think that due to star imports, this can easily not be seen by Nuitka though.

I guess, I will take it onto myself to extend the plugin API so module imports and their usage can be seen by plugins. For multiprocessing this works nicely:

```
    def onModuleSourceCode(self, module_name, source_code):
        if module_name == "__main__":
            if "multiprocessing" in source_code and "freeze_support" in source_code:
                self.warnUnusedPlugin("Multiprocessing workarounds for compiled code on Windows.")

        return source_code
```

But I guess, I would rather want to see:

```
    def onModuleVariableCall(self, module_name, variable_name):
        if module_name == "__main__":
                self.warnUnusedPlugin("Multiprocessing workarounds for compiled code on Windows.")
```

Now that I think of it, if the main program imports `tkinter`, what else do we need to know? A similar approach like that one would work indeed, if we can limit to `__main__` name, would it not? I guess, I am going with that then and it will be enough.

Currently I have disabled the warning by commenting it out on factory. But for the pre-release later today, I intend on making that kind of change instead.

kayhayen:
As for the NumpyPlugin, I consider moving the implicit import to the plugin that contains most of them already, because I don't think anybody wants to ever not use that, and that's on by default.

Also, I was then considering to rename it to something `numpy-mkl-plugin` or so. But I have yet to figure out, if there are binaries for numpy without MKL at all. I have to admit, I have to read your comment explaining things more carefully as you already explained why the copying can become useful or necessary.

It's detector then ought to maybe trigger based on presence of MKL installation, if it's limited. Or if plain numpy needs it, I would split it up in two plugin classes, instead of one, where one would become on by default, and the other would not, something like that @JorjMcKie . For that I have to try it out myself which I also have to admit to not doing yet.

Generally looks good to me though, more about shuffling code around.

JorjMcKie:
Fully agree with all you wrote, @kayhayen. The best plugin is a dead plugin so to say ...

**Tkinter:**
Yes, if the main script visibly does a tkinter import, then that's sufficient info. But if it only imports something like ``PySimpleGUI`` then what?

Numpy:
Yes, the implicit import part should go to where the rest of them sits, ``ImplicitImports.py``.
I can think of two triggers to know we have an MKL-installation:
(1) the sheer presence of its root binary ``mkl_rt``
(2) There is a ``_distributor_init.py`` in all numpy installations. It is empty in my Linux numpy installation, but it has an attribute ``NUMPY_MKL`` in my numpy+MKL on Windows:

```python
>>> from numpy import _distributor_init as dist
>>> dist.NUMPY_MKL
True
>>> dist.path
'C:\\Users\\Jorj\\AppData\\Local\\Programs\\Python\\Python37\\lib\\site-packages\\numpy\\core'
```
I have no idea how other speed-up libraries (OpenBlas, LAPACK) would handle it, but that file is meant as a place where to document such stuff. It's __doc__ on Linux says this:

```python
>>> print(dist.__doc__)
 Distributor init file

Distributors: you can add custom code here to support particular distributions
of numpy.

For example, this is a good place to put any checks for hardware requirements.

The numpy standard source distribution will not put code in this file, so you
can safely replace this file with your own version.

>>>
```
Maybe a viable way is to concentrate on the binary decision "plain numpy" vs. "numpy+MKL" for a start. If plain numpy, then nothing special is required.
Otherwise we take the path in ``_distributor_init.py`` and copy the binaries from there treating them as data files.
Maybe I find examples for other numpy distributions so we can see what the world for them looks like ...

FabsM:
> 
> 
> @FabsM This is about warning of not having a plugin enabled. Doing it for every compilation is too broad.

Sorry, I didn't understand what you meant. I totally agree with you then. Just wanted to say, if you need me to test anything, just ask and I will be willing to help you with this project.

JorjMcKie:
I just installed OpenBlas packages on some of my Pythons:

# Linux (Ubuntu)
pip-installed blis. This package installs openblas (plus vanilla ``numpy`` where necessary). After this, additional ``.so`` files are locatable like this:
```python
>>> numpy.__config__.extra_dll_dir
/usr/local/lib/python3.6/dist-packages/numpy/.libs
>>> os.path.exists(numpy.__config__.extra_dll_dir)
True
>>> os.listdir(numpy.__config__.extra_dll_dir)
['libgfortran-ed201abd.so.3.0.0', 'libopenblasp-r0-382c8f3a.3.5.dev.so]
```

# Windows
Again ``pip install blis``. Same procedure as above. This leads to the following results:
```python
In [3]: np.__config__.extra_dll_dir
Out[3]: 'c:\\users\\jorj\\appdata\\local\\programs\\python\\python36\\lib\\site-packages\\numpy\\.libs'

In [4]: x=np.__config__.extra_dll_dir

In [5]: import os

In [6]: os.path.exists(x)
Out[6]: True

In [7]: os.listdir(x)
Out[7]: ['libopenblas.IPBC74C7KURV7CB2PKT5Z5FNR3SIBV4J.gfortran-win_amd64.dll']
```
Note that ``extra_dll_dir`` is contained in my search meachanism of the NumpyPlugin.

I will now try to standalone-compile with this configuration, but **without** the NumpyPlugin and including the implicit import.
"I'll be back ..."

JorjMcKie:
Okay, here are the new numpy test results (Windows, using standard ``ImplicitImports.py``, ``NumpyPlugin.py`` changed correspondingly):

Invocation:
```
python -m nuitka --standalone --disable-dll-dependency-cache --python-flag=nosite --experimental=useInternalDependencyWalker numpy-test.py
```

This caused @deajan's dependency walker to realize the use of ``libopenblas.IPBC74C7KURV7CB2PKT5Z5FNR3SIBV4J.gfortran-win_amd64.dll``, but it couldn't locate it. The compile otherwise worked, therefore the implicit import was successful, too.

Executing the generated EXE lead to an exception - not surprisingly. When manually copying over the above-mentioned `.lib` folder to form ``dist\numpy\.lib`` everything worked fine again.

Bottom-line to me this means, that an extra file copy like the one currently done in NumpyPlugin is indeed required. Although I now have to tweak it a little ...

JorjMcKie:
1. uninstall my previous numpy (via pip)
2. ``pip install blis`` - this installs openblas and also vanilla numpy (12 MB or so)
3. modify nuitka plugin ``ImplicitImports.py`` to somewhere include the 2 lines
```python
        elif full_name == "numpy":
            yield "numpy.core._dtype_ctypes", True
```
4. the above nuitka compile invocation

This should lead to a successful compilation. But - as per my version of the dependency walker - **not** to an executable EXE:
During compilation one will see 2 or 3 complaints, that the said dll can not be found, and it is indeed missing in the dist folder and sub folders afterwards. If manually copying the ``.libs`` subfolder from the Python numpy installation folder to ``dist\numpy\.dist``, then everything is ok.

JorjMcKie:
@deajan I have just finished testing my modified numpy plugin.
It now supports OpenBlas and MKL (as before) versions of numpy. I have also removed this implicit import invocation, which must now be present in the **standard** plugin for that purpose.

The attractive aspect of it is that numpy vanilla and numpy + OpenBlas are way, **way** smaller than this giant MKL thing: my numpy test script dist folder is just 64 MB in size (20 MB UPX compressed!).

Here is the run log and the used NumpyPlugin.py. Your current ``Standalone.py`` has been used, too.
[Nuitka.zip](https://github.com/Nuitka/Nuitka/files/2805160/Nuitka.zip)

I'm not sure what Kay is currently focussing on, so I didn't PR the new plugin version.

I don't know if you agree: it seems to me that we cannot / should not completely get rid of some logic that copies those extra ``.dll`` / ``.so`` files.
This is very similar to the tkinter situation.
It looks like there must some logic present in plugins which triggers those file copies.
Even if one would undergo the effort to automatically trigger all those copies (and PyInstaller in fact **does** that!), the outcome would be less than favorable because of all the size bloat.

kayhayen:
@JorjMcKie do not hesitate, I am currently somewhat distracted by child illness, will not change much outside of factory myself and definitely will welcome a PR that does this. I suspect however that `numpy.core` should be more correct as the one that really has that implicit dependency, but that is only a suspect.

When you execute normal Python with `-v` flag you can see the order of imports relatively OK, and it should only import that after `numpy.core`, or so I am guessing.

kayhayen:
I am assuming I will make currently factory to develop this evening.

JorjMcKie:
Best wishes for your kids! Hope not anything serious?
I am glad this type of distraction lies behind me for 3 decades now ... ;-)

You are right, this ``_dtype_ctypes`` will be invoked during the ``numpy.core`` import. I will try using ``numpy.core`` instead.

PR #235, opened 2019-01-28
Pe file doc

deajan:
### What does this PR do?
Document the PE file internal dependency walker.

### PR Checklist
- [X] Correct base branch selected? `develop` for new fetures and bug fixes too. If it's
- [X] Ideally new features or fixed regressions ought to be covered via new tests.
    Not applicable
- [X] Ideally new or changed features have documentation updates.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -359,6 +359,36 @@ all contained in one binary.
 So if feasible, aim at static linking, which is currently only possible with
 AnaConda Python on non-Windows.
 
+Windows Standalone executables and dependecies
+----------------------------------------------
+
+The process of making Standalone executables for Windows traditionnaly involves
+using an external dependency walker in order to copy necessary libraries along
+with the compiled executables to the distribution folder.
+Usng the external dependency walker is quite time consuming, and may copy
+some unnecessary libraries along the way (better have too much than missing).
+Since Nuitka 0.6.2, there's an experimental alternative internal dependency
+walker that relies on pefile which analyses PE imports of executables / libraries.
+This implementation shall create smaller Standalone distributions since it won't
+include Windows' equivalent of the standard library, and will speed-up first
+Nuitka compilations by an order of magnitude.
+In order to use it, make sure you have pefile installed via ```python -m pip install pefile```
+Once installed, you may enable the internal dependency walker by using the
+following switch:
+```--experimental=use_pefile```
+The pefile dependency walker will test all dependencies of the distribution folder.
+
+Optionnaly, it is also possible to check all recursive dependencies of included libraries
+using the following switch along with the above one:
+```--experimental=use_pefile_recurse```
+
+Some modules may have hidden dependencies outside of their directory. In order for
+the pefile dependency walker to find them, you may also scan the whole site-packages
+directory for missing dependencies using:
+```--experimental=use_pefile_fullrecurse``` along with the two above switches.
+Be aware that using this switch will increase compilation time alot.
+
+
 Where to go next
 ================
 
--- a/README.rst
+++ b/README.rst
@@ -385,7 +385,7 @@ following switch:
 
     The pefile dependency walker will test all dependencies of the distribution folder.
 
-    Optionnaly, it is also possible to check all recursive dependencies of included libraries
+    Optionnally, it is also possible to check all recursive dependencies of included libraries
     using the following switch along with the above one:
 
 .. code-block:: bash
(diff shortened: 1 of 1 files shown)


deajan:
First write up, RFC.

README.rst
```diff
@@ -359,6 +359,55 @@ all contained in one binary.
...
+
+.. note::
+
+    Some modules may have hidden dependencies outside of their directory. In order for
+    the pefile dependency walker to find them, you may also scan the whole site-packages
+    directory for missing dependencies using the following switch along with the two above:
+
+.. code-block:: bash
+
+    python -m nuitka --standalone --experimental=use_pefile --experimental=use_pefile_recurse --experimental=use_pefile_fullrecurse myprogram.py
```
kayhayen: I have to see how this works. For `depend.exe` it seemed barely necessary. I think it uses all sub-directories of the package we are working for.

Also more search directories, why should they make things slower in a way we notice. I will however check that out.

kayhayen:
What I was thinking, would you @deajan and @JorjMcKie be interested in joining the Nuitka organisation on Github, I think you both demonstrate that you not only want to do contributions in a drive by fashion, but actually care about maintaining it. Which is fantastic. That said. I have yet to add you to the list of contributors, that is coming anyway.

README.rst
```diff
@@ -359,6 +359,55 @@ all contained in one binary.
...
+using an external dependency walker in order to copy necessary libraries along
+with the compiled executables to the distribution folder.
+Usng the external dependency walker is quite time consuming, and may copy
+some unnecessary libraries along the way (better have too much than missing).
+Since Nuitka 0.6.2, there's an experimental alternative internal dependency
+walker that relies on pefile which analyses PE imports of executables / libraries.
+This implementation shall create smaller Standalone distributions since it won't
+include Windows' equivalent of the standard library, and will speed-up first
+Nuitka compilations by an order of magnitude.
+In order to use it, make sure you have pefile installed via ```python -m pip install pefile```
```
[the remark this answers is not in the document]
deajan: I actually modded requirements.txt for this, but since it's only Windows related, it was overkill for every nuitka install.

README.rst
deajan: It is useful to scan all directories for missing dll files.
I had a case with pywin32, where pythoncom37.dll is in pywin_system32 and not in the win32 module folder. This resolves such issues.
Of course for pywin32, since most windows executables will use it, I made a hard coded fix.

And yes, it's very slow since it will check the whole site-packages root for every dependency :)

deajan:
@kayhayen I'll be honored to join the Nuitka organisation, but keep in mind that I am not a python expert nor am I working as a developer, I'm just a curious guy who likes to code on spare time :)
I have a couple of projects that require Nuitka, and therefore am willing to contribute of course.

I can indeed maintain the windows dependency walker, and still have some ideas that could make it faster and more reliable. Those ideas might even get ported to the Linux dependency walker, since both do more or less the same job (thinking of a list of already satisfied dependencies that don't need to walk through files again, and some better recurse code).

JorjMcKie:
@kayhayen sure, con mucho gusto.
As I once told @deajan, I am a retired person with lots of spare time - as my own repo, PyMuPDF, has gotten quite saturated lately ...

kayhayen:
Thanks for the contribution. I am going to live up on the issue of detecting `mscrt90.dll` as soon as I recover from my currently sickness, and resume comparing it. The goal is definitely to make your code the default.

deajan:
Hope you'll be well soon.
Do you have an example code which misses mscrt90.dll so I can try to work something out ?

kayhayen:
This is the code Nuitka uses to determine where the CPython DLL comes from:

https://github.com/Nuitka/Nuitka/blob/d4cab74cd3a189c407484ab623d5e62b3070f406/nuitka/PythonVersions.py#L142

And if you want to take it on, you can make it a generic function or copy it more less. Like I said, the SxS config of CPython 2.x contains only that DLL.

But that might be totally unnecessary, if you look at this:

https://github.com/Nuitka/Nuitka/blob/d4cab74cd3a189c407484ab623d5e62b3070f406/nuitka/build/SingleExe.scons#L1680

For accelerated binaries, they will not find `mscrt90.dll` unless through SxS, therefore Nuitka is adding it, however not doing that for standalone would be good. Just add `and standalone_mode` and there is no SxS anymore. Then still, your `pefile` code will still not find it (seems it was supposed to be used only through SxS), but you could special case that library then, and find it from the running CPython via above code.

You do now have push rights to Nuitka/Nuitka repository. Feel free to use it for creating your feature branches right there. But that is not urgent. I will write more about that later this week though.

Issue #232, opened 2019-01-26
GSOC 2019: Nuitka all built-ins optimized


kayhayen:
Project description: Nuitka has support for many built-ins, e.g. `len` already, which means dedicated C code, compile time evaluation, type shapes produced (in this case an `int`), but there are some notable exceptions, e.g. `enumerate` where we know types too, that are still missing, but definitely can have high performance impact on some loops.

Your task would be to immitate existing built-in codes to achieve a complete support for ultimately all C built-ins. The first step would be to identify which ones are missing (by means of a warning added), then to find out in test runs of the test suites, which ones are warned about, and to resolve as many of those as possible. It is assumed that all should be possible.

Skills: Python and C programming, platform wouldn't matter

bksahu:
Hello @kayhayen

This sounds really interesting so I am interested to work on it this summer. I am proficient in Python and also have some experience in C. Could you please let me know how should I get started? Also could you please provide some relevant pointers to this issue?

bksahu:
Hello @kayhayen

This sounds really interesting so I am interested to work on it this summer. I am proficient in Python and limited knowledge of C. Could you please let me know how should I get started? Also could you please provide some relevant pointers to this issue?

bksahu:
Hello @kayhayen

This sounds really interesting so I am interested to work on it this summer. I am proficient in Python and also have some experience in C. Could you please let me know how should I get started? Also could you please provide some relevant pointers to this issue?

kayhayen:
I am currently updating the ideas page with more details. I will also post information to this issue later today.

kayhayen:
So that update is out. Basically this is pointer number one, each (called) built-in at compile time is represented with a node like this one: 

https://github.com/Nuitka/Nuitka/blob/cb0e0beb72041adeab1065306d5413a6bd868611/nuitka/nodes/BuiltinLenNodes.py#L52

Then, should this node not be optimized away during compile time, there is code generation for it that then points to a function that emits C code with an entry in a dictionary:

https://github.com/Nuitka/Nuitka/blob/cb0e0beb72041adeab1065306d5413a6bd868611/nuitka/codegen/CodeGeneration.py#L540

This results then in pretty generic code:

```
def generateBuiltinLenCode(to_name, expression, emit, context):
    generateCAPIObjectCode(
        to_name          = to_name,
        capi             = "BUILTIN_LEN",
        arg_desc         = (
            ("len_arg", expression.getValue()),
        ),
[...]

Issue #138, opened 2018-09-16
Autoformat of Nuitka source code


brupelo:
My guess it's at this point in time would be very time consuming to achieve and really hard to execute but I still can ask about it. I'm also interesting to know the best way to tackle this particular problem in a general way myself as I've got quite a lot of old legacy code lying on my hard drive from time using the old python2.6 code, at that time I was using C coding conventions all over the code instead following the more standard practices.

Anyway, consider this some sort of hypothetical question rather than a forma request, do you think would be feasible to convert the whole Nuitka project into [PEP8](https://www.python.org/dev/peps/pep-0008/) by using some working out of the box tool or by using some sort of fast refactoring method? If you asked me whether this would have to be done manually I would think of the whole issue as a quite costly few weeks job :/

I've found there is this tool [https://github.com/abarker/camel-snake-pep8](https://github.com/abarker/camel-snake-pep8) that could help somehow but I'm quite skeptical about its usage and definitely I wouldn't trust that tool to parse/transform my projects magically with it, dunno, haven't tried.

Anyway, do you think this would be possible to achieve? Even better to ask, would this conversion to PEP8 bring real benefits to the table or it would be just a waste of effort?

Btw, I already know PEP8 says:

> Many projects have their own coding style guidelines. In the event of any conflicts, such project-specific guides take precedence for that project.

but even so... ;)

kayhayen:
I have evaluated "black" semi-successfully for work, crashed on a few files I only hate how it treats named arguments by unaligning them.

Earlier this week I tried to evaluate it on Nuitka again. The autoformat based on RedBaron that i created could undo it. In fact, I was contemplating to make things first go through black and then only adjust the worst things it does or bite the apple and go full black.

I don't feel like deciding this right now. I was going to query black community what they think of my named argument love, probably not too much, although they do clearly like alignment.

The alternative, but not for Nuitka, is autopep8 which I was initially checking out. But it just makes code unreadable, hating vertical whitespace, which is wrong. Black is so much better.

The diff of blackened Nuitka vs. unblackend Nuitka is curiously small other than named arguments for what I recall. I like their stance too. But it means to let go of some alignments. We will see.

kayhayen:
Quick 0.6.0rc4 has the C code formatted with clang-format to the LLVM coding style, this should also aid that direction. Generated code probably ought to move in the same direction for sanity.

kayhayen:
[...]

PR #246, opened 2019-02-06
Adding support for "any" built-in

### LINKED ISSUE #349: Unable to predict for dict and set while optimizing "any" built-in
As discussed in #246 : Unable to predict `dict` and `set` datatypes while optimizing "any" built-in because those datatypes are not indexable. 
I suspect this is primarly because this line while optimization

https://github.com/Nuitka/Nuitka/blob/efd876c1e02fa9a420d39100356e0bbffccde7b1/nuitka/nodes/ExpressionBases.py#L397

**Reproducer**

**1.** Script:  `set.py`

```
print(any(set([0, 1, 2, 3, 3])))
```
Executed:
`$ ~/Nuitka/bin/nuitka3-run --verbose set.py `

Output:
```
Nuitka:DEBUG:set.py:1 : new_expression : Using original '__file__' value.
Nuitka:DEBUG:set.py:1 : new_builtin_ref : Module variable 'print' found to be built-in reference.
Nuitka:DEBUG:set.py:1 : new_builtin_ref : Module variable 'any' found to be built-in reference.
Nuitka:DEBUG:set.py:1 : new_builtin_ref : Module variable 'set' found to be built-in reference.
Nuitka:DEBUG:set.py:1 : new_builtin : Replaced call to built-in 'set' with built-in call 'EXPRESSION_BUILTIN_SET'.
Nuitka:DEBUG:set.py:1 : new_builtin : Replaced call to built-in 'any' with built-in call 'EXPRESSION_BUILTIN_ANY'.
Nuitka:DEBUG:set.py:1 : new_constant : Built-in call to 'set' pre-computed. Predicted constant result.
Problem with statement at set.py:1:
-> print(any(set([0, 1, 2, 3, 3])))

Nuitka:INFO:Interrupted while working on '<Node 'PYTHON_MAIN_MODULE' with {'filename': 'set.py', 'main_added': False, 'mode': 'compiled'}>'.
  File "/home/bksahu/Nuitka/nuitka/__main__.py", line 184, in <module>
    main()
  File "/home/bksahu/Nuitka/nuitka/__main__.py", line 177, in main
    MainControl.main()
  File "/home/bksahu/Nuitka/nuitka/MainControl.py", line 745, in main
    main_module = createNodeTree(filename=filename)
  File "/home/bksahu/Nuitka/nuitka/MainControl.py", line 139, in createNodeTree
    Optimization.optimize(main_module.getOutputFilename())
  File "/home/bksahu/Nuitka/nuitka/optimizations/Optimization.py", line 545, in optimize
    finished = makeOptimizationPass(initial_pass=False)
  File "/home/bksahu/Nuitka/nuitka/optimizations/Optimization.py", line 456, in makeOptimizationPass
    changed = optimizeModule(current_module)
  File "/home/bksahu/Nuitka/nuitka/optimizations/Optimization.py", line 167, in optimizeModule
    changed = optimizeCompiledPythonModule(module)
  File "/home/bksahu/Nuitka/nuitka/optimizations/Optimization.py", line 100, in optimizeCompiledPythonModule
    module.computeModule()
  File "/home/bksahu/Nuitka/nuitka/nodes/ModuleNodes.py", line 461, in computeModule
    trace_collection=self.trace_collection
  File "/home/bksahu/Nuitka/nuitka/nodes/StatementNodes.py", line 165, in computeStatementsSequence
    new_statement = statement.computeStatementsSequence(trace_collection)
  File "/home/bksahu/Nuitka/nuitka/nodes/FrameNodes.py", line 182, in computeStatementsSequence
    new_statement = trace_collection.onStatement(statement=statement)
  File "/home/bksahu/Nuitka/nuitka/optimizations/TraceCollections.py", line 550, in onStatement
    new_statement, change_tags, change_desc = statement.computeStatement(self)
  File "/home/bksahu/Nuitka/nuitka/nodes/StatementNodes.py", line 229, in computeStatement
    trace_collection.onExpression(expression=self.getExpression())
  File "/home/bksahu/Nuitka/nuitka/optimizations/TraceCollections.py", line 531, in onExpression
    r = expression.computeExpressionRaw(trace_collection=self)
  File "/home/bksahu/Nuitka/nuitka/nodes/ExpressionBases.py", line 1038, in computeExpressionRaw
    expression = trace_collection.onExpression(sub_expression)
  File "/home/bksahu/Nuitka/nuitka/optimizations/TraceCollections.py", line 531, in onExpression
    r = expression.computeExpressionRaw(trace_collection=self)
  File "/home/bksahu/Nuitka/nuitka/nodes/ExpressionBases.py", line 1038, in computeExpressionRaw
    expression = trace_collection.onExpression(sub_expression)
  File "/home/bksahu/Nuitka/nuitka/optimizations/TraceCollections.py", line 531, in onExpression
    r = expression.computeExpressionRaw(trace_collection=self)
  File "/home/bksahu/Nuitka/nuitka/nodes/ExpressionBases.py", line 1057, in computeExpressionRaw
    return self.computeExpression(trace_collection=trace_collection)
  File "/home/bksahu/Nuitka/nuitka/nodes/BuiltinAnyNodes.py", line 46, in computeExpression
    any_node=self, trace_collection=trace_collection
  File "/home/bksahu/Nuitka/nuitka/nodes/ConstantRefNodes.py", line 383, in computeExpressionAny
    truth_value = value.getIterationValue(i).getTruthValue()
  File "/home/bksahu/Nuitka/nuitka/nodes/ConstantRefNodes.py", line 208, in getIterationValue
    constant=self.constant[count], source_ref=self.source_ref
TypeError: 'set' object does not support indexing
```

**2.** Script: `dict.py`

```
print(any({1:"One", 2:"Two"}))
```
Executed:
`$ ~/Nuitka/bin/nuitka3-run --verbose dict.py `

Output:
```
Nuitka:DEBUG:dict.py:1 : new_expression : Using original '__file__' value.
Nuitka:DEBUG:dict.py:1 : new_builtin_ref : Module variable 'print' found to be built-in reference.
Nuitka:DEBUG:dict.py:1 : new_builtin_ref : Module variable 'any' found to be built-in reference.
Nuitka:DEBUG:dict.py:1 : new_builtin : Replaced call to built-in 'any' with built-in call 'EXPRESSION_BUILTIN_ANY'.
Problem with statement at dict.py:1:
-> print(any({1:"One", 2:"Two"}))

Nuitka:INFO:Interrupted while working on '<Node 'PYTHON_MAIN_MODULE' with {'filename': 'dict.py', 'main_added': False, 'mode': 'compiled'}>'.
  File "/home/bksahu/Nuitka/nuitka/__main__.py", line 184, in <module>
    main()
  File "/home/bksahu/Nuitka/nuitka/__main__.py", line 177, in main
    MainControl.main()
  File "/home/bksahu/Nuitka/nuitka/MainControl.py", line 745, in main
    main_module = createNodeTree(filename=filename)
  File "/home/bksahu/Nuitka/nuitka/MainControl.py", line 139, in createNodeTree
    Optimization.optimize(main_module.getOutputFilename())
  File "/home/bksahu/Nuitka/nuitka/optimizations/Optimization.py", line 545, in optimize
    finished = makeOptimizationPass(initial_pass=False)
  File "/home/bksahu/Nuitka/nuitka/optimizations/Optimization.py", line 456, in makeOptimizationPass
    changed = optimizeModule(current_module)
  File "/home/bksahu/Nuitka/nuitka/optimizations/Optimization.py", line 167, in optimizeModule
    changed = optimizeCompiledPythonModule(module)
  File "/home/bksahu/Nuitka/nuitka/optimizations/Optimization.py", line 100, in optimizeCompiledPythonModule
    module.computeModule()
  File "/home/bksahu/Nuitka/nuitka/nodes/ModuleNodes.py", line 461, in computeModule
    trace_collection=self.trace_collection
  File "/home/bksahu/Nuitka/nuitka/nodes/StatementNodes.py", line 165, in computeStatementsSequence
    new_statement = statement.computeStatementsSequence(trace_collection)
  File "/home/bksahu/Nuitka/nuitka/nodes/FrameNodes.py", line 182, in computeStatementsSequence
    new_statement = trace_collection.onStatement(statement=statement)
  File "/home/bksahu/Nuitka/nuitka/optimizations/TraceCollections.py", line 550, in onStatement
    new_statement, change_tags, change_desc = statement.computeStatement(self)
  File "/home/bksahu/Nuitka/nuitka/nodes/StatementNodes.py", line 229, in computeStatement
    trace_collection.onExpression(expression=self.getExpression())
  File "/home/bksahu/Nuitka/nuitka/optimizations/TraceCollections.py", line 531, in onExpression
    r = expression.computeExpressionRaw(trace_collection=self)
  File "/home/bksahu/Nuitka/nuitka/nodes/ExpressionBases.py", line 1038, in computeExpressionRaw
    expression = trace_collection.onExpression(sub_expression)
  File "/home/bksahu/Nuitka/nuitka/optimizations/TraceCollections.py", line 531, in onExpression
    r = expression.computeExpressionRaw(trace_collection=self)
  File "/home/bksahu/Nuitka/nuitka/nodes/ExpressionBases.py", line 1038, in computeExpressionRaw
    expression = trace_collection.onExpression(sub_expression)
  File "/home/bksahu/Nuitka/nuitka/optimizations/TraceCollections.py", line 531, in onExpression
    r = expression.computeExpressionRaw(trace_collection=self)
  File "/home/bksahu/Nuitka/nuitka/nodes/ExpressionBases.py", line 1057, in computeExpressionRaw
    return self.computeExpression(trace_collection=trace_collection)
  File "/home/bksahu/Nuitka/nuitka/nodes/BuiltinAnyNodes.py", line 46, in computeExpression
    any_node=self, trace_collection=trace_collection
  File "/home/bksahu/Nuitka/nuitka/nodes/ConstantRefNodes.py", line 383, in computeExpressionAny
    truth_value = value.getIterationValue(i).getTruthValue()
  File "/home/bksahu/Nuitka/nuitka/nodes/ConstantRefNodes.py", line 208, in getIterationValue
    constant=self.constant[count], source_ref=self.source_ref
KeyError: 0
```

**python -m nuitka --version**
0.6.3rc3 (Branch: optimize_buildin_any  #246)
Python: 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
Executable: ~/anaconda3/envs/nuitka/bin/python3
OS: Linux
Arch: x86_64

bksahu:
### What does this PR do?

**Enhancements**
- [x] Any builtin Node added
- [x] Specialized optimizing algorithm
- [x] Tests for Any built-in

**Fixes**
- [x] #349 

- optimizes the `any` built-in from #232   
- fixes #349 

### PR Checklist


nuitka/nodes/BuiltinAnyNodes.py
```diff
@@ -0,0 +1,56 @@
+#     Copyright 2019, Kay Hayen, mailto:kay.hayen@gmail.com
```
kayhayen: I invite you to put your name there.

nuitka/nodes/BuiltinAnyNodes.py
```diff
@@ -0,0 +1,56 @@
...
+from .ExpressionBases import ExpressionBuiltinSingleArgBase
+from .shapes.BuiltinTypeShapes import ShapeTypeBool
+
+
+class ExpressionBuiltinAny(ExpressionBuiltinSingleArgBase):
+    kind = "EXPRESSION_BUILTIN_ANY"
+
+    builtin_spec = BuiltinParameterSpecs.builtin_any_spec
+
+    def getIntegerValue(self):
```
kayhayen: This function `getIntegerValue` is tasked to say what the node would be should it be considered as an integer. For `any` it ought to be removed.

I am nearly finished on pushing API documentation to the public. The base class has a documentation for the methods to overload.

nuitka/specs/BuiltinParameterSpecs.py
```diff
@@ -267,6 +267,7 @@ def makeBuiltinExceptionParameterSpec(exception_name):
 builtin_len_spec = BuiltinParameterSpecNoKeywords("len", ("object",), 0)
 builtin_dict_spec = BuiltinParameterSpec("dict", (), 0, "list_args", "dict_args")
 builtin_len_spec = BuiltinParameterSpecNoKeywords("len", ("object",), 0)
+builtin_any_spec = BuiltinParameterSpecNoKeywords("any", ("object",), 0)
```
kayhayen: The "any" is special in that is a function that takes varargs.

nuitka/nodes/BuiltinAnyNodes.py
bksahu: Thank you so much. API will be really helpful.

nuitka/specs/BuiltinParameterSpecs.py
bksahu: I am little confused. I read the documentation for the `len` and `any` and found that they takes similar arguments.

nuitka/nodes/BuiltinAnyNodes.py
```diff
@@ -0,0 +1,56 @@
...
+        value = self.getValue()
+
+        if value.hasShapeSlotAny():
+            return value.getIterationLength()
+        else:
+            return None
+
+    def computeExpression(self, trace_collection):
+        return self.getValue().computeExpressionAny(
```
kayhayen: The `computeExpressionAny` would be what is missing. You can see something similar at:

https://github.com/Nuitka/Nuitka/blob/cb0e0beb72041adeab1065306d5413a6bd868611/nuitka/nodes/ExpressionBases.py#L345

kayhayen:
Thanks for starting off right away, I will follow up on your question regarding how one can see the effect, later today, also with hints about testing specifically the built-in.

nuitka/specs/BuiltinParameterSpecs.py
kayhayen: You are correct @bksahu and I was wrong, which I realized and changed the comment to say "not true:".

Somehow I was thinking of `any(1,2,3)` being a thing, but it is rather `any([1,2,3])` or the `any(x for x in something)`.

kayhayen:
The API documentation, which is not current automatically updated (but then again, it's not like it's changing every minute) is here:

http://nuitka.net/apidoc/html/classnuitka_1_1nodes_1_1ExpressionBases_1_1ExpressionBase.html

There is a couple of flaws in there, mostly because it becomes visible for the first time. For instance, top level "namespaces" contains only `nuitka.containers.odict` which is non-sense. But other things are good. The page I linked above is actually pretty insightful.

I will be aiming at making answers into API documentation improvements as well, but for now we only got what is there. If you find an undocumented virtual method, it is extremely like to be documented in the base class what it is supposed to do. The doc is not duplicated into every derived class. Not sure if it should be too, but I totally see how that is less friendly to you as a reader of the `len` built-in node.

bksahu:
I would also like to contribute to the docs. How can I do so?

kayhayen:
@bksahu so we are using doxygen for the API docs, and that is easy to do yourself. Obviously you need to install doxygen, then top level in the Nuitka repository, you execute:

`doxygen doc/Doxyfile`

And there is a `html` folder that you can browse and open the `index.html` in your web browser.

The doc is influenced by the function Python docstrings mainly of course. I am currently unsure how C++ documentation should look like, what it expects there. But the Python side is generally the one that needs it most.

nuitka/nodes/BuiltinAnyNodes.py
kayhayen: Actually I was sort of misleading you here. The slot concept in Nuitka is the one of Python. There are slots for `__float__`, `__int__`, and there is `__len__` for types to overload. For any, I do not think that is the case necessarily, although I would have to check. 

My suspect is that is uses mere iteration. We would have to check the implementation in Python. I am going to send you a link for what code to look at, once I made a move of a repository not currently under Nuitka organisation, although it should be. So allow a few hours for that.

kayhayen:
I am currently considering to go black formatted code, checkout #138 where I have been discussing this. It touches a lot of code, and will cause merge conflicts with existing PRs.

Therefore, I would offer you to merge this one, disabling the extractor link much like with I did with the `len` example I gave in the GSoC issue. And only then would I apply black, and you would resume.

For that it probably would be a good idea for you to push all you have so far to this branch, should you have been working some more. What you got so far is perfect, apart of the missing `BUILTIN_ANY` C code, and the `computeExpression` not being all that correct yet. Let me know @bksahu when you are ready for the `black` re-format.

Goal is to eliminate manual formatting of code entirely, and with many PRs from many people going to come, I think that is necessary and must be done ASAP indeed. I definitely want you to not suffer anything at all from that though.

kayhayen:
Ah, and one more thing, in `tests/basics/BuiltinsTest.py` you find examples of optimizable and non-optimizable uses of Python built-ins. For `any` you would add tests. The reason CI is saying it's OK, is because we have no use of `any` in the test codes.

You can execute that test by running `tests/basics/run_all.py search BuiltinsTest` which will compare CPython and Nuitka outputs. Therefore a lot of `print` is going on in these tests.

bksahu:
@kayhayden Please give me some time. I will push the remaining code ASAP.

bksahu:
@kayhayen I have pushed the C code. And as per my understanding, `any` should be doing what `len` is doing i.e checking the input value.

nuitka/nodes/ExpressionBases.py
```diff
@@ -385,7 +385,42 @@ def computeExpressionLen(self, len_node, trace_collection):
...
+        has_any = shape.hasShapeSlotAny()
+
+        if has_any is False:
+            return makeRaiseTypeErrorExceptionReplacementFromTemplateAndValue(
+                template      = "object of type '%s' has no any()",
+                operation     = "any",
+                original_node = any_node,
+                value_node    = self
+            )
+        elif has_any is True:
```
bksahu: @kayhayen I'm not sure about this part. Could you please give me some hints?

bksahu:
@kayhayen What do you think about rather than slots for every method (eg. `hasShapeSlotAny` and `hasShapeSlotLen`) we can have slots for input types (eg. `hasShapeIterable`)? This way we can reduce the manual work.

nuitka/nodes/BuiltinAnyNodes.py
bksahu: @kayhayen You are right. Both `len` and `any` use iteration as input. Still I have created the slot for `any`.

bksahu:
Compiling with Nuitka on this script with verbose flag : 
```
def f():
    x = something()
    return any(x)
```

Gives the following output :
```
Nuitka:DEBUG:any_test.py:1 : new_expression : Using original '__file__' value.
Nuitka:DEBUG:any_test.py:3 : new_builtin_ref : Module variable 'any' found to be built-in reference.
Nuitka:DEBUG:any_test.py:3 : new_builtin : Replaced call to built-in 'any' with built-in call 'EXPRESSION_BUILTIN_ANY'.
```

kayhayen:
Very well, I was updating the Developer_Manual with an explanation of what slots are in Python and how they are represented in Nuitka. It's currently on factory branch, but I will soon make a new pre-release with blacked source code and enhanced manual.

Getting black and pylint to play along is sort of a struggle. Please read what I wrote there and let me know if you think it helps. I was offline for a couple of hours yesterday and used the time to write up what I thought is missing:

https://github.com/Nuitka/Nuitka/blob/factory/Developer_Manual.rst#python-slots-in-optimization

I sort of try to add what we do. Also added instructions for the API doc how to create that. The next chapter is also new, and then this one too:

https://github.com/Nuitka/Nuitka/blob/factory/Developer_Manual.rst#built-in-call-optimization

Although I stopped short there. But this is going to be information that you mostly have. What is missing there is how you deduce what a built-in does. I will later send you links like these ones:

https://github.com/Nuitka/Nuitka-references/blob/471b125f6ab586e91d2e200667a62345ac0eb58a/final/Python-3.7.2/Python/bltinmodule.c#L1580

and

https://github.com/Nuitka/Nuitka-references/blob/471b125f6ab586e91d2e200667a62345ac0eb58a/final/Python-3.7.2/Python/bltinmodule.c#L411
https://github.com/Nuitka/Nuitka-references/blob/471b125f6ab586e91d2e200667a62345ac0eb58a/final/python2.7-2.7.6/Python/bltinmodule.c#L127

These are CPython API uses, that you will have to understand. You can read about those here:

https://docs.python.org/2/c-api/index.html or https://docs.python.org/3/c-api/index.html

I would give you an introduction into the critical part for us in a tele session. More about that in another note.

kayhayen:
I meant to say, I recommend you have your own checkout of that references repo too. It is missing a `README` and contains very obsolete PyPy (in fact I never looked at it after I added it, but that it's probably wrong) and has a few inconsistencies.

Also I would like to describe how it's meant to be used, but yeah, until the other day, it was still private, but I think it's very useful to not have to search for pointers to make all over the place, and makes it easy to diff things, `git grep` on multiple versions, etc.

kayhayen:
@bksahu You can try and resolve the conflicts from blackened code change, but I can do it for you too, as this is fairly unusual and I have done this before. It should be relatively easy. 

I need and want to make pre-commit hooks that execute automatic Python and C code formatting at the moment of commit, but they are not there yet. Basically `black nuitka` is what should be done. I have excluded tests for now.

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -541,6 +541,43 @@ PyObject *BUILTIN_LEN(PyObject *value) {
...
+        return NULL;
+    iternext = *Py_TYPE(it)->tp_iternext;
+
+    for (;;) {
+        item = iternext(it);
+        if (item == NULL)
+            break;
+        cmp = PyObject_IsTrue(item);
+        Py_DECREF(item);
+        if (cmp < 0) {
```
kayhayen: For cases when an exception is raised, i.e. typically `NULL` returns, the condition is unlikely true. For branch prediction it seems the compilers enjoy a hint about it, and it's also for readability useful, check out the many uses of `unlikely` and sometimes `likely`. If we expect that a branch is practically never taken, we should be able to tell from the source immediately. Therefore `if (unlikely(...)))` is also information.

nuitka/nodes/shapes/BuiltinTypeShapes.py
```diff
@@ -233,6 +241,10 @@ def getTypeName():
     def hasShapeSlotLen():
         return False
 
+    @staticmethod
```
kayhayen: There is no such "slot". Any will work for `hasShapeSlotIter()`.

kayhayen:
So, you should after this add tests to `tests/basics/BuiltinsTest.py`, where you do not have to go for a lot of coverage. Later we will check, and most likely the CPython test suite has good coverage of `any` for at least the C side of things (most often static optimization is not happening for that code).

I suggest put `any(float)` as an example of something that ought to raise and be detected statically, this will not have the iteration slot, and raise in your `ExpressionBuiltinAny` node. Currently with the wrong exception text, but still. If you print out the exception and make it match, it will be a step ahead.

Then `any(range(20))` would be an example of something that ought to be predicted at run time, but may currently go to C, no big deal though. Then maybe do `any([0] * 20000)` and make sure, we are not doing something 20000 times at compile time, i.e. we do not investigate all predicted members past a certain limit. And then put something, where Nuitka doesn't understand the list currently (I think there must already be examples in that file) and will then have to use the C layer, and in one instance come to a false result, to a true result in another.

We can trust `MAKE_ITER` well enough, and your use of `iternextfunc` (btw, use that as the type for the tp_iternext value, that's how Python declares it) to not worthy of any deeper tests.

Getting that large predictable value right is kind of the challenge, give it a try, but tell me once you will (likely) despair, as that is not even clear in my mind right now.

nuitka/nodes/shapes/BuiltinTypeShapes.py
bksahu: Is it because there is no `__any__ ` slot or because it also take iterable as a input?

bksahu:
One part I'm confused is about this method. I am not sure whether it for the input shape or output shape. My best guess is that it is for output. Also, how do I debug this to see the intermediate values that each method are returning? 

https://github.com/Nuitka/Nuitka/blob/a9f4d5cc85ad6ff26513ca4ef8c5ec24f7f7247f/nuitka/nodes/BuiltinLenNodes.py#L51-L52

kayhayen:
So it's the output of the node, i.e. a description of the type it's going to have what `getTypeShape` is supposed to return.

With intermediate values, you mean what `computeExpression` does, right. In its return value, it gives a description of what it did, either as a string, or as a function if it's too involved to create it. With the command line option `--verbose`, it will then output things like this

> Nuitka:DEBUG:MiniIntExtract.py:8 : new_expression : Replaced negated comparison 'Lt' with inverse comparison 'GtE'.

For `getTypeShape()` results, I typically look at the call site that interests me and print it, so I know.

Right now, the type shape is not in the XML produced with --xml, but I think I would appreciate if it were there, because for real tests of optimization, we may want to check variable types, expression types, and so on, but we currently don't have it. It wouldn't be necessary on your back to produce it right now. I think that "pdb" or "print-style" debugging will suffice here.

Does this address your question @bksahu ?

bksahu:
By intermediate values I meant, suppose I want to import and create an instance of `ExpressionBase` class from print the return value `getTypeShape` method. I have tried "pdb" and "print-style" but I'm getting `ImportError: attempted relative import with no known parent package`

kayhayen:
@bksahu the issue you are facing is on a Python level that `nuitka.nodes.BuiltinAnyNodes` is doing module local imports. So if you import it as `BuiltinAnyNodes`, and an import wants to go from `.` or even `..`, that will then raise the Python error you are seeing. Make sure to import it as its full name.

But then, type shapes are kind of often created by the result of tracing. So the `trace_collection` that enters the `computeExpression` is carrying the state. For your node, as it's independent of the context, always the same value, probably less interesting.

Therefore I recommend you to have some kind of `Mini.py´ with a function that you believe will show your node in the state, and then to use `--verbose` to get there, and place `print` or whatever and see the flow over recurrent visits.

Like I said, currently we do not dump this information in the XML, and that's a shame. Once I am done with the organizational preparations, I am going to add that though, so it becomes visible, and testable.

Right now we only have one test that checks static optimization, it's called `tests/basics/Operations.py` and is used only by XML and expects full static optimization, but we don't have that yet for `any´, so there won't be much point right now.

Checking the bool nature, is something, you can do with `assert type(any(...))` is bool, but that is not a lot yet.

But for now, Nuitka kind of sucks with unit level testing, at least for the nodes.

bksahu:
@kayhayen. Thank you so much. This is what I was looking for. Could you please review the code. I have update the files and also added tests. There is only one test that I am unable to get it right i.e. `any([0] * 20000)`

kayhayen:
@bksahu that test is probably going to be toughest. I see how it's not easy to put the pieces together, but I will try and modify the API documentation a bit to clear things up. I am also proposing to have a joint session, for an existing one, so we can debug a bit together, and see how the pieces are supposed to work.

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -541,6 +541,50 @@ PyObject *BUILTIN_LEN(PyObject *value) {
...
+        Py_DECREF(item);
+        if (unlikely(cmp < 0)) {
+            Py_DECREF(it);
+            return NULL;
+        }
+        if (cmp > 0) {
+            Py_DECREF(it);
+            Py_INCREF(Py_True);
+            return Py_True;
+        } else {
```
bksahu: @kayhayen This fixes `print(any([0] * 20000))` tests but breaks `S = [40, 60, 30]
print(any(x > 42 for x in S))`

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -541,6 +541,50 @@ PyObject *BUILTIN_LEN(PyObject *value) {
...
+            Py_DECREF(it);
+            Py_INCREF(Py_False);
+            return Py_False;
+        }
```
bksahu: @kayhayen This else part fixes `print(any([0] * 20000))` but fails `S = [40, 60, 30]
print(any(x > 42 for x in S))`

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -541,6 +541,44 @@ PyObject *BUILTIN_LEN(PyObject *value) {
...
+            return NULL;
+        }
+        if (cmp > 0) {
+            Py_DECREF(it);
+            Py_INCREF(Py_True);
+            return Py_True;
+        }
+    }
+    Py_DECREF(it);
+    if (CHECK_AND_CLEAR_STOP_ITERATION_OCCURRED())
```
kayhayen: Need to checkout what this function `CHECK_AND_CLEAR_STOP_ITERATION_OCCURRED` does.

 believe the handling of the return value is definitely incorrect. Ok, the documentation of it is really not telling a lot, so please extract and add a better one from this.

It returns true, if no error occured. In CPython, the `iternext` is allowed to return `NULL` with no error set, and that means end of iteration. Alternatively, the CPython exception can be set as `StopIteration`, which would be wasteful, but allowed. In these 2 cases it returns true, and the loop is properly terminated. I guess, in this case, we didn't find a true value from `PyObject_IsTrue`, so you ought to return `Py_False ` in that case.

When it returns false, the loop iteration aborted with another error, but `StopIteration`, e.g. `KeyError` of the `next` of an iteration, that must be returned. So `return FALSE` should be done.

And actually, removing the `PyErr_Clear()` that ought to be harmless (clearing no exception, because none will the set at that point).

BTW, loop exit without exception should be considered unlikely, therefore I would prefer a condition like this:

```
          if (unlikely(!CHECK_AND_CLEAR_STOP_ITERATION_OCCURRED())) {
              return NULL;
         }
```

Just to make it clear, that this is going to be rare.

nuitka/nodes/ExpressionBases.py
```diff
@@ -800,6 +843,14 @@ def computeExpressionLen(self, len_node, trace_collection):
 Compile time constant len value pre-computed.""",
         )
 
+    def computeExpressionAny(self, any_node, trace_collection):
```
kayhayen: Should be unused now, right, remove those

nuitka/nodes/shapes/StandardShapes.py
```diff
@@ -54,6 +54,10 @@ def hasShapeSlotComplex():
     def hasShapeSlotLen():
         return None
 
+    @staticmethod
```
kayhayen: Same, should be unused, and wrong to have

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -541,6 +541,44 @@ PyObject *BUILTIN_LEN(PyObject *value) {
     return PyInt_FromSsize_t(res);
 }
 
+PyObject *BUILTIN_ANY(PyObject *value) {
+    CHECK_OBJECT(value);
+
+    PyObject *it = PyObject_GetIter(value);
+    PyObject *(*iternext)(PyObject *);
```
kayhayen: This ought to be `iternextfunc` and push down declarations to their first use, i.e. `iternextfunc = Py_TYPE(it)->tp_iternext"

kayhayen:
One thing, I meant to highlight.

You see, understanding the C part, and the C algorithm used in CPython is absolutely crucial. Not only do we have to replicate it exactly in C, @bksahu, but also in Python as abstract code execution.

The iteration has a representation in the `ExpressionBuiltinAny.computeExpression`, with much the same handling of early exits becoming optimizations, and the final nothing found, becoming a result. But, and this is an important but, you always need to factor in "unknown" as a value too.

So, your version of `any` needed to also consider 'True', 'False', and 'None' (unknown truth value), and therefore becomes more complex.

bksahu:
@kayhayen How can I benchmark our `any()` implementation against the CPython's?

bksahu:
@kayhayen I'm going to improve the API docs in another PR.

bksahu:
> One thing, I meant to highlight.
> 
> You see, understanding the C part, and the C algorithm used in CPython is absolutely crucial. Not only do we have to replicate it exactly in C, @bksahu, but also in Python as abstract code execution.
> 
> The iteration has a representation in the `ExpressionBuiltinAny.computeExpression`, with much the same handling of early exits becoming optimizations, and the final nothing found, becoming a result. But, and this is an important but, you always need to factor in "unknown" as a value too.
> 
> So, your version of `any` needed to also consider 'True', 'False', and 'None' (unknown truth value), and therefore becomes more complex.

@kayhayen So if there is no optimiaztion during compile time or it fails somehow then is the final result (i.e the "unknown factor) is computed by the C part like normaly CPython would? And is that why we have to replicate the C part of the CPython?

nuitka/build/include/nuitka/exceptions.h
```diff
@@ -441,8 +441,16 @@ NUITKA_MAY_BE_UNUSED static inline void ADD_EXCEPTION_CONTEXT(PyObject **excepti
 }
 #endif
 
-// Special helper that checks for StopIteration and if so clears it, only
-// indicating if it was set.
+/* Special helper that checks for StopIteration and if so clears it, only
```
kayhayen: Very well. Just for my information, did you check how these things look in Doxygen? To my shame I didn't look at the C part at all yet. I was wondering where Doxygen expects comments to pick them up. So far Nuitka has followed no rules.

nuitka/build/include/nuitka/helpers.h
```diff
@@ -247,6 +247,9 @@ extern PyObject *BUILTIN_TYPE3(PyObject *module_name, PyObject *name, PyObject *
 // For built-in built-in len() functionality.
 extern PyObject *BUILTIN_LEN(PyObject *boundary);
 
+// For built-in built-in any() functionality.
+extern PyObject *BUILTIN_ANY(PyObject *boundary);
```
kayhayen: I don't think you ought to use "boundary" here. Not even sure it was a good name for `BUILTIN_LEN` then.

kayhayen: You ought to match declaration and definition, I think `valuè` is fine indeed.

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -541,6 +541,42 @@ PyObject *BUILTIN_LEN(PyObject *value) {
     return PyInt_FromSsize_t(res);
 }
 
+PyObject *BUILTIN_ANY(PyObject *value) {
+    CHECK_OBJECT(value);
+
+    PyObject *it = PyObject_GetIter(value);
+    PyObject *(*iternextfunc)(PyObject *) = *Py_TYPE(it)->tp_iternext;;
+    PyObject *item;
```
kayhayen: Move the declaration of item to where it is used. Unlike CPython, we do C99, so this can be at any point, but esp. inside the loop. Bonus effect, using it outside of the loop, which would be likely a bug symptom will be caught.

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -541,6 +541,42 @@ PyObject *BUILTIN_LEN(PyObject *value) {
     return PyInt_FromSsize_t(res);
 }
 
+PyObject *BUILTIN_ANY(PyObject *value) {
+    CHECK_OBJECT(value);
+
+    PyObject *it = PyObject_GetIter(value);
+    PyObject *(*iternextfunc)(PyObject *) = *Py_TYPE(it)->tp_iternext;;
```
kayhayen: I think the Python header object.h defines `iternextfunc`, and I recommend using it like `iternextfunc iternext = Py_TYPE(it)->tp_iternext;` (btw there is two `;` there in your code).

kayhayen: And besides, in similar vein to my next line comment, move this done to where it is actually used, that would be right before the loop.

Here you call `GetIter` and then assign around some pointers, and only then use the value. The C compiler will be unsure if it doesn't have to segfault before you return and actually do it early, only to not use it. It definitely is easier to read, although it is probably more distant to the CPython code (does it do it that way, I doubt though), but that does not matter, just keep variable declarations and assignments together, and as late as possible please.

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -541,6 +541,42 @@ PyObject *BUILTIN_LEN(PyObject *value) {
...
+            return NULL;
+        }
+        if (cmp > 0) {
+            Py_DECREF(it);
+            Py_INCREF(Py_True);
+            return Py_True;
+        }
+    }
+    Py_DECREF(it);
+    if (unlikely(!CHECK_AND_CLEAR_STOP_ITERATION_OCCURRED()))
```
kayhayen: An important style preference, please always use blocks with if statements. The `if () {}` is such a huge bug avoider, and readability aid, it is worth the extra characters.

kayhayen:
@bksahu I am reversing my stance on an important point. Please read carefully and try to not be confused. :-)

I am starting to like `computeExpressionAny`. I originally thought that the `any` argument has little to contribute. But when discussing the more difficult cases, like a `[0] * 10000` I realized, that the generic algorithm may have to specialized. In that instance that `*` responsible node (will you find it? :-) could very well know that repeating values doesn't change the `any` (or `all) value for that matter. And for `any` on compile time constants, we could and should potentially specialize. 

Although I am highly concerned, if `all(range(1,100000))` in Python3 will not be still slow, and needs to not be done like this, but most built-in types are constrained to small iteration sizes and won't notice, whereas the generic algorithm will e.g. choke over iteration length 100000, and yet, the `range` could also specialize `any` on it, basically it would be `iteration length > 1` or that single value is not `0`.

These are advanced topics, but best solved in a method to overload by the argument.

Also, in many instances, the "control flow escape" and exception part is not true. For any(`[x] * 10000`) there would only be control flow escape when `x` has a type known to do that for `bool(x)` which many times could be decided. So that means, there is a high probability that the generic implementation is not going to cut all cases, and might be over the edge in others.

Therefore you convinced me now, that `computeExpressionAny` is the way to go even though it's not a slot. And in fact, `computeExpressionSum` ought to exist for similar reasons.

kayhayen:
For the PyLint issue of trailing whitespace, I recommend you run `bin/autoformat-nuitka-source` on the source code. And definitely head the advice as for installed Plugins as in Developer Manual. The AnyEdit plugin seems necessary to prevent trailing whitespace in Eclipse and PyDev. The others are just unused imports. Eclipse normally marks these as yellow. I forgot how it's named, but the map on the right, that you can enable in PyDev I think, will show yellow and red markers for issues PyDev finds, and this can be very helpful and avoid roundtrips with Travis.

Also `bin/check-nuitka-with-pylint` is recommended before you push.

I believe the Python side of things looks good, and my C comments are not too hard, so I think we might merge this soonish if you fix all the things timely.

nuitka/build/include/nuitka/exceptions.h
bksahu: I will look into it.

nuitka/build/include/nuitka/helpers.h
bksahu: Looks like length declaration and definition of `BUILTIN_LEN` is also different.

nuitka/build/static_src/CompiledCodeHelpers.c
bksahu: > The C compiler will be unsure if it doesn't have to segfault before you return and actually do it early, only to not use it.

Okay sure I will keep variable declarations and assignments together but I don't get it how declaring at once is going to give me a segfault?

bksahu:
> @bksahu I am reversing my stance on an important point. Please read carefully and try to not be confused. :-)
> 
> I am starting to like `computeExpressionAny`. I originally thought that the `any` argument has little to contribute. But when discussing the more difficult cases, like a `[0] * 10000` I realized, that the generic algorithm may have to specialized. In that instance that `*` responsible node (will you find it? :-) could very well know that repeating values doesn't change the `any` (or `all) value for that matter. And for `any` on compile time constants, we could and should potentially specialize.
> 
> Although I am highly concerned, if `all(range(1,100000))` in Python3 will not be still slow, and needs to not be done like this, but most built-in types are constrained to small iteration sizes and won't notice, whereas the generic algorithm will e.g. choke over iteration length 100000, and yet, the `range` could also specialize `any` on it, basically it would be `iteration length > 1` or that single value is not `0`.
> 
> These are advanced topics, but best solved in a method to overload by the argument.
> 
> Also, in many instances, the "control flow escape" and exception part is not true. For any(`[x] * 10000`) there would only be control flow escape when `x` has a type known to do that for `bool(x)` which many times could be decided. So that means, there is a high probability that the generic implementation is not going to cut all cases, and might be over the edge in others.
> 
> Therefore you convinced me now, that `computeExpressionAny` is the way to go even though it's not a slot. And in fact, `computeExpressionSum` ought to exist for similar reasons.

Okay so are you interested to do constant folding using `nuitka.nodes.NodeMakingHelpers`?

nuitka/build/include/nuitka/exceptions.h
bksahu: It works fine. :smile:

bksahu:
>  I realized, that the generic algorithm may have to specialized. In that instance that * responsible node (will you find it? :-)

Is it `ConstantRefNodes` ?

> I realized, that the generic algorithm may have to specialized. 

Not sure how can we do that.

@kayhayen I think we are ready to merge this PR.

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -541,6 +541,43 @@ PyObject *BUILTIN_LEN(PyObject *value) {
...
+    CHECK_OBJECT(value);
+
+    PyObject *it = PyObject_GetIter(value);
+
+    if (unlikely((it == NULL))) {
+    	return NULL;
+    }
+
+
+    int cmp;
```
kayhayen: The `cmp` is only used inside the loop, move this declaration it to its assigment

nuitka/nodes/BuiltinAnyNodes.py
```diff
@@ -0,0 +1,67 @@
...
+from nuitka.specs import BuiltinParameterSpecs
+
+from .ExpressionBases import ExpressionBuiltinSingleArgBase
+from .shapes.BuiltinTypeShapes import ShapeTypeBool
+
+
+class ExpressionBuiltinAny(ExpressionBuiltinSingleArgBase):
+    """ Builtin Any Node class.
+
+    Parameter
```
kayhayen: I am not sure this is the proper way to document a class. I think it has a value argument that is an expression base object, and we use `Args` as the title according to the developer manual part from Jorj.

kayhayen:
@bksahu You didn't address all the review comments, and I have 2 new ones.

As for `ConstantRefNodes` that is a module. It's classes have a base class that does things for "compile time constants". Find it, and you will see lots of overloads that take advantage of being allowed to use the constant value for a compile time computation. I want you to add a `computeExpressionAny` there too.

nuitka/nodes/BuiltinAnyNodes.py
```diff
@@ -29,14 +29,12 @@
...
-    Parameter
-    ---------
-    ExpressionBase: 'any' expression
-
-    Returns
-    -------
-    Bool: True if at least one element of an iterable is true
-          else return False
+    Args:
+        ExpressionBase: 'any' expression
```
kayhayen: According to this: http://nuitka.net/doc/developer-manual.html#api-documentation-and-guidelines

It should be 'any - ...'

nuitka/nodes/BuiltinAnyNodes.py
```diff
@@ -29,14 +29,12 @@
...
+
+    Returns:
```
kayhayen: I don't think classes should have a `Returns` ... we do not have a good example for a class, but I would say 

```
Returns:
    Node that represents built-in 'any' call.
```

nuitka/nodes/BuiltinAnyNodes.py
bksahu: Sorry, I can't find it in the docs. Do you it should be 

```
Args:
    ExpressionBase: 'any - expression'
```

nuitka/build/static_src/CompiledCodeHelpers.c
kayhayen: I didn't respond to this yet.

Well, what you do is basically in Python equivalent to this:

```
x = f()
a = y.z
if x is None:
   return None
```

Even if the C compiler might optimize the assignment to `a` away and it will, why ask the reader to also read it like that?

bksahu: Okay. So it's better to keep related expressions together for better readability.

bksahu:
@kayhayen There is a little problem with our generic algorithm.
It is failing for types that are not indexable eg. `set` and `dict` due to this line

https://github.com/Nuitka/Nuitka/blob/fbc71f767cdb95992dccd9576820ad5c059bc9cb/nuitka/nodes/ExpressionBases.py#L398

I looked into the `isIndexable()` method but it returns `False` for every type. And for `dict` type I couldn't find out a method in Nuitka for `get()`. Maybe I should write it.

As temporary work around I have setup checks for both types.

bksahu:
@kayhayen I have finally added the specialized method you asked for. 

For types that are not indexable i.e `set` and `dicts` etc what should we do with them? Should we not optimize them or should we convert them into indexable types and then optimize them?

bksahu:
@kayhayen could you please give me some pointers on the `experimental` flag. I can't figure out where to use `Options.isExperimemtal()`

kayhayen:
@bksahu I added a chapter for those on the factory branch, it explains how to use them.

bksahu:
> @bksahu I added a chapter for those on the factory branch, it explains how to use them.

@kayhayen Sorry, I can't find it. Are you yet to push it or it's already there?

bksahu:
@kayhayen I have added the `experimental` flag. As I have mentioned earlier I will be pretty much inactive for the next couple of days due to my exams. I will start working on `set` and `dict` after my exams are over.

nuitka/nodes/ConstantRefNodes.py
```diff
@@ -357,6 +357,65 @@ def computeExpressionIter1(self, iter_node, trace_collection):
...
+        iteration_length = self.getIterationLength()
+
+        if (
+            iteration_length is not None
+            and iteration_length < 256
+            and self.canPredictIterationValues()
+        ):
+            all_false = True
+            for i in range(iteration_length):
+                truth_value = self.getIterationValue(i).getTruthValue()
```
bksahu: @kayhayen I just found out that this `self` doesn't have any method named `getIterationValue()` and we got away with it because we always tested it on `iteration_length`>256. However the `self` in `ConstantRefNodes.computeExpressionAny` has such a method which actually computes the result for us. So, what do you want me to do about it?

kayhayen: Yes, I was bad there. You got to have a function indicating that ability, and then have an interface which is not in the general interface.

I think if you heed my request, these kinds of interfaces move into the new kind of handle, and are its methods rather. So, the function to return that handle would in base class return `None` meaning "I don't know", for some  a handle that says so, could be a shared re-used global variable, and from some an actual handle that can give iteration values, etc.

I sometimes do this on purpose, so if I use a method which I am not supposed to, it will crash. That is normally a sign of "still in flux". :-)

nuitka/__past__.py
```diff
@@ -111,6 +111,22 @@ def total_ordering(cls):
 else:
     intern = sys.intern  # @ReservedAssignment @UndefinedVariable
 
+
+def getMetaClassBase(meta_class_prefix):
+    """ For Python2/3 compatible source, we create a base class that has the metaclass
+    used and doesn't require making a choice.
+    """
+
+    from abc import ABCMeta
```
kayhayen: Why not have that import top level

nuitka/codegen/Contexts.py
```diff
@@ -338,7 +328,7 @@ def _calcHash(self, key):
             return hash_value.hexdigest()
 
 
-class PythonContextBase(ContextMetaClassBase):
+class PythonContextBase(getMetaClassBase("Context")):
```
kayhayen: Very nice!

tests/basics/BuiltinsTest.py
```diff
@@ -524,3 +524,32 @@ def __complex__(self):
...
+# Test for any builtins
+print(any([None, None, None]))
+print(any([None, 4, None]))
+print(any([]))
+print(any(range(20)))
+print(any([0] * 20000))
+print(any([0] * 255))
+
+S = [40, 60, 30]
+print(any(x > 42 for x in S))
```
kayhayen: Can you make it so that we can see if and when any stops iteration, i.e. could you make that iterator some that does an output.

```
def S():
    print("Yielding 40")
    yield 40
    print("Yielding 60")
    yield 60
    print("Yielding 30")
    yield 30
```

I think that will prove that any is short circuiting iteration, or is it not?

kayhayen:
So, regarding experimental flag, how about I setup nightshift.nuitka.net next, which would run test jobs of your PRs over all Python suites and so on. That is basically the hurdle for merge, maybe then we can skip the experimental part entirely.

kayhayen:
Also, I think there is a few places to add iteration handles for, e.g. `(x)range` and that would lead to another change, in what makes sense. Basically all that does `canPredictIterationValues()` should receive a handle that is not None.

bksahu:
@kayhayen If I fix this pylint error for py27 then I'm going to get many errors for py3. What should I do?

bksahu:
> Also, I think there is a few places to add iteration handles for, e.g. `(x)range` and that would lead to another change, in what makes sense. Basically all that does `canPredictIterationValues()` should receive a handle that is not None.

As every node is a child node of `ExpressionBase`, isn't it already has access to `getIterationHandle`? Do you want me to replace every `getIterationValue` with our handle?

kayhayen:
Regarding `getIterationHandle` vs `getIterationValue`, this is what I want you do do:

a) Look at every existing `getIterationValue` and consider adding an `getIterationHandle` that could later be used as a replacement. Or like you did, if we can find a way to really have neither, then add a TODO item for that.

b) By doing so, you will make sure that the new style `any` will be roughly as optimal as the older approach, where we used `getIterationValue` but had to change to the new one.

c) Later, we can make the switch, when we think we are there. For new stuff, definitely we will prefer the handle, as cleanups, gradually switch over all methods related to iteration right now.

Yours,
Kay

kayhayen:
The PyLint for Python2 on Travis is old and sometimes may have bugs. The promise to also lint Python2 code using their new Python3 code base, has not yet been fulfilled by the PyLint project yet.

tests/basics/BuiltinsTest.py
```diff
@@ -524,3 +524,39 @@ def __complex__(self):
     print("Oops, must not happen.")
 else:
     print("Oops, must not happen.")
+
+# Test for any builtins
+print(any([None, None, None]))
+print(any([None, 4, None]))
+print(any([]))
+print(any(range(20)))
```
kayhayen: Because it is so small, 20 elements only, it is optimized into a list soon. Only ranges larger than 256 elements are not optimized, so I suggest you go with 260 elements, and try all kinds of ranges, with one, two and 3 arguments, each time making sure it's more than 256 elements.

nuitka/nodes/ContainerMakingNodes.py
```diff
@@ -24,6 +24,7 @@
 from nuitka.PythonVersions import needsSetLiteralReverseInsertion
 
 from .ExpressionBases import ExpressionChildrenHavingBase
+from .IterationHandles import ListAndTupleContainerMakingIterationHandle
```
kayhayen: Did you consider putting them next to the node? I am not sure, but my suspect is their are so close friends, they probably need to be looked at as one thing very often.

nuitka/nodes/IterationHandles.py
```diff
@@ -0,0 +1,310 @@
...
+    @abstractmethod
+    def __repr__(self):
+        """Returns a printable representation of the IterationHandle
+        and it's children object."""
+        return
+
+    @abstractmethod
+    def getNextValueExpression(self):
+        """Abstract method to get next iteration value."""
+        return
```
kayhayen: Doesn't pylint hate those useless returns? Puting the docstring is enough, can leave out the return statement, unless this would be a default implementation actually called by base class manually.

nuitka/nodes/IterationHandles.py
```diff
@@ -0,0 +1,310 @@
...
+        assert constant_node.isIterableConstant()
+        self.constant = constant_node.constant
+        self.constant_node = constant_node
+        self.iter = iter(self.constant)
+
+    def __repr__(self):
+        return "<%s of %r>" % (self.__class__.__name__, self.constant_node)
+
+    def getNextValueExpression(self):
+        """Return the next iteration value or StopIteration exception
```
kayhayen: This one has a better doc string explaining the contract than the base class. It should say that None return thing too.

nuitka/nodes/IterationHandles.py
```diff
@@ -0,0 +1,310 @@
...
+        object
+    getNextValueExpression()
+        Returns the next iteration value
+    getNextValueTruth()
+        Returns the boolean value of the next handle
+    getIterationValueWithIndex(value_index)
+        Sequential access of the expression
+    """
+
+    def __init__(self, constant_node):
```
kayhayen: So this one is getting a list with the element values, instead of a node, right? So why is it named `constant_node`, that seems misleading.

nuitka/nodes/IterationHandles.py
```diff
@@ -0,0 +1,310 @@
...
+        self.iter = iter(self.constant)
+
+
+class ConstantIterationHandleRange3(ConstantRangeIterationHandleBase):
+    """Iteration handle for ranges(low, high, step)"""
+
+    def __init__(self, constant_value):
+        ConstantRangeIterationHandleBase.__init__(self, constant_value)
+        assert self.constant_value.isExpressionBuiltinRange3()
+        self.constant = range(self.low, self.high, self.step)
```
kayhayen: This is calling `range` which creates an arbitrarily large list. I think we said, you can use a constant, but it would have to be an `xrange` (the one from `nuitka.__past__`).

nuitka/nodes/IterationHandles.py
```diff
@@ -0,0 +1,310 @@
...
+        except IndexError:
+            return None
+
+    def getNextValueTruth(self):
+        """Return the boolean value of the next iteration handle."""
+        try:
+            iteration_value = next(self.iter)
+        except StopIteration:
+            return StopIteration
+        return bool(iteration_value)
```
kayhayen: The type of `iteration_value` is not what you treat it as here. It is a node, and should be asked its truth value. This is going to be `true` in this form always.

nuitka/nodes/IterationHandles.py
bksahu: Nope. Initially, I used `pass` in place of `return`, pylint2 was not happy with it.

nuitka/nodes/ContainerMakingNodes.py
bksahu: Do you mean next to `NodeMakingHelpers` ?

nuitka/nodes/IterationHandles.py
kayhayen: Yes, `pass` is also not needed.

You see, doc strings in Python are statements. A string as a statement has not much effect, and therefore, Python allows its parser to special case those strings as first statements.

But that is why `def f(): ""` is a complete function definition. Python injects a `return` at the end of each function body automatically. And the `pass` is not needed, because, well there is already a statement, that is why it complained. The fact that PyLint does not complain about the return doesn't convince me of it being good. Just remove them, and PyLint and I will both be happy.

As far as blocks go, you can do these:

```
while 1:
   ""
```

Not that are any good, just to illustrate what is happening. PyLint is probably going to hate that one though, because it's not a doc string. Statement without effect is what I would expect to be warned about.

So documenting your functions has the extra bonus of allowing to remove the `pass` that you initially likely had there.

nuitka/nodes/ContainerMakingNodes.py
kayhayen: No, more generally I think of a layout, where related things stay together.

`nuitka.nodes.IterationNode.IterationHandleOfSomething`
`nuitka.nodes.Somethings.SomethingNode`

is a bit tearing apart knowledge about `SomethingNode`. Because the iteration handle is going to be deeply tied to it, I would prefer this layout:

`nuitka.nodes.Somethings.IterationHandleOfSomething`
`nuitka.nodes.Somethings.SomethingNode`

In this, related information is dealt with in one place. The iteration handle details are private and result of being huge friends to the node type.

Only base classes, or relatively generic stuff should go into the `IterationHandles` module, and generic stuff needs no private details and is used by multiple things.

With the range nodes, I think it's very evident, that they are something I want to look at the same time as its iteration handles.

For now, leave them where they are. I just wanted to make sure you considered the question, and I want to discuss with you the distribution of responsibility between iteration handle and node separately, and that is going to affect such decisions a lot.

tests/basics/BuiltinsTest.py
bksahu: Okay. So, I have added tests with >256 elements. But I'm failing those tests because of this assert statement https://github.com/Nuitka/Nuitka/pull/246/files/ba4ad49960f2fcac76973cf6c503210e3bc7a6af#diff-bcd50821425789a3c9bfd17b9c3d16f1R229

I tried to move them into a try-catch block but that didn't help either.

bksahu: So, Instead of asserting should I return `None`? Is that a good idea?

kayhayen: Your assert looks suspicious, can it be that you are comparing a function and an integer? That is going to give pretty random results.

kayhayen: Also, why would be size matter. We took care in the `any` node to not try too many things, by aborting after 256 checks, that could be good enough, but it means the user needs to take care, but the producer of the handle, it cannot tell if that is an issue to you or not, so I think it should just represent the best possible knowledge.

kayhayen: And another thing I didn't check immediately, but are you making sure that e.g. `getLow().getIntegerValue()` is actually working and not returning `None`. I didn't yet see that, but it would definitely be needed to have `None` tests.

Also generally, I am in favor of delaying as much work as possible to until when it's needed. I was going to discuss the responsibilities of these classes with you online though. But generally, I think the `__init__` ought to take note only of what is absolutely necessary and no more. When too much is done in the constructor, it can quickly become too much overhead for what is probably a simple question.

But I also think we need to draw a line, and accept this at some point when it's working, and then cleanups come later. But correctness is not there yet fully.

bksahu: > Your assert looks suspicious, can it be that you are comparing a function and an integer? That is going to give pretty random results.

I think Assert is fine as it is giving AssertionError which is because the number of elements is greater than 256

bksahu: > And another thing I didn't check immediately, but are you making sure that e.g. getLow().getIntegerValue() is actually working and not returning None. I didn't yet see that, but it would definitely be needed to have None tests.

As the get methods are in the Base class, I have put them in try catch block which will catch if `getHigh().getIntergerValue()` gives an attribute error which it will when `getHigh()` returns a `None`.  https://github.com/Nuitka/Nuitka/pull/246/files#diff-bcd50821425789a3c9bfd17b9c3d16f1R223

kayhayen: I really, really dislike that kind of error handling. I prefer them to be checks, please change these things to explicit checks. 

E.g. will `getHigh().getIntergerValue()` hopefully always raise an `AttributeError` and that is very hard to spot (misspelling), and a failed re-factoring, etc. could also cause this. Code should never be that fragile. I could well imagine something deep behind that function to raise an `AttributeError` due to other programming mistakes.

kayhayen: And this is definitely wrong in one way or the other: `assert constant_value.getIterationLength < 10000`, either it is a coding rule violation, namely an integer attribute named like a function, or a function not being called.

I think you do this here:

```
>>> class C(object):
...    def f():
...       pass
... 
>>> C().f < 10000
False
```

The value is pretty random. I believe, it is comparing the pointer to the object that `False` is to the bound function object, and that can go either way.

Since this is a common mistake, in Python3, it gives: `TypeError: unorderable types: method() < int()` but not so in Python2.

tests/basics/BuiltinsTest.py
```diff
@@ -529,11 +529,14 @@ def __complex__(self):
...
-print(any(range(260)))
-print(any(range(1, 270)))
-print(any(range(2, 1024, 5)))
 print(any([0] * 20000))
 print(any([0] * 255))
+try:
+    print(any(range(260)))
+    print(any(range(1, 270)))
+    print(any(range(2, 1024, 5)))
+except Exception as e:
```
kayhayen: Why catch an exception, there is none that should be raised?

tests/basics/BuiltinsTest.py
kayhayen: Just to be clear, avoid using exception handling for control flow at all costs. When an exception hits, your first instinct seems to be to catch it. Instead, you need to make sure it doesn't raise.

The first approach leads to very fragile and potentially slow code that makes it hard to understand the context you are in. If things can happen in an exception handler, and mean `AttributeError` is supposed to me e.g. `something is None` that is very un-obvious.

bksahu: > And this is definitely wrong in one way or the other: `assert constant_value.getIterationLength < 10000`, either it is a coding rule violation, namely an integer attribute named like a function, or a function not being called.
> 
> I think you do this here:
> 
> ```
> >>> class C(object):
> ...    def f():
> ...       pass
> ... 
> >>> C().f < 10000
> False
> ```
> 
> The value is pretty random. I believe, it is comparing the pointer to the object that `False` is to the bound function object, and that can go either way.
> 
> Since this is a common mistake, in Python3, it gives: `TypeError: unorderable types: method() < int()` but not so in Python2.

Yes, I know that I was missing `()` after `getIterationLength` and I have already updated it. You are probably looking at an older commit.

nuitka/nodes/BuiltinRangeNodes.py
```diff
@@ -354,6 +395,8 @@ def getIterationValue(self, element_index):
 
         result = low + step * element_index
 
+        result = low + step * element_index
```
kayhayen: That seems like a mistaken change to `getIterationValue`

nuitka/nodes/ConstantRefNodes.py
```diff
@@ -828,7 +844,6 @@ def getDetails(self):
 def makeConstantRefNode(constant, source_ref, user_provided=False):
     # This is dispatching per constant value and types, every case
     # to be a return statement, pylint: disable=too-many-branches,too-many-return-statements
-
```
kayhayen: These are two separate comments, should not remove the new line between them.

nuitka/nodes/BuiltinRangeNodes.py
kayhayen: Still present?

PR #256, opened 2019-02-13
Windows Qt window normal theme

katant:
The default Qt window looks just awful. This fix adds a normal windows theme.

### What does this PR do?
Adds Qt plugin, which makes window look way better
### PR Checklist

### Before

Before

### After

After

### DIFF
--- a/nuitka/plugins/standard/PySidePyQtPlugin.py
+++ b/nuitka/plugins/standard/PySidePyQtPlugin.py
@@ -137,6 +137,7 @@ if os.path.exists(guess_path):
                 # Seems platforms is required on Windows.
                 if os.name == "nt":
                     plugin_options.add("platforms")
+                    plugin_options.add("styles") # Makes Qt window look much better, than before
 
             info(
                 "Copying Qt plug-ins '%s' to '%s'."


.travis.yml
```diff
@@ -1,5 +1,6 @@
 # Python project, check against all supported things.
 language: python
+os: osx
```
kayhayen: Is this really all it takes to enable MacOS on Travis?

nuitka/plugins/standard/PySidePyQtPlugin.py
```diff
@@ -137,6 +137,7 @@ def considerExtraDlls(self, dist_dir, module):
                 # Seems platforms is required on Windows.
                 if os.name == "nt":
                     plugin_options.add("platforms")
+                    plugin_options.add("styles") # Makes Qt window look much better, than before
```
kayhayen: So this is limited to Windows (os.name == "nt") then? Would it apply to other platforms too.

kayhayen:
Travis hates the `osx` change, although it seemed to do something (where else is the `10.3` portion of the refused path from), it is not enough to enable macOS testing.

However, the change is made Windows only, so I don't know what that means at all.

.travis.yml
katant: Oh, I commited that after pull request. This just was a test

nuitka/plugins/standard/PySidePyQtPlugin.py
katant: This folder contains qwindowsvistastyle.dll, so I think yes

PR #260, opened 2019-02-15
Fix several ResourceWarning: unclosed file

BoboTiG:
### What does this PR do?

- Fix several `ResourceWarning: unclosed file`
- Also uniformaize the use of the "with" context manager when dealing with files.

Wanted to shut down noisy warnings in my console. Don't know if you are interested in fixing them, but I try :)

### PR Checklist

Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>

### DIFF
--- a/nuitka/freezer/Standalone.py
+++ b/nuitka/freezer/Standalone.py
@@ -75,7 +75,8 @@ if Utils.isWin32Windows() and Options.isExperimental("use_pefile"):
 def loadCodeObjectData(precompiled_filename):
     # Ignoring magic numbers, etc. which we don't have to care for much as
     # CPython already checked them (would have rejected it otherwise).
-    return open(precompiled_filename, "rb").read()[8:]
+    with open(precompiled_filename, "rb") as f:
+        return f.read()[8:]
 
 
 module_names = set()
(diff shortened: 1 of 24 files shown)


nuitka/freezer/Standalone.py
```diff
@@ -753,7 +755,9 @@ def _parseDependsExeOutput(filename, result):
     inside = False
     first = False
 
-    for line in open(filename):
```
kayhayen: I think for `for line in open(filename)` were better to become something like 

```
with open(filename) as f: 
   for line in f:
```

That preserves the idea of not ever having the file in memory, which is a good idea by default. 

So I think uses of `f.readlines()` added, if you could change those instances? Otherwise looks perfect to me.

kayhayen:
This is a very good one. The Travis error is from an unfortunate branching point. I will fix develop later this evening, rebasing then should leave you cleaner.

nuitka/freezer/Standalone.py
BoboTiG: You are totally right, I did not want to change too many lines idendation. I will update the PR tomorrow (including the rebase).

BoboTiG:
This is a much bigger PR now, but mostly indentation changes :)

kayhayen:
I will merge later today. The PyLint warning ist something I can take care of.

PR #263, opened 2019-02-18
Fix imports from the collections module

BoboTiG:
### What does this PR do?

Fixes such warnings: `DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working`.

Despite the fact that it will ease the Python 3.8 compliance, it will specially remove noisy warnings :)

### PR Checklist

Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>

### DIFF
--- a/nuitka/__past__.py
+++ b/nuitka/__past__.py
@@ -94,6 +94,16 @@ except ImportError:
 
         return cls
 
+if str is bytes:
+    from collections import (  # pylint: disable=I0021,import-error,no-name-in-module
+        Iterable,  # @UnresolvedImport @UnusedImport
+        MutableSet,  # @UnresolvedImport @UnusedImport
+    )
+else:
+    from collections.abc import (  # pylint: disable=I0021,import-error,no-name-in-module
+        Iterable,  # @UnresolvedImport @Reimport pylint: disable=I0021,import-error
+        MutableSet,  # @UnresolvedImport @Reimport pylint: disable=I0021,import-error
+    )
 
 if str is bytes:
     intern = intern  # @ReservedAssignment pylint: disable=I0021,undefined-variable
@@ -109,3 +119,5 @@ assert type(xrange) is type, xrange
 assert total_ordering
 assert intern
 assert builtins
+assert Iterable
+assert MutableSet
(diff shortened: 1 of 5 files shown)


kayhayen:
Looks very good to me, let me know if you think of it as complete.

BoboTiG:
It is complete :)

BoboTiG:
Should I move the `MutableSet` to `nuitka.__past__` too?

kayhayen:
@BoboTiG You mean given that it is only used in the `oset` module which already performs a kind of function to hide its uses? I think I agree, no leave it there then. But `bdist_nuitka` should try and stay clear of version abstractions as far as possible.

Oh, and now I remember, there is a good reason to not rely on `try ... except ImportError`, but to use Python version checks, because that is less vulnerable to version changes. I think e.g. it has happened that Python2 backports, or user packages in a namespace, have severely confused Nuitka. Which it got changed for all of Nuitka at one time.

BoboTiG:
I updated the PR.
I hope I well understand what you were saying :) Let me know if something is not right.

nuitka/__past__.py
```diff
@@ -94,6 +94,10 @@ def total_ordering(cls):
 
         return cls
 
+try:
```
kayhayen: Actually I think this ought to use `if str is bytes:` instead of relying on the import error, much like this one:

```
if str is bytes:
    from urllib import (  # pylint: disable=I0021,import-error,no-name-in-module
        urlretrieve,  # @UnresolvedImport @UnusedImport
    )
else:
    from urllib.request import (  # pylint: disable=I0021,import-error,no-name-in-module
        urlretrieve,  #  @Reimport @UnresolvedImport
    )
```

Also not the comments, PyDev and PyLint will otherwise complain.

The reason is that sometimes people make backports of Python3 package names or forward ports of Python2 package names, and it's nicer if you express which version has which one. Since it's 3.3 or higher, for Nuitka that is Python3, and `str is bytes` is a very good test for that.

BoboTiG: Indeed, done :+1:

PR #265, opened 2019-02-19
Enable options for user plugins

JorjMcKie:

### PR Checklist

### DIFF
--- a/nuitka/plugins/standard/TkinterPlugin.py
+++ b/nuitka/plugins/standard/TkinterPlugin.py
@@ -1,4 +1,4 @@
-#     Copyright 2019, Jorj McKie, mailto:<lorj.x.mckie@outlook.de>
+#     Copyright 2019, Jorj McKie, mailto:<jorj.x.mckie@outlook.de>
 #
 #     Part of "Nuitka", an optimizing Python compiler that is compatible and
 #     integrates with CPython, but also works on its own.
--- a/Developer_Manual.rst
+++ b/Developer_Manual.rst
@@ -536,9 +536,20 @@ those with Doxygen only relatively late.
    doxygen ./doc/Doxyfile
    xdg-open html
 
-There is going to be enhancements to this API documentation in the
-next releases though, hopefully making it useable for at least the
-Nuitka plugin development.
+Improvements have already been implemented for plugins: The plugin base
+class defined in ``PluginBase.py`` (which is used as a template for all
+plugins) is almost fully documented in Doxygen now. The
+same is true for the recently added standard plugins ``NumpyPlugin.py`` and
+``TkinterPlugin.py``. These will be uploaded very soon.
+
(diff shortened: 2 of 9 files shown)


nuitka/plugins/standard/TkinterPlugin.py
```diff
@@ -1,4 +1,4 @@
-#     Copyright 2019, Jorj McKie, mailto:<lorj.x.mckie@outlook.de>
+#     Copyright 2019, Jorj McKie, mailto:<jorj.x.mckie@outlook.de>
```
kayhayen: Sorry about these, my bad

kayhayen:
Hello @JorjMcKie very good. Lots of new doc. Way to small a notice in the commit.

However, one small request, can you wait for Travis to check and allow me a chance to review before merging. Even if it looks simple, it's better to go slow. When Travis turns red while people branch off from develop, they get errors that they cannot understand. So we must strive to avoid it. Luckily it is not the case right now.

I am guilty of not yet writing up, how I think it should be done. But I think we should use feature branches in the Nuitka/Nuitka repository directory, and make the pull request from there, the another should review timely and ack or click merge. But as you are probably aware, I am not following that with my factory branch, but it definitely is the goal.

I think you only need to use Nuitka/Nuitka as an upstream. But like I said, I need to write something up for, definitely on my list. I think it will occasionally be very nice to directly cherry-pick if we share branch name space.

I noticed there is trailing white space in the `.rst` files, need to tune your editor there. I have AnyEdit in Eclipse, not sure what you are using. But I will also make the (slow, not cached yet `bin/autoformat-nuitka-source`) binary away of these file types and strip trailing white space. On my factory branch (sic) that tool now does black, isort, relative imports, Windows new lines, all the things, but I need to push it there.

Looking forward to my Buildbot picking up on it and updated the API docs.

Very happy right now. :-)

Yours,
Kay

kayhayen:
So I pushed my weekend's work for autoformat integrating isort (it and black are kind of fighting over formatting of imports), and it should be usable. I will try and follow up with something that trims the ".rst" files, so don't worry about "correcting" that, I am going to use it as a test case for the tool improvement.

JorjMcKie:
ok, got it, will wait for that in future.

I am using VS Code. Need to juggle a bit getting it delete the trailing spaces.

kayhayen:
@JorjMcKie I pushed the tool improvement. It runs with command line arguments, or by default over all python source of Nuitka, but it is very slow. And now covers `.rst` files too.

I want to make it so that for files changes, at `pre-commit` a hook will call it with the files of the commit, and updates their format automatically. I noticed you did use black yourself. But from random drive by contributors we cannot expect that to work well unless it's fully automatic. On the other hand, they may not enable commit hooks, but I hope they will. But if anything, this becoming automatic will be great.

The autoformat tool also may not work fully well on Windows, unless things are in PATH. Can you try it out?

Issue #258, opened 2019-02-15
Tkinter standalone test doesn't use TCL


kayhayen:
The test of Tkinter for standalone mode is far too weak. It has always passed despite #137 and that was probably the root cause of this not being detected earlier.

However, I lack all the knowledge to make a tkinter test that is headless. For PyQt5 and other things, we tend to have it, but often the headless nature makes things difficult. In my Buildbot tests, there should be no Windows popping up (I use my main desktop for it, and my own account).

So we need to enhance the test to cover this. I am hoping @JorjMcKie can help us out here.

That way his glorious plugin which I am going to release later today will be tested, and cannot break as easily in the future as it otherwise could.

kayhayen:
I have assigned this to you @JorjMcKie which I hope is OK for you, let me know if I ought to wait for you to take it. :) It would be very nice to have a PR that adds something that will make it fail when executed without your plugin and pass with it. The test runner will use your plugin for that test automatically btw.

I am speaking of this one:

https://github.com/Nuitka/Nuitka/blob/bc22aa833c22e2f5f55304ec47b57c71c6adf16d/tests/standalone/TkInterUsing.py#L27

As you can see, very impressive test coverage. I hope I can rely on you to expand this is a bit. If it's hard (headless for a GUI framework) then OK, we can also try and make me shout out about this on Twitter.

JorjMcKie:
@kayhayen - sure, I will take it on.

JorjMcKie:
submitted the PR

JorjMcKie:
Given the design of the Tkinter plugin, I could also have checked the existence of the respective `os.environ` value(s), i.e. without even importing tkinter - just doing a `assert os.environ.get("TCL_LIBRARY", None)`.
But then ... who knows what may happen to the plugin one day?

kayhayen:
Thanks a lot. I also forgot that when I enabled the plugin.

With your changes, Linux was now properly failing, I have a fix for that on my factory, adding system TCL/Tk installation to the whilelisted files that are allowed to be used. I am about to push as soon as Travis liked it for real. Closing this and thanks a lot.

To me this is mostly about having a test that will break when I make re-factorings to the plugins, and then will only learn much later about it breaking.

Also thanks for the speedy reaction. Made my day. :)

JorjMcKie:
any time again - I assure you.
As I mentioned: I am a retired man having plenty of time

kayhayen:
[...]

PR #266, opened 2019-02-20
Create Doxygen relevant documentation

JorjMcKie:

### What does this PR do?
No actual code changes, just changes of inline documentation.

We need to improve or API documentation, starting with the most relevant parts for user contribution. In a first round trip, Nuitka plugin modules are receiving inline documentation that can be used by Doxygen to create relevant and meaningful API documentation.
Also refer to mentions in #258 and comments in the [Developer Manual[(https://github.com/Nuitka/Nuitka/blob/develop/Developer_Manual.rst#internal-plugin-api).

### PR Checklist


kayhayen:
@JorjMcKie The trailing whitespace is hated by PyLint, it seems 2.6 is currently broken on Travis for at least a day now. You can ignore that part.

Absolutely love all the content of course. So fixing the trailing white space will be good. I will look into disable 2.6 on Travis, or if they are going to repair that. Wouldn't be a too big surprise if it stopped working, now that the RHEL 7 are on Python 2.7 too, aren't they. On the other hand, I have people who are using it with RHEL 5.

One thing I was thinking, if docstrings are really not workable for Doxygen, but I found this:

https://stackoverflow.com/questions/58622/how-to-document-python-code-with-doxygen

There is also mention of something called `doxypy`, but I think it's better to go with some sort of standard, and not a mix form that nobody can know.

Thanks a lot for taking this on!

Like I said, merge once Travis is green for all but 2.6, fine by me. And I truly mean, this is a huge relief to me, because this removes an important missing piece off my chest, and with your plugins now that well documented, we can even make them show cases, how things ought to be done.

kayhayen:
Ah, and if you get `bin/autoformat-nuitka-source` to work on Windows, it ought to automatically fix your source, although I fear on Windows, it may even be more slow to walk all the files one by one @JorjMcKie 

Otherwise I can give it a shot later this weekend, although I will probably be too busy responding to the many GSoC people. I think we too should have a hangout call if you don't mind. So we can agree on which students/ideas you go with.

JorjMcKie:
Thanks for all that feedback, indeed!
I am still working on my editor VS Code to wellbehave. I now managed to inautomatically invoke black on save which removes trailing spaces as well.
Still need to test this autoformat stuff.
As per doxygen, thanks for the pointer, will look at that.

I will keep an eye on those travis checks and make sure they pass - at least the indispensible ones.

As per the call: would be great. Let me send you the logistics for that per your personal e-mail (need to use Skype b/o Venezuela, but shitty internet connection, therefore pesky echo effects ... and, and, and).

JorjMcKie:
Topic Doxygen:

* doxypy seems to have died
* there is a successor doxypypy which claims to be even better: a pure Python solution found on PIP.

I tried doxypypy. Looks good and lets us stay with the pythonic way of documenting stuff (""" ... """), given that certain restrictions as per its formatting it are obeyed (nothing prohibitive, not even serious).
It correctly recognizes classes and methods with their arguments, kw arguments, returns, and notes.

There are some other conditions that must be met:

* activate a filter for `*.py` files in `Doxyfile`
* have a shell command / Windows bat available which gets invoked by doxygen when it generates the documentation for a `.py` file. What happens is, that doxygen passes the py-file to that batch script and actually scans its piped output.

So that system, that creates the API doc, needs to have doxypypy set up correctly . Is this your machine?
Or should I do that on my machine and send around the generated html folder (compressed with 7zip its just 800 KB)?

Or ... I continue with the way I have started the 3 plugin scripts with.

We need to decide

kayhayen:
BTW: @JorjMcKie I broke develop for Windows, force pushed something this morning, therefore your conflict, the `readlines` is necessary. Sorry about that. I was too eager to get `autoformat` in your hands, got to stick to my own advice and be sufficiently patient. :-)

As for the other thing, there is a script that does the building of the API doc and uploads it. It runs on my machine when develop changes on the `nuitka.net` git repo, which I currently only do when I make pre-releases. I will change that soon. The script lives here:

https://github.com/Nuitka/Nuitka/blob/b3d6b50e8486f513a9ef18a6469beb30a6f3b6f5/misc/make-apidoc-upload.py#L39

Building and uploading should always be automatic, just to be able to forget about those steps. My moving to fully Github is relatively new. I have to think how making pre-releases fits into this model where develop moves more often. Maybe will use tags for them.

And for the formatting, doc strings are more natural to people. I would have no issue installing that software, add it to `requirements-devel.txt` with a versioned requirement, then it's easy to pull in for everybody. I just don't want to find us using something that is dead tomorrow.

On the other hand, non-doc strings will confuse Python people deeply. If you trust that other project to live, I am OK with going there. Your pick. I am going with your decision either way. But like I said, adding a software to the build wouldn't be a problem.

kayhayen:
Are you OK with me merging this as it is now, just to resolve the merge conflicts? Or do you want to move to the doc strings before merging. Since it's my force pushes, I sort of feel obliged to cleanup, which btw I could actually do better (without having to merge), if the branch lived in the Nuitka repository.

JorjMcKie:
of course - just go ahead. You can also delete the PR, whatever is most convenient for you.

I am preparing another version which is based on that new pointer you provided (doxypypy).
The approach required to re-format the three plugin modules **again**.
But they now look like familiar Python again, e.g.

```python
    def getPluginOptions(self):
        """ Return all options for the plugin.

        This method will always return a list of strings.
        To specify options, append a '=' to the plugin_name.
        Everything what follows up to and excluding the next space will be taken to form a 'raw' value.
        The result of raw.split(',') will be returned by this method.
        Apart from this procedure you are free to specify anything, e.g. filenames, etc.
        To include spaces use enclosing double apostrophies. Here are some examples for a plugin named "name":

        name=a,b,c ----> ['a', 'b', 'c']
        name=a=0,b=1,c=2 ----> ['a=0', 'b=1', 'c=2']
        name ----> []
        name= ----> ['']
        "name={'a':0, 'b':1, 'c':2}"  ----> ["{'a':0", " 'b':1", " 'c':2}"]
                                            (note the spaces preceding 'b' and 'c'!)

        Returns:
            list of strings
        """
        return Options.getPluginOptions(self.plugin_name)
```

With just some very minor restrictions, doxypypy takes this format and temporarily creates the format which I had been using before if you remember that.

My next PR will hence also contain a modified `Doxyfile` (1 line changed only), and 2 batch scripts (1 for Windows and 1 hopefully Linux) which invoke this doxypypy tool.
With some luck, a few more documented existing modules might also be included.

PR #271, opened 2019-02-21
Update README.rst

anderspersson:
Fixed a typo in **CC=c:\\mingw64\bin\gcc.exe**
where a single \ was giving a bingcc exe..
/A


anderspersson:
Fixed a typo in **CC=c:\mingw64\bin\gcc.exe**
where a single \ was giving a bingcc exe..
/A

PR #276, opened 2019-02-23
Call to plugins after end of standalone compilations

JorjMcKie:

### What does this PR do?
1. After successful creation of the ``dist`` folder in standalone compilations, a new plugin method ``onStandaloneDistributionFinished`` is called with the distribution folder as parameter. Plugins are free to execute any downstream action.

2. User plugins are now loaded **before** any standard plugins. This allows user plugins to in turn enable standard plugins by themselves.

With these changes, a more complete coverage of the software creation process is possible, while also reducing the need for extra logic to wrap Nuitka. Profiles for various purposes and audiences can be created and implemented as user plugins, like e.g. setting company standards for distributed software, or special debugging, testing or benchmarking plugins can be developed.

### PR Checklist

### DIFF
--- a/Using-Plugin-Options.rst
+++ b/Using-Plugin-Options.rst
@@ -66,7 +66,7 @@ check single option items via convenience method ``self.getPluginOptionBool("val
 
 Remark
 --------
-Obviously, you can recover the original "raw" string by ``raw = "".join(self.getPluginOptions())``.
+Obviously, you can recover the original "raw" string by ``raw = ",".join(self.getPluginOptions())``.
 If your plugin knows which format to expect, it could do something like this: ``my_option_dict = json.loads(raw)``, or
 this: ``my_option_list = raw.split(";")``.
 
--- a/nuitka/MainControl.py
+++ b/nuitka/MainControl.py
@@ -799,6 +799,8 @@ def main():
 
                 shutil.copy2(source_filename, target_filename)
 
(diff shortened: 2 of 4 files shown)


kayhayen:
Very nice @JorjMcKie . I had to disable white space in the file diff, not sure if it turned to or away from Windows new lines, I will check later and possibly amend the commit if it introduced white space. Just be aware that a `git pull --rebase` is always a good idea. :-)

This is definitely a very useful interface. I noticed that something about priorities was apparently needed, being called before standard plugins. Conflicts between plugins or desire to do something after or before all the others, definitely a thing. But I have no idea how that can work. But it's probably going to be OK. Too involved things should be standard plugins most of the time anyway.

JorjMcKie:
Thank you very much @kayhayen ... also for your nice feedback!

Maybe I haven't been clear enough about this "priority" thing: it only means that in ``Plugins.py`` method ``initPlugins()``, invoking ``importUserPlugins()`` is the first thing to happen.

The example I published in this other thread inserts ``tk-plugin`` into the resp. ``Options.options`` list. So this standard plugin is getting handled fully as if entered via command line.

Of course we still need to evaluate if we want to allow that happening at all and / or if we want to expose that to the user plugin developer. From a certain point of view this could be regarded an abuse ...

JorjMcKie:
Maybe I should consider a proper public interface for this type of options manipulation ...

PR #285, opened 2019-03-02
Test4.py is incomplete

girijamanoj:
I am not sure, why it is left incomplete (I hope it's not intentional), I tried making a minor change which will do the intended work for us.

Thank you for contributing to Nuitka!

### PR Checklist

### DIFF
--- a/tests/type_inference/Test4.py
+++ b/tests/type_inference/Test4.py
@@ -25,9 +25,9 @@ def someUnpackingFunction():
     return i1+i2+i3
 
 
-print someFunction()
+print someUnpackingFunction()
 
 def someShortUnpackingFunction():
-    a, b = 1,
-
-someShortUnpackingFunction()
+    a, b = 1,2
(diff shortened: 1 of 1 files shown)


kayhayen:
Good catch. These tests are not put to use yet. I haven't really developed a test strategy there that allows to check Nuitka's ultimate idea about a value.

If you want to go for it, lets add a plugin API where the final node tree is handed to you, and you would call `getTypeShape` on variable references, and match it with comments (not in the node tree, need to find that from source).

Do you think that would be a workable PR for you @girijamanoj ?

girijamanoj:
> Good catch. These tests are not put to use yet. I haven't really developed a test strategy there that allows to check Nuitka's ultimate idea about a value.
> 
@kayhayen I guess we have to start developing a strategy as fast as we can.

> Do you think that would be a workable PR for you @girijamanoj ?
>
I am ready to work on this. :)

PR #290, opened 2019-03-05
Tests: Cover more Trick Assignments tests

bksahu:
List of new tests added
--------------------------------
- [x] ipow
- [x] iand
- [x] ifloordiv
- [x] ilshift
- [x] irshift
- [x] ior
- [x] itruediv
- [x] ixor

Also updated inplace matmul in py35.

### DIFF
--- a/tests/basics/TrickAssignments32.py
+++ b/tests/basics/TrickAssignments32.py
@@ -682,7 +682,7 @@ def someFunctionThatReturnsDeletedValueViaInplaceAnd():
             nonlocal a
             del a
 
-            return 7
+            return False
     c = C()
 
 
@@ -728,7 +728,7 @@ def someFunctionThatReturnsDeletedValueViaInplaceLshift():
             nonlocal a
             del a
 
-            return 7
+            return False
     c = C()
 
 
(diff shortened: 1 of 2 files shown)


kayhayen:
There is already a matmul test in `TrickAssignments35.py`.

The Travis build was failing for the one that you added to `TrickAssignments32.py` because that means it's used for Python 3.2 (OK, we don't do that any more) or higher, and that is then a syntax error for Python 3.3 and 3.4, which didn't yet have that `@` (matmul) operator.

But otherwise, good job completing the missing ones. I am also glad to see them pass. :-)

bksahu:
Yeah you are right. `@` support started from 3.5. I'm going to remove `matmul`

tests/basics/TrickAssignments32.py
```diff
@@ -702,9 +882,6 @@ def __len__(self):
     print("Ouch.!")
 
 
-# TODO: The matrix multiply needs a dedicated test too.
-
-
 # TODO: There is a whole lot more operations to cover.
```
kayhayen: So that's probably a solve TODO now

bksahu: Yup

Issue #299, opened 2019-03-18
Standalone: Doesn't include the Qt platform plugin "xcb"


dark-penguin:
I'm building a "hello world" with PyQt5 with options:
```
nuitka3 test.py --show-progress --recurse-stdlib --standalone
```
The resulting binary does not work:
```
This application failed to start because it could not find or load the Qt
platform plugin "xcb".
Reinstalling the application may fix this problem.
Aborted (core dumped)
```
This happens even on the system where I built it.

Adding `--plugin-enable=qt-plugins` does not help.

Also, my application includes a "test mode": if I run it with "--test", it
prints a line "Test passed" and exits before importing anything PyQt-related.
And launching it this way works!

Actually, I just had the same issue with PyInstaller, but then I updated it to
the latest version, and now it's fixed. (I thought this information may be valuable.)

I tried this on Debian Jessie, Debian Stretch and Ubuntu Cosmic, with stable
Nuitka (from pip) and with development Nuitka (from sources), with whatever Python versions they have in the repos (IIRC that's 3.4, 3.5 and 3.6). So basically, it happens "always".

Here is my test.py :
```
import sys
if '--test' in sys.argv:
    print("Test passed")
    sys.exit(0)

from PyQt5.QtWidgets import QApplication, QMainWindow, QTextEdit
app = QApplication(sys.argv)
mw = QMainWindow()
textedit = QTextEdit(mw)
textedit.show()
mw.setCentralWidget(textedit)
mw.show()
sys.exit(app.exec_())
```

kayhayen:
Hello,

it works for me though. I am definitely positive, that PyQt works on Stretch and current Buster both.

It seems "xcb" platform plugin might be needed on Linux under some circumstances. 

```
                # Seems platforms is required on Windows.
                if os.name == "nt":
                    plugin_options.add("platforms")
```

I think we also have `--enable-plugin-qt-plugins=all` as a solution for now. 

Make that `or True` for the if statement, and report if it's working there. I think, what we ought to do is have a set of minimal chroots (`debootstrap`) in the standalone tests and run the resulting binary there as well. It seems that if python-qt5 is installed on the target machine, it works. But that is besides the point of standalone distribution.

On Windows, it always failed due to not having PyQt installed system wide and my Linuxes all have PyQt installed I guess, and probably most systems do. It may also be the case that the platform plugins actually are supposed to come from system installation, but I am not sure of that at all, I doubt it, but it could be.

[...]

PR #304, opened 2019-03-19
Add "platforms" to the list of "sensible" qt-plugins (temporary fix for #299)

dark-penguin:
For qt-plugins, "platforms" plugin is needed not only on Windows, but also on Linux sometimes. Hence, it should be a part of the "sensible" plugin list - at least until we can figure out when is it not necessary (see #299).


kayhayen:
I do not think #299 is all that relevant for this change, I would have it either way, simply because it makes sense. The relevancy of the issue is more in explaining why it ever worked for me, despite the `strace` approach with which I tried eliminate opening unwanted system files.

Issue #283, opened 2019-03-01
Generating specialize C helper code automatically


kayhayen:
This issue is about generating specialized code helpers automatically. A code helper is considered specialized, if it doesn't work on the most general form, e.g. two `PyObject *`, but instead e.g. on two `PyUnicode_Object` instead, where the task can be identified more immediately as `PyUnicode_Concat`, just concatenating the two strings.

In Nuitka as a result of compile time analysis during optimization we have increasingly more shape knowledge. Often with relatively concrete types. A shape is a set of characteristics that the uses of a expose, e.g. could mean it's "iterable", "indexable", as example shapes. Many times however, shapes are known to be one or two types specifically, `int`, `bool`, `unicode`, etc. but often also one of the other, e.g. `unicode` *or* `str`.

Currently in Nuitka there are helper functions used for code that does all sort of operations. There are very general ones, and then the more specialized ones. The general ones use `OBJECT` as a type indicator in the function names (where there is specialization). As of writing this, this is only the case for '+'  and `+=` (`_INPLACE`) are optimized for a subset of types and shapes. And there are comparison operations, `<`, `<=`, `==`, `!=`, `>=, `>`, but they are only specialized for `INT`, i.e. Python2 integers.

These helpers are hand optimized versions of more general code that represents what e.g. `PyNumber_Add` does. It looks at type slots and implements special behavior like the concatenation that `+` also does in Python. Many of their forms are not optimized, or not yet fully optimized. The current form is more of a demonstration of feasibility than a practical approach. This will only be sufficient to get few operations optimized, but it will never scale and doing this by hand is error prone and could lead to bugs.

Instead, what we need is a code generator with sufficient type knowledge about the two arguments, so that it can make decisions automatically. For instance, having a `nb_add` is a given in for `long`, `float`, but not so for a `list`. So instead of making checks, and executing branched code, the code generator would not generate that check, or do so in a way that the C compiler will know it's not happening. 

Other examples are checks like `type1 == type2`. Many times, this comes down to branch decisions. And the different shapes allow some of these, some do not.

kayhayen:
The slots in the Python type system, e.g. `nb_add` are related to, but still somewhat different to slots in the Python language itself:

http://nuitka.net/doc/developer-manual.html#python-slots-in-optimization

kayhayen:
Consider e.g. this very general function used for `+`:

[...]

PR #306, opened 2019-03-20
Generating C helper code using Jinja2 and python

riya-17:
hey @kayhayen, I implemented this code and it is working fine yet there are some problems with jinja2 and I am trying to figure that out like there are no global variables in jinja2 which is creating a problem to pass values from jinja2 to python.

one of the solution that i got is creation of dictionary as it always points to its address so no ambiguity is caused but it is also creating a little problem. If you know any relevant solution please do guide me. 

### What does this PR do?
This PR add two given numbers with the help of jinja2. This code is a combination of python and jinja2.

### Why was it initiated? Any relevant Issues?
#283 

### PR Checklist
- [ ] Adding global variable(this can solve almost every problem)

PR #307, opened 2019-03-20
Generating C helper code using Jinja2

riya-17:
Hey @kayhayen i tried to implement this code in complete Jinja2. There is a code block named macros which is same as function in any other programming language. There are some setbacks in jinja2 like it doesn't have global variables which is creating problem in passing values. I couldn't find any other coding block that can behave equally as macro but if I could find one I will replace that.

### What does this PR do?
This PR add two numbers with the help of Jinja2

#283 

### PR Checklist
- [ ] Global variable (can solve most of the problem)
- [ ] Finding better solution than macros or improve macros.
- [ ] improving macros : number as Exact_integer confirmation.

### DIFF
--- a/nuitka/tools/gen_helper_code/BINARY_ADD.py
+++ b/nuitka/tools/gen_helper_code/BINARY_ADD.py
@@ -1,0 +1,33 @@
+
+
+import jinja2
+
+"""
+def BINARY_ADDITION(a, b):
+	if(isinstance(a,str)):
+		print("Not an integer") 
+	else:
+		if (type(int(a)))==int:
+			if(isinstance(a,float)):
+				print("Not an Integer")
+			else:
+				c = a+b
+				return c
+"""
+macro = ("{% macro add(a,b) %}"
+			"{# if PYTHON_VERSION < 300 #}"
+         		"{% if a is number %}"
+         			"{% if b|int !=0 %}"
+         				"{{ a+b }}"
+         			"{% endif %}"
+         		"{% endif %}"
+         	"{# endif #}"
+        "{% endmacro %}")
+use_macro = "Add a + b = {{ add(a=2,b=5)}}"
+loader = jinja2.DictLoader({'template': use_macro})
+env = jinja2.Environment(loader=loader)
+macro_template = env.from_string(macro)
+env.globals['add'] = macro_template.module.add
+template = env.get_template('template')
+rendered = template.render()
+print(rendered)


nuitka/tools/gen_helper_code/BINARY_ADD.py
```diff
@@ -0,0 +1,33 @@
...
+		print("Not an integer") 
+	else:
+		if (type(int(a)))==int:
+			if(isinstance(a,float)):
+				print("Not an Integer")
+			else:
+				c = a+b
+				return c
+"""
+macro = ("{% macro add(a,b) %}"
```
kayhayen: I would try to avoid macros, pure Python code seems more powerful and readable.

nuitka/tools/gen_helper_code/BINARY_ADD.py
```diff
@@ -0,0 +1,33 @@
...
+         				"{{ a+b }}"
+         			"{% endif %}"
+         		"{% endif %}"
+         	"{# endif #}"
+        "{% endmacro %}")
+use_macro = "Add a + b = {{ add(a=2,b=5)}}"
+loader = jinja2.DictLoader({'template': use_macro})
+env = jinja2.Environment(loader=loader)
+macro_template = env.from_string(macro)
+env.globals['add'] = macro_template.module.add
```
kayhayen: So I see you figured out how to use an environment, but I still don't think we need one really. Recall my comments about how methods of type description objects can basically do all we want already. There is barely a need for global stuff. And if, indeed, put python_version as a global variable like that.

nuitka/tools/gen_helper_code/BINARY_ADD.py
riya-17: This was about implementing it in pure jinja2 form that's why I included macro. If you think Python code is more powerful with jinja2 than what is the problem with this code https://github.com/Nuitka/Nuitka/pull/306

nuitka/tools/gen_helper_code/BINARY_ADD.py
riya-17: Environment was necessary to call macro that's why that was needed to be implemented.

riya-17:
@kayhayen I tried to implement pure jinja2 code because jinja2 + python consists mostly of python and just passing of data takes place with the help of jinja2. So, I thought that this will not help in increasing the speed of the functionality but if you are saying python + jinja2 is better than I will go with it :)

kayhayen:
@riya-17 The run time speed is no concern for the Jinja templating at all, because it is only run seldom and we are then going to use the generated C code after that. 

I was contemplating to use it in other places too, but I doubt that speed will be a problem at all even then. Generating the C code will always be way less CPU time than compiling it with say gcc, and optimizing the Python node tree in Nuitka, even if it became dramatically less efficient. Readability of the code will also be important. And my impression is that the more that happens in the template, the less clear things become.
 
I am OK, with Jinja2 being a very thin layer only, in-line to C code, and most of the logic being in Python and it getting passed two objects that describe the type pair to generate code for, and maybe the operation at hand.

kayhayen:
So I am closing this one, as I think we agree it's not a good option for the task.

Issue #308, opened 2019-03-20
Module from package scipy is not being included as expected and error when including explicitly


robguinness:
# Description of issue

When building a standalone application that utilizes the package scipy (included via the flag `--include-package=scipy`), I get a runtime `ModuleNotFoundError` for `scipy.special._ufuncs_cxx` when the application is run. In an attempt to fix this, I added the flag `--include-module="scipy.special._ufuncs_cxx"` on a subsequent run, but this results in a build time error of:

```
Error, '_ufuncs_cxx.cpython-36m-x86_64-linux-gnu.so' is not a proper python module name.
```
After looking at the source for `nuitka/tree/Building.py`, this error is thrown when there is a `.` in the module name. But `_ufuncs_cxx.cpython-36m-x86_64-linux-gnu.so` is the name of the actual file where this module resides, so perhaps this is a bug.

# Detailed version/environment info

```
$ python -m nuitka --version
0.6.2
Python: 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
Executable: /home/rob/anaconda3/bin/python
OS: Linux
Arch: x86_64
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.2 LTS
Release:	18.04
Codename:	bionic
```

*  How did you install Nuitka and Python (pip, anaconda, deb, rpm, from source,
   what is a virtualenv ...), this is very important usually.

Nuitka installed via `conda install nuitka`
(Also tried installing system wide using `pip install nuitka`, but the above error is the same in either case)

# Short Example to Reproduce Issue

1. Install `scipy` using pip, anaconda, or build from source.

2. Create a short test script `scipy_test.py`

```
from scipy import special
kth_zero = special.jn_zeros(1, 1)
print(kth_zero)
```
(This is to just make sure that the above module is utilized in the application)

3. Build it into a standalone application using nuitka:

```
$ python -m nuitka --standalone --show-progress --show-modules \
        --python-flag=-S \
        --include-package=scipy \
        scipy_test.py
```

_Note this will pick up a lot of modules that aren't really needed for this simple example, so I recommend excluding them by adding --nofollow-import-to flags, e.g.:_

```
$ python -m nuitka --standalone --show-progress --show-modules \
        --python-flag=-S \
        --include-package=scipy \
	--nofollow-import-to=Cython \
	--nofollow-import-to=tornado \
[...]

PR #309, opened 2019-03-21
Add basic "scipy" package support in ImplicitImports plugin

JorjMcKie:

### PR Checklist

### DIFF
--- a/UserPlugin-Creation.rst
+++ b/UserPlugin-Creation.rst
@@ -78,3 +78,80 @@ Sample invocation line and output::
 
     $
 
+Nuitka Options Overview
+------------------------
+About 60 methods are available to access option information for the
+current Nuitka execution. Import the ``Options`` module
+(``"from nuitka import Options"``) and use one of the following.
+
+Please note that ``str`` results in general may return ``None``.
+Except very few, the methods have no argument.
+
+===================================== ======================================================================================
+**Method**                            **Description**
+===================================== ======================================================================================
+assumeYesForDownloads                 *bool* = ``--assume-yes-for-downloads``
+getExperimentalIndications            *tuple*, items of ``--experimental=``
+getFileReferenceMode                  *str*, one of ``"runtime"``, ``"original"`` or ``--file-reference-mode``
+getIconPath                           *str*, value of ``--windows-icon``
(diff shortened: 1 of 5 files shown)


JorjMcKie:
Change in ``Recursion.py``, function ``checkPluginSinglePath``: now use the ``module_kind`` value in call to ``recurseTo`` instead of a fixed value of ``py``. This addresses issue #308.
Adding scipy support also adresses #308, by adding some new implicitely imported modules.

Please note that a full Scipy support will require a new plugin - rather similar to numpy-plugin.

JorjMcKie:
@kayhayen - could you review my change to ``Recursion.py`` please? There always is a fat chance that I am missing something crucial ...

UserPlugin-Creation.rst
```diff
@@ -78,3 +78,80 @@ Sample invocation line and output::
 
     $
 
+Nuitka Options Overview
```
kayhayen: Wow! This is an amazing by product :)

nuitka/plugins/standard/ImplicitImports.py
```diff
@@ -259,6 +259,11 @@ def getImplicitImports(self, module):
             yield "apt_pkg", True
         elif full_name == "numpy.core":
             yield "numpy.core._dtype_ctypes", False
+        elif full_name == "scipy.special":
+            yield "scipy.special._ufuncs_cxx", False
```
kayhayen: The `False` is really good for things that are not always that way. We don't explain that anywhere, do you?

nuitka/importing/Recursion.py
```diff
@@ -275,7 +275,7 @@ def checkPluginSinglePath(plugin_filename, module_package):
                 module_filename=plugin_filename,
                 module_relpath=module_relpath,
                 module_package=module_package,
-                module_kind="py",
+                module_kind=module_kind,  # "py",
```
kayhayen: As a comment, `# py` sucks of course, but I take, that it forgot to forward the `module_kind`, good catch.

kayhayen:
@JorjMcKie Fantastic job, you absolutely exceeded all realistic expectations I had about the PR. Very good that you added so much documentation too.

JorjMcKie:
Thanks a lot for your feedback!

That lonesome # "py" comment was an eyecatcher for you only. Should be deleted of course.

Issue #294, opened 2019-03-10
Specialized C containers for known constraints


kayhayen:
Traditionally containers are designed to be good at typical usage patterns. The Python dictionary and the Python list are well optimized for what code typically needs. However, in order to shave of the maximum performance, this involves trade offs, that are not ideal for all the cases and we might be better off with other approaches in some cases.

Consider this code:

```
def f():
   result = []
    
   for value in unknown_iterator_size():
         result.append(something_unknown(value))
 
   return sum(result)

```

This is a very common usage pattern in Python code. The list is purely local. It doesn't escape. Unfortunately we cannot know the size of the list we are talking here. We could achieve in this particular instance, and earlier `sum`, but lets assume that is not done, and it's very high levels optimization needed to get us there, and `sum` is not the issue here.

The point is rather, that `PyList_New()` Python C/API is going to have to be called, to create a list right now. That does `malloc` an object, and inside of it, a storage space, both of which will later have to be released with `free`. For every `append` we have to consider that the list storage space might have to be expanded, which is not every time, but leads to a couple more `malloc` during `PyList_Append` Python C/API.

Imagine, we control the `sum` well enough to know it is going to make only start to finish value iteration, and nothing more besides this. Imagine we create a container, that is only good for `append`, in fact need not provide any other operation at all, or at least doesn't need those to be fast, as e.g. they are done only once, e.g. `len()`, the equivalent of `PyList_Size` would be allowed to make an actual calculation, and not be a stored value in the structure, it would e.g. only contain a pointer to the end.

Imagine further that we keep storage spaces pre-allocated for non-overlapping lists, that are deemed sufficiently large to hold the list information and relatively many elements, and the container would have bounds checking only, but normally no growing needed. 

CPython, not knowing the usage of the list, has to cover the case of many small lists with its implementation, cannot allocate too much storage in the beginning, and wants to limit the over allocation of lists, but we can start out relatively big for this local list, no worry.

Avoiding the `malloc` entirely, will do us good already.

Obviously, we also want the container to also be more flexible, and store other values than `PyObject *`, suppose (not the case right now), we knew that it was a list of float values. We would want to store C `double` instead of the `PyObject *` for as long as possible, and let `sum` know, that is it, what it's working on, and be more direct.

Avoiding these `PyObject *` boxes for the values, would be a even larger memory saving measure, and one that is not currently possible without specialized containers.

[...]

PR #311, opened 2019-03-24
 Specialized C containers for unknown sizes.

sannanansari:
This PR is for creating an `python list` using C.
 #294 Specialized C containers for known constraints.


kayhayen:
This is missing the API functions that are similar to the list API of CPython as described in the original issue.

The use of an embedded CPython to compare PyList performance and API usage with your replacement. Mind you, it should be near similar API.

sannanansari:
Sorry I am little bit confused about API. According to me API is like  [C example](https://github.com/Nuitka/Nuitka/blob/9797f77c991ff8507ce84cff35f0d35011189940/nuitka/build/static_src/MainProgram.c).
Please you can explain a bit about API like this. And how can I run it. Sorry for my inexperience.

kayhayen:
@sannanansari That is not meant at all, rather something like this from the CPython API:

https://docs.python.org/2/c-api/list.html

As described in the issue #294 what we are looking at, is a similar container that doesn't do C `malloc` but uses a static global buffer. So when the Nuitka static analysis has shown that to be possible, we could have a faster form here, but one that fits into templates. Having a similar API that the C compiler will often inline away allows for easy generation of C code for cases where it can be used and for cases where it can not be used.

sannanansari:
@kayhayen Please you can review it and tell my mistakes. I have write this code with the help of documentation.

sannanansari:
@kayhayen But problem is that it is not running with nuitka-env. It is running with python.

kayhayen:
@sannanansari not running in Nuitka is OK for this exercise, this is foremost about comparing containers that use e.g. PyList, and PyDict, API etc. and ones that do not do that.

You would e.g. create a container that has a similar API, but is as described in the GSoC issue, not doing a malloc, outside of a static pre-allocated buffer. Maybe if you wrap it in a Python function, the Python API and the @sannanansari API, you can compare them with `timeit`, or the ticks with `/misc/run-with-valgrind` (spelling), what you prefer.

That would be your test bed. And then try to squeeze out things with e.g. "created empty/append some/full Iteration only" lists. That seems good stuff.

kayhayen:
That last commit is quite a shocker to me. You copied the bulk of another PR, including unfixed issues, and definitely not good code. And I am unsure how it relates to your task. Previously you seemed to be generating a C extension that could be ran, and that will save your from building hassle, and now you "stole" a failed build system for a main program binary that calls gcc manually, and won't work. Not a goo idea.

sannanansari:
I am using Github Desktop first time. I want to go from `nuitka->develop` to `nuitka->master`. I accidentally press `ok`. It get commited. Sorry for that.

kayhayen:
No problem, what you have there looks like a fair and probably working small extension module to test the PyList performance. Now make another one, which uses an alternative list implementation of yours, that has similar API.

kayhayen:
This is unlikely to be completed at this time.

PR #314, opened 2019-03-28
run_all.py new special-comment mechanism & Urllib3Using.py

tommyli3318:
### What does this PR do?
This PR changes the mechanism in which run_all checks for whether import requirements are satisfied. Now each test file should include a special comment in one of the following formats:

`# nuitka-skip-unless-expression: expression to be evaluated`
(a test would be skipped if this expression evaluated to false)
OR
`# nuitka-skip-unless-imports: module1,module2,...`
(a test would be skipped if any modules are missing)

The new helper function `checkRequirements` inside run_all.py will search for these comments in every test module and do the respective checks as specified.
This new mechanism makes it more clear what imports are required for each test file, as they will be stated in a special comment inside each test file. It also improves code readability and eliminates repetition in run_all.py 

This PR also changed every existing test file to include the special comment stating their required imports

In addition, the new Urllib3Using.py module is added, which run_all.py uses to test Nuitka's compatibility with urllib3.

This is part of my Google Summer of Code proposal.

### PR Checklist

### DIFF
--- a/tests/standalone/Urllib3Using.py
+++ b/tests/standalone/Urllib3Using.py
@@ -1,0 +1,22 @@
+import urllib3
+
+# testing request
+http = urllib3.PoolManager()
+r = http.request('GET', 'http://httpbin.org/robots.txt')
+r = http.request('POST','http://httpbin.org/post',fields={'hello': 'world'})
+
+# testing response
+r.status
+r.data
+r.headers
+
+# testing JSON content
+import json
+r = http.request('GET', 'http://httpbin.org/ip')
+json.loads(r.data.decode('utf-8'))
+
+# testing binary content
+r = http.request('GET', 'http://httpbin.org/bytes/8')
+r.data
+
+print('Reached end of file')
--- a/tests/standalone/Urllib3Using.py
+++ b/tests/standalone/Urllib3Using.py
@@ -1,5 +1,7 @@
 import urllib3
 
+# nuitka-skip-test-unless: urllib3
+
 # testing request
 http = urllib3.PoolManager()
 r = http.request('GET', 'http://httpbin.org/robots.txt')
--- a/tests/standalone/Urllib3Using.py
+++ b/tests/standalone/Urllib3Using.py
@@ -1,24 +1,59 @@
 import urllib3
+import os
+from http.server import HTTPServer, BaseHTTPRequestHandler
+from threading import Thread
 
 # nuitka-skip-unless-imports: urllib3
 
+def runHTTPServer():
+    class myServer(BaseHTTPRequestHandler):
+        def do_GET(self):
+            if self.path == '/':
+                self.path = '/index.html'
+            try:
+                file_to_open = open(self.path[1:]).read()
+                self.send_response(200)
+            except:
+                file_to_open = 'File not found'
+                self.send_response(404)
+            self.end_headers()
+            self.wfile.write(bytes(file_to_open,'utf-8'))
+    
+    server_address = ("127.0.0.1",8020)
+    global server
+    server = HTTPServer(server_address,myServer)
+    server.serve_forever()
+
+Thread(target=runHTTPServer).start()
+print('Server started')
+
 # testing request
 http = urllib3.PoolManager()
-r = http.request('GET', 'http://httpbin.org/robots.txt')
-r = http.request('POST','http://httpbin.org/post',fields={'hello': 'world'})
-
-# testing response
-r.status
-r.data
-r.headers
+r = http.request('GET', 'http://localhost:8020/')
+# print response
+print(r.status,r.data,r.headers)
 
 # testing JSON content
 import json
-r = http.request('GET', 'http://httpbin.org/ip')
-json.loads(r.data.decode('utf-8'))
+# make a temporary test file
+with open('testjson.json','w') as f:
+    f.write('{"origin": "128.195.97.166, 128.195.97.166"}')
 
-# testing binary content
-r = http.request('GET', 'http://httpbin.org/bytes/8')
-r.data
+r = http.request('GET', 'http://localhost:8020/testjson.json')
+print(json.loads(r.data.decode('utf-8')))
+os.remove('testjson.json')
+
+server.shutdown()
+print('Server shutdown')
+
+# test ssl
+import socket
+import ssl
+hostname = 'www.google.com'
+context = ssl.create_default_context()
+with socket.create_connection((hostname, 443)) as sock:
+    with context.wrap_socket(sock, server_hostname=hostname) as ssock:
+        print(ssock.version())
 
 print('Reached end of file')
+
--- a/tests/standalone/Urllib3Using.py
+++ b/tests/standalone/Urllib3Using.py
@@ -1,6 +1,6 @@
 import urllib3
 
-# nuitka-skip-test-unless: urllib3
+# nuitka-skip-unless-imports: urllib3
 
 # testing request
 http = urllib3.PoolManager()
--- a/tests/standalone/run_all.py
+++ b/tests/standalone/run_all.py
@@ -291,7 +291,7 @@ for filename in sorted(os.listdir(".")):
     extra_flags = ["expect_success", "standalone", "remove_output"]
 
     # specials: these modules need custom checks(implemented below), skip automated checking
-    specials = {'TkInterUsing.py'}
+    specials = ['TkInterUsing.py']
     # skip each test if their respective required import is not present
     required_m = find_required_module(filename)
     if required_m and not filename in specials and not hasModule(required_m):
--- a/tests/standalone/run_all.py
+++ b/tests/standalone/run_all.py
@@ -21,6 +21,7 @@
 
 import os
 import sys
+import subprocess
 
 # Find nuitka package relative to us. The replacement is for POSIX python
 # and Windows paths on command line.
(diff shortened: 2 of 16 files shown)


tommyli3318:
@kayhayen Please check over the new mechanism I've implemented for run_all.py!

Also, Urllib3Using.py works fine using manual testing, but during automated testing the following error is raised:

Comparing output of 'Urllib3Using.py' using 'C:\Users\Tommy\AppData\Local\Programs\Python\Python37-32\python.exe' with flags silent, expect_success, standalone, remove_output ...
--- C:\Users\Tommy\AppData\Local\Programs\Python\Python37-32\python.exe (stderr)
+++ nuitka (stderr)

@@ -1 +1,2 @@

+Nuitka:WARNING:Unresolved '__import__' call at 'C:\Users\Tommy\AppData\Local\Programs\Python\Python37-32\lib\site-packages\urllib3\packages\six.py:82' may require use of '--include-plugin-directory' or '--include-plugin-files'.

Error, outputs differed.
Skipped, Win32ComUsing.py (win32com not installed for this Python version, but test needs it).

I understand that urllib3 is fully working and this is just a warning message for the six compatibility layer, so should I just suppress errors like these?

kayhayen:
As for the `__import__` warning consider this function:

https://github.com/Nuitka/Nuitka/blob/69183f6567befd62bc1d768dd070fb99475b8cb4/nuitka/importing/Importing.py#L154

I would say that `six` should be added here. Probably there is a bunch more of those, although I sort of feel that we should have to look and make sure that the line is not disguising hidden dependencies, but I am pretty confident I would know by now from user bug reports, so it feels unlikely.

tests/standalone/run_all.py
```diff
@@ -264,6 +264,15 @@
     "MFCM140U.DLL",
 )
 
+# finds the special comment in each test module and returns the import module needed for each test
+# special comments are the the following format:
+# "# nuitka-skip-test-unless: module"
+def find_required_module(filename):
```
kayhayen: Ought to be named `findRequiredModules`, we use came case for methods.

And yes, I think you should support plural to be safe, i.e. yield module names.

tests/standalone/run_all.py
```diff
@@ -264,6 +264,15 @@
...
+    with open(filename,'r') as f:
+        for line in f:
+            if '# nuitka-skip-test-unless: ' in line:
```
kayhayen: You made a mistake here, you are checking with 'in' for mere presence. I would argue, that is better than 'startswith', but then you use hardcoded slicing below, which presumes that it's at the beginning of the line.

tests/standalone/run_all.py
```diff
@@ -281,125 +290,79 @@
 
     extra_flags = ["expect_success", "standalone", "remove_output"]
 
-    if filename == "PySideUsing.py":
+    # specials: these modules need custom checks(implemented below), skip automated checking
+    specials = ['TkInterUsing.py']
+    # skip each test if their respective required import is not present
+    required_m = find_required_module(filename)
```
kayhayen: Please name it `required_module`, do not abbreviate things that hurt readability.

tests/standalone/run_all.py
```diff
@@ -281,125 +290,79 @@
...
+    if required_m and not filename in specials and not hasModule(required_m):
```
kayhayen: I think we ought to be able to do without `specials` and just don't `elif` after this.

Typically you use `elif` if you know branches are exclusive, but if they are not you don't mind, and make it two if statements, rather than repeating that knowledge in a condition.

tests/standalone/Urllib3Using.py
```diff
@@ -0,0 +1,24 @@
...
+r.data
+r.headers
+
+# testing JSON content
+import json
+r = http.request('GET', 'http://httpbin.org/ip')
+json.loads(r.data.decode('utf-8'))
+
+# testing binary content
+r = http.request('GET', 'http://httpbin.org/bytes/8')
```
kayhayen: Using external resources kind of is a tough issue for the tests. They tend to be flaky, causing errors at non opportune times.

May I suggest to instead use standard library HTTP server package, and listen to a specific port, then making queries there.

kayhayen: Also I would like to cover https:// protocol usage, as that is where for standalone the biggest challenges rest. You need the SSL libraries, and they often are having lots of dependencies that can be hidden. That way your test is likely to find something in some setups.

But also, Python may not be able to do SSL. I surely have seen versions that will give an exception. When that happens, just print it, much like many tests in `tests/basics` do, and print the exception, so we know we fail in the same way.

tommyli3318:
> As for the `__import__` warning consider this function:
> 
> [Nuitka/nuitka/importing/Importing.py](https://github.com/Nuitka/Nuitka/blob/69183f6567befd62bc1d768dd070fb99475b8cb4/nuitka/importing/Importing.py#L154)
> 
> Line 154 in [69183f6](/Nuitka/Nuitka/commit/69183f6567befd62bc1d768dd070fb99475b8cb4)
> 
>  def isWhiteListedImport(node): 
> I would say that `six` should be added here. Probably there is a bunch more of those, although I sort of feel that we should have to look and make sure that the line is not disguising hidden dependencies, but I am pretty confident I would know by now from user bug reports, so it feels unlikely.

tommyli3318:
> As for the `__import__` warning consider this function:
> 
> [Nuitka/nuitka/importing/Importing.py](https://github.com/Nuitka/Nuitka/blob/69183f6567befd62bc1d768dd070fb99475b8cb4/nuitka/importing/Importing.py#L154)
> 
> Line 154 in [69183f6](/Nuitka/Nuitka/commit/69183f6567befd62bc1d768dd070fb99475b8cb4)
> 
>  def isWhiteListedImport(node): 
> I would say that `six` should be added here. Probably there is a bunch more of those, although I sort of feel that we should have to look and make sure that the line is not disguising hidden dependencies, but I am pretty confident I would know by now from user bug reports, so it feels unlikely.

Do you want me to whitelist it? Or should we leave it for now? And you mean to add them as a special case check inside `isWhiteListedImport` right?

tests/standalone/run_all.py
tommyli3318: I understand, but the issue with TkInter was that the import which should be checked for depends on python version, "Tkinter" for python2 and "tkinter" for python3. In this case without the "specials", a test might be skipped altogether even if the requirement was met, and changing elif to if wouldn't help the issue. If we do this without specials Should we develop a new syntax for the special comment? e.g. `# nuitka-skip-unless: Tkinter || tkinter`, or something along the lines with that?

tommyli3318:
> Generally, I would suggest you allow Python code behind the marker, and then execute it in Python, seeing if it crashes (ImportError), so do like `hasModule` does, but use `sys.exit(__import__("flask")`, and the user would specifiy # nuitka-skip-test-unless: **import**("flask").

I understand what you mean, but need a little more elaboration. So shuold I make `findRequiredModules `so that it uses the approach you named above to check whether all the required imports exist, and return a boolean value, and skip the usage of `hasModule` altogether? And it would probably be renamed to `checkRequiredModules`, and it would just be called for every single test file. This would eliminate the need for using `hasModule`.

> Then it's not limited to modules, can be any expression, and of course can do checks in `sys.version` etc. as we see fit, and you would probably statically add `import os, sys` before the call. Let me know if I was clear enough.

So how would this work in the case of Tkinter, where it's "Tkinter" for python2 and "tkinter" for python3? Also would we need to change the syntax for our special comment? Refer to my reply above:

> I understand, but the issue with TkInter was that the import which should be checked for depends on python version, "Tkinter" for python2 and "tkinter" for python3. In this case without the "specials", a test might be skipped altogether even if the requirement was met, and changing elif to if wouldn't help the issue. If we do this without specials, should we develop a new syntax for the special comment? e.g. # nuitka-skip-unless: Tkinter || tkinter, or something along the lines of that?

kayhayen:
> the issue. If we do this without specials, should we develop a new syntax for the special comment? > e.g. # nuitka-skip-unless: Tkinter || tkinter, or something along the lines of that?

I would be fine with this:

`# nuitka-skip-unless:  __import__("Tkinter" if sys.version_info[0] < 3 else "tkinter")`

That means, the comment becomes code to evaluate. And it's run much like the import check currently is done with the "-c", and would have to be an expression, so you can feed it into a template of say `import sys, os; sys.exit(%s)` which we can use to tell if it passes. An exception (`ImportError here`) or a false expression can then both be detected from the exit code not being 0.

kayhayen:
Then we can throw away the imports checking, and only use this, and have `__import__("flask")` and the like as the simple solution. The test runner would then only concern itself with that kind of thing, or you have both, `# nuitka-skip-unless-expression` and `# nuitka-skip-unless-imports`, your pick.

tests/standalone/Urllib3Using.py
tommyli3318: I've tried using a standard library HTTP server, but the problem is that the script would be dedicated to running the server, and another script would be needed to make the queries. How can I do this without using multiprocessing or threading?

tommyli3318:
>or you have both, `# nuitka-skip-unless-expression` and `# nuitka-skip-unless-imports`, your pick.

Updated with the new `checkRequirements` method supporting both, please check over.
University resumes tomorrow for me, so I'll probably be less active until the weekends.

tests/standalone/Urllib3Using.py
kayhayen: I think threading ought to be OK for this test there. I don't think I want the test to be more than one file.

tommyli3318:
Updated Urllib3Using.py to use only on localhost, eliminating external source dependency. For SSL tests, should it be in a separate module?

kayhayen:
Fantastic job, Tommy, I will merge once I put out the stable release, so this doesn't bring up unexpected failures. Most probably later today.

kayhayen:
As for SSL, I would use the same test. Using 2 tests will incur too much cost in compile times.

tests/standalone/run_all.py
```diff
@@ -264,6 +265,35 @@
     "MFCM140U.DLL",
 )
 
+# checks requirements needed to run each test module, according to the specified special comment
+# special comments are in the following formats:
+#     "# nuitka-skip-unless-expression: expression to be evaluated"
+#       OR
+#     "# nuitka-skip-unless-imports: module1,module2,..."
+def checkRequirements(filename):
+    with open(filename,'r') as f:
```
kayhayen: One last quibble, can you make use of `getFileContentByLine` from `nuitka.Utils.FileOperations`, because that avoids the deep nesting here.

tommyli3318: Done. Anything else?

kayhayen:
I am very excited about this one. Really liking how this simplifies the Top 50 task by allowing to not have to edit the `run_all.py` for every package.

Thanks for your contribution!

tab1tha:
@tommyli3318 this has helped me a lot. Thanks for your contribution!

PR #315, opened 2019-03-28
[WIP] Specialize Helper BinaryAdd.py

riya-17:
### What does this PR do?
This PR is about generating specialized Binary Addition code using Templating language Jinja2. @kayhayen Please do Review this PR, I would like to know the changes I can make further to improve it.

It is part of GSoC 2019 application. Relevant issue is #283 

### PR Checklist
- [X] Integer Addition
- [X] Float Addition
- [ ] Object Addition
- [ ] Different combination of Data types Addition like int, long or int, float etc.

### DIFF
--- a/nuitka/tools/gen_helper_code/BinaryAdd.py
+++ b/nuitka/tools/gen_helper_code/BinaryAdd.py
@@ -1,0 +1,155 @@
+import subprocess
+import os
+from jinja2 import Template
+
+
+def isType(op1, op2):
+	# Operands are compared and checked for their type and value is returned
+	# It checks for operand to be int, float or object
+	if type(op1) == int:
+		if type(op2) == int:
+			return 1
+	elif type(op1) == float:
+		if type(op2) == float:
+			return 0
+	else:
+		return "PyInt_Check(%s)" % op1
+
+
+def add(operand1, operand2):
+	# operands are passed for data type check to isType()
+	operand_type = isType(operand1, operand2)
+	a = str(operand1)
+	b = str(operand2)
+	t = str(operand_type)
+
+	# Subprocess call is used to run C code with passed values 
+	if not os.path.exists('binary_add'):
+		f = open('binary_add.c', 'w')
+		f.write(prog)
+		f.close()
+		subprocess.call(["gcc", "binary_add.c"])
+	temp = subprocess.call(["./a.out", a, b, t])
+	# Summed-up values recieved from C code are printed and then returned to jinja2
+	print("Sum of two operands is : ", temp)
+	return temp
+
+
+prog = r"""
+#include<stdio.h>
+#include<ctype.h>
+#include <string.h>
+
+int sum_int(int operand1, int operand2)
+{
+	int i = 0, a, b;
+	a = operand1;
+	b = operand2;
+
+	// adding both Integer operands
+	i = a + b;
+
+	return (i);
+}
+
+float sum_float(float operand1, float operand2)
+{
+	float i = 0.0, a, b;
+	a = operand1;
+	b = operand2;
+
+	// adding both Float operands
+	i = a + b;
+
+	return (i);
+}
+
+int toInteger(char a[]) {
+	int c, sign, offset, n;
+	
+	if (a[0] == '-') {  // Handle negative integers
+		sign = -1;
+	}
+	if (sign == -1) {  // Set starting position to convert
+		offset = 1;
+	}
+	else {
+		offset = 0;
+	}
+	
+	n = 0;
+
+	for (c = offset; a[c] != '\0'; c++) {
+		n = n * 10 + a[c] - '0';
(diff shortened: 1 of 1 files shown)


nuitka/tools/gen_helper_code/BinaryAdd.py
```diff
@@ -0,0 +1,155 @@
+import subprocess
+import os
+from jinja2 import Template
+
+
+def isType(op1, op2):
```
kayhayen: Not all branches of this function return, never a good sign

nuitka/tools/gen_helper_code/BinaryAdd.py
```diff
@@ -0,0 +1,155 @@
...
+
+def add(operand1, operand2):
+	# operands are passed for data type check to isType()
+	operand_type = isType(operand1, operand2)
+	a = str(operand1)
+	b = str(operand2)
+	t = str(operand_type)
+
+	# Subprocess call is used to run C code with passed values 
+	if not os.path.exists('binary_add'):
```
kayhayen: This will never exist, but if it did, you wouldn't update your binary, which would be bad.

nuitka/tools/gen_helper_code/BinaryAdd.py
```diff
@@ -0,0 +1,155 @@
...
+		f = open('binary_add.c', 'w')
+		f.write(prog)
+		f.close()
+		subprocess.call(["gcc", "binary_add.c"])
```
kayhayen: There is a `-o` option to gcc to control output filename, `a.out` is the default

nuitka/tools/gen_helper_code/BinaryAdd.py
```diff
@@ -0,0 +1,155 @@
...
+	print("Sum of two operands is : ", temp)
+	return temp
+
+
+prog = r"""
+#include<stdio.h>
+#include<ctype.h>
+#include <string.h>
+
+int sum_int(int operand1, int operand2)
```
kayhayen: This is not too impressive C code, `return operand1+operand2;` would be very sufficient.

nuitka/tools/gen_helper_code/BinaryAdd.py
```diff
@@ -0,0 +1,155 @@
...
+{
+	int i = 0, a, b;
```
kayhayen: As a general principle, do not initialize variables that are supposed to be assigned. You will hide mistakes made later, e.g. not assigning in all branches from the C compiler that can give you a warning. Use -Wall when compiling, and allow for it to find such bugs. The assignments of `i` in your function are potentially bug hiding and optimized away in the good case.

nuitka/tools/gen_helper_code/BinaryAdd.py
```diff
@@ -0,0 +1,155 @@
...
+	a = operand1;
+	b = operand2;
+
+	// adding both Float operands
+	i = a + b;
+
+	return (i);
+}
+
+int toInteger(char a[]) {
```
kayhayen: Checkout C standard library functions, e.g. `atoi` that does this.

nuitka/tools/gen_helper_code/BinaryAdd.py
```diff
@@ -0,0 +1,155 @@
...
+	for (c = offset; a[c] != '\0'; c++) {
+		n = n * 10 + a[c] - '0';
+	}
+	if (sign == -1) {
+		n = -n;
+	}
+	return n;
+}
+
+float ToFloat(char *string)
```
kayhayen: I am pretty sure standard C library has this too.

kayhayen:
So you wrote a Python program that uses Jinja to call the C compiler to execute a static C program, which is not nearly what the task is about, but it's closer than the last times.

Notice how you have these 2 functions to add types. If these were taking `PyObject *` as arguments, and if their content was generated by Jinja templating, it would be closer to what the task is about. But in this form, Jinja is not actually used to influence the C code at all. It however is meant to produce it.

Actually, generating those specializations of `BINARY_OPERATION_ADD_OBJECT_OBJECT` is all we are looking at for a start. I am willing to assume that my description of the issue to solve is too bad and needs to be improved. This may however take time. It feels we are not making good progress right now, I am also concerned that your understanding of C is not sufficient.

riya-17:
@kayhayen So what should I aim for at present. Should I work on the flaws of the code and implement the changes you asked for?

kayhayen:
I don't know @riya-17 as the insufficient C knowledge will mean, that there are missing pre-requisites for the task, which is a shame (of mine), because we had discussed before, and I failed to recognize that fact. My mistake there is that I should have tested your self assessment, which I did not. 

I don't think we can rapidly change this, as in you re-learning C, therefore I recommend to find another idea that does not require C at all.

PR #317, opened 2019-03-30
Change "numpy-plugin" to just "numpy"

JorjMcKie:
Also added support for scipy as an option to this plugin.
Copying any additional binaries has been moved to the end of the compile.

### What does this PR do?
- Changed `plugin_name` from "numpy-plugin" to "numpy" following a recommendation of @kayhayen.
- Added scipy support as an option like this: ``--enable-plugin=numpy=scipy``.
- Also added copying undetectable scipy DLLs (Windows only), similar to what had been done already for numpy.
- Moved the process of copying binaries: this will now happen when the distribution folder is otherwise finished (i.e. in `onStandaloneDistributionFinished`).

Scipy will not work without numpy installed. Therefore, scipy should be treated as a numpy add-on. A separate scipy plugin makes no sense, while numpy without including scipy must remain possible.

### PR Checklist

### DIFF
--- a/nuitka/plugins/standard/NumpyPlugin.py
+++ b/nuitka/plugins/standard/NumpyPlugin.py
@@ -198,23 +198,24 @@ class NumpyPlugin(UserPluginBase):
 
     plugin_name = "numpy"  # Nuitka knows us by this name
 
-    def __init__(self):
-        self.files_copied = False  # ensure one-time action
-        self.scipy = self.getPluginOptionBool("scipy", False)
-
-    def onStandaloneDistributionFinished(self, dist_dir):
+    def considerExtraDlls(self, dist_dir, module):
         """ Copy extra shared libraries for this numpy / scipy installation.
 
         Args:
             dist_dir: the name of the program's dist folder
+            module: module object (not used here)
         Returns:
-            None
+            empty tuple
         """
+        full_name = module.getFullName()
+        if full_name not in ("numpy", "scipy"):
+            return ()
 
-        binaries = get_numpy_core_binaries()
-        bin_total = len(binaries)  # anything there at all?
-
-        if bin_total > 0:
+        if full_name == "numpy":
+            binaries = get_numpy_core_binaries()
+            bin_total = len(binaries)  # anything there at all?
+            if bin_total == 0:
+                return ()
             info("")
             info(" Copying extra binaries from 'numpy' installation:")
             for f in binaries:
@@ -222,7 +223,6 @@ class NumpyPlugin(UserPluginBase):
                 idx = bin_file.find("numpy")  # this will always work (idx > 0)
                 back_end = bin_file[idx:]  # => like 'numpy/core/file.so'
                 tar_file = os.path.join(dist_dir, back_end)
-                # info(" " + bin_file)
 
                 # create any missing intermediate folders
                 if not os.path.exists(os.path.dirname(tar_file)):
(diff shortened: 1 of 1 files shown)


nuitka/plugins/standard/NumpyPlugin.py
```diff
@@ -173,47 +196,72 @@ class NumpyPlugin(UserPluginBase):
...
-    plugin_name = "numpy-plugin"  # Nuitka knows us by this name
+    plugin_name = "numpy"  # Nuitka knows us by this name
 
     def __init__(self):
         self.files_copied = False  # ensure one-time action
+        self.scipy = self.getPluginOptionBool("scipy", False)
 
-    def considerExtraDlls(self, dist_dir, module):
-        """ Copy extra shared libraries for this numpy installation.
+    def onStandaloneDistributionFinished(self, dist_dir):
```
kayhayen: This is too late, at the time this is called, dependency analysis has already been done, and e.g. rpaths, etc. may still be wrong. On MacOS this seems very common thing. Also other plugins `onStandaloneDistributionFinished` would rather want to see the already completed thing I guess and not help in finishing it.

Basically I think, we need a pattern, where there is only one module that triggers the extra DLLs, and otherwise nothing is done. Like I changed to Tk plugin:

```
        if not _isTkInterModule(module):
            return ()

        if not isWin32Windows():  # if not Windows notify wrong usage once
            info("tkinter plugin supported on Windows only")
            return ()

```

Maybe we allow plugins to overload one method, e.g. `isResponsibleFor(module)` and then have default implementations that make sure to forward the calls only if that returns `True` and allows to implement this easily.

I believe you switching to that method is mainly driven by the fact that you do not want to be called per module, but only once. But I also kind of like the idea to enable a plugin, and maybe have it complain in the end that it was never relevant, but to not do anything unless you have `import numpy`.

nuitka/plugins/standard/NumpyPlugin.py
```diff
@@ -223,7 +271,7 @@ class NumpyPluginDetector(UserPluginBase):
         We are given the chance to issue a warning if we think we may be required.
     """
 
-    plugin_name = "numpy-plugin"  # Nuitka knows us by this name
```
kayhayen: I agree, that name is indeed only confusing.

JorjMcKie:
oh of course you are right! Stupid me.
I'll make the change.
There is no way but having this done underway, on a per module level and maintain a switch in the plugin once the copy has been done once.
It used to be like that, but I was carried away by the thought to save processing time by making it a once-only thing.

JorjMcKie:
btw I have noticed a few hours ago, that the recent changes to tk-inter plugin have no more any such switch, and that the plugin is now trying to copy the data more than once?

JorjMcKie:
I wasn't clear enough in my last post:
checking for "tkinter" module is not enough, because this module may be handed in several times or is that wrong?

JorjMcKie:
When testing my other idea with the profile guided module inclusion, I am being asked several times for the same module (in `onModuleEncounter`) - independent from having said NO before ...

JorjMcKie:
Just tested the tk-inter again - it does work in the form you have changed it, sorry about the confusion.

I was confused myself by this:
I have a PYD which imports tkinter internally. In the test script there is no independent import of tkinter (and even if it were, it were too late).
This shared library will of course not be detected by the test `_isTkInterModule` as being a secret incarnation of tkinter ...
So the TCL/TK copy never took place and I started following a wrong path of analysis.

JorjMcKie:
Sorry for bombarding you with piece-wise thoughts and questins:

Are you saying, that `considerExtraDlls` is being called ever only once per every module, but currently also once per active plugin?

But it intentionally may happen, that `onModuleEncounter` is called several times with the same module?

JorjMcKie:
I hope this is better now.
Please ignore most of my noise above. Except for one concern:
If I have a shared library, e.g. `PySimpleGUI.pyd` which imports tkinter, and which is the first (or only) module which does that, then I need to handle this in the tkinter plugin ...

kayhayen:
@JorjMcKie It's a pleasure to merge this. We took the discussion about API change offline, and I think should be a separate issue anyway.

PR #322, opened 2019-03-31
Test Nuitka support for package Six

Pihu1998:
This is a test to check if PyPI package Six is supported by Nuitka

### DIFF
--- a/tests/standalone/SixUsing.py
+++ b/tests/standalone/SixUsing.py
@@ -1,0 +1,41 @@
+import six
+import sys
+import unittest
+
+
+def dispatch_types(value):
+   
+    if isinstance(value, six.integer_types):
+        return int
+
+    elif isinstance(value, six.string_types):
+        return str
+    
+    elif isinstance(value, six.binary_type):
+        return bytes
+
+def get_iterator(dictionary):
+    return six.viewkeys(dictionary), six.viewvalues(dictionary), six.viewitems(dictionary)
+
+
+class TestAssertCountEqual(unittest.TestCase):
+    def test(self):
+        six.assertCountEqual(self, (1, 2), [2, 1])
+        
+
+def main():
+    print(dispatch_types(6))
+    print(dispatch_types('hi'))
+    u='hi'
+    print(dispatch_types(u.encode('utf-8')))
+    Dict = {'Tim': 18,'Charlie':12,'Tiffany':22,'Robert':25} 
+    print(get_iterator(Dict))
+    
+    t=TestAssertCountEqual()
+    print(t.test)
+   
+
+if __name__ == "__main__":
+    main()
+
+
--- a/tests/standalone/six.py
+++ b/tests/standalone/six.py
@@ -4,24 +4,22 @@ import sys
 
 MAXSIZE = sys.maxsize
 
-def test_dispatch_types(value):
+def dispatch_types(value):
     assert isinstance(six.MAXSIZE + 23, six.integer_types)
     assert isinstance(six.u("hi"), six.string_types)
-    
     class X(object):
         pass
-    
+
     assert isinstance(X, six.class_types)
 
-    
 
-def test__import_module(name):
+def import_module(name):
     from logging import handlers
     m = six._import_module("logging.handlers")
     assert m is handlers
 
 
-def test__add_doc(func, doc):
+def add_doc(func, doc):
     def f():
         """Icky doc"""
         pass
--- a/tests/standalone/six.py
+++ b/tests/standalone/six.py
@@ -1,28 +1,58 @@
 from __future__ import absolute_import
 import six
 import sys
+from logging import handlers
 
-MAXSIZE = sys.maxsize
+
+PY3 = sys.version_info[0] == 3
+if PY3:
+ string_types = str,
+ integer_types = int,
+ class_types = type
+
 
 def dispatch_types(value):
-    assert isinstance(six.MAXSIZE + 23, six.integer_types)
-    assert isinstance(six.u("hi"), six.string_types)
-    class X(object):
-        pass
+   
+    if isinstance(value, six.integer_types):
+        print(value)
 
-    assert isinstance(X, six.class_types)
-
+    elif isinstance(value, six.string_types):
+        print(value)
+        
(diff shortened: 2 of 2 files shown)


tests/standalone/six.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import absolute_import
+import six
+import sys
+
+MAXSIZE = sys.maxsize
+
+def test_dispatch_types(value):
```
kayhayen: The Nuitka test runner is executing these as programs, with CPython and Nuitka, and will check the outputs, and merely defining a function named `test_*` is not going to do anything to run it. You can put stuff in functions, but then you have to also call it. Most often for standalone tests, we do just put things on module level.

Pihu1998:
@kayhayen  Please review the above changes and please inform if there is any room for further improvements.

tests/standalone/six.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import absolute_import
+import six
+import sys
+
+MAXSIZE = sys.maxsize
+
+def dispatch_types(value):
```
kayhayen: Well, not being named `test_*` still doesn't make the code being called, you realize that?

Pihu1998:
@kayhayen Please review the above changes. I have included a main() function that calls all of the functions. Please inform if it can be improved further.

tests/standalone/six.py
```diff
@@ -0,0 +1,58 @@
...
+PY3 = sys.version_info[0] == 3
+if PY3:
+ string_types = str,
+ integer_types = int,
+ class_types = type
+
+
+def dispatch_types(value):
+   
+    if isinstance(value, six.integer_types):
```
kayhayen: This is not a big coverage of the six package, is it?

tests/standalone/six.py
```diff
@@ -0,0 +1,58 @@
+from __future__ import absolute_import
```
kayhayen: You need that absolute_import only due to your strange naming of the test as "six", why not go with SixUsing.py to match what other tests have.

tests/standalone/six.py
```diff
@@ -0,0 +1,58 @@
...
+def dispatch_types(value):
+   
+    if isinstance(value, six.integer_types):
+        print(value)
+
+    elif isinstance(value, six.string_types):
+        print(value)
+        
+    elif isinstance(value, six.class_types):
+        value
```
kayhayen: That's not doing anything.

tests/standalone/six.py
```diff
@@ -0,0 +1,58 @@
...
+ class_types = type
+
+
+def dispatch_types(value):
+   
+    if isinstance(value, six.integer_types):
+        print(value)
+
+    elif isinstance(value, six.string_types):
+        print(value)
```
kayhayen: That's not distinguishable from above branch.

tests/standalone/six.py
```diff
@@ -0,0 +1,58 @@
...
+        
+    elif isinstance(value, six.class_types):
+        value
+        
+
+def import_module(name):
```
kayhayen: Why is this here?

tests/standalone/six.py
```diff
@@ -0,0 +1,58 @@
...
+    __import__(name)
+    return sys.modules[name]
+    
+
+def add_doc(func, doc):
```
kayhayen: This is not using the doc parameter, what is it intended for?

kayhayen:
@Pihu1998 Also do not just push new code, instead also respond to review comments I make, I failed to notice this one had changed, but assumed you didn't work on it.

tests/standalone/six.py
Pihu1998: Okay, I think I should expand the usage of Six specifically.

tests/standalone/six.py
Pihu1998: Yeah, Sure!

tests/standalone/six.py
Pihu1998: Oh sorry, my bad. I will remove it.

kayhayen:
Well, maybe I was not precise enough. Respond to Github review comments right after you addressed them. Otherwise respond to ask a question about the comment or an outright refutation of it. But do not acknowledge, promise to make changes, or apologize.

I would much rather prefer to see a comment mean there is new code, or information I sought, so I can resolve the issue. Now there will be comments with responses, and I am left guessing when it makes sense to look at the code. That was not the point of asking you to reply, just to have something there, but rather to learn about what you have addressed.

Pihu1998:
Sorry for the late push of code. I was working on it. I have added some more usage of the package Six in the code and also changed the test file name to 'SixUsing.py'. Please review the above changes.

tests/standalone/SixUsing.py
```diff
@@ -0,0 +1,41 @@
...
+    def test(self):
+        six.assertCountEqual(self, (1, 2), [2, 1])
+        
+
+def main():
+    print(dispatch_types(6))
+    print(dispatch_types('hi'))
+    u='hi'
+    print(dispatch_types(u.encode('utf-8')))
+    Dict = {'Tim': 18,'Charlie':12,'Tiffany':22,'Robert':25}
```
kayhayen: Just a minor quibble, don't name local variables this way, makes me think it's a class or when I see it used.

As for the prints, import the `print_function` from future, and add some identifying prefixes, e.g. `dispatch_types int:`, the so output is more human readable. It works the way you did it, but it's not nice.

tests/standalone/six.py
kayhayen: Won't you delete this file then, it seems you forgot about that?

tests/standalone/SixUsing.py
```diff
@@ -0,0 +1,41 @@
...
+        return str
+    
+    elif isinstance(value, six.binary_type):
+        return bytes
+
+def get_iterator(dictionary):
+    return six.viewkeys(dictionary), six.viewvalues(dictionary), six.viewitems(dictionary)
+
+
+class TestAssertCountEqual(unittest.TestCase):
```
kayhayen: So, is this is a six extension for unit tests here?

kayhayen:
Unlikely to be finished.

PR #323, opened 2019-03-31
Cover more trick assignments

ddatsko:
### What does this PR do?
I added some more tests to Nuitka/tests/basics/TrickAssignments32 where was a comment with TODO. Also made "__int\__" method return anobject of type ```int``` instead of ```bool``` as I was getting warning while running tests.
### Why was it initiated? 
I just wanted to make a contribution to take one of Nuitka's projects as my GSoC project and while examining code I saw this TODO comment and found out that this job is suitable for me.(I'm sorry if that was the job of someone else, but the file haven`t been changed for 26 days, so I decided to try)

### PR Checklist

### DIFF
--- a/tests/basics/TrickAssignments32.py
+++ b/tests/basics/TrickAssignments32.py
@@ -1152,28 +1152,25 @@ except UnboundLocalError:
 else:
     print("Ouch.!")
 
-# Got segmantation fault (core dumped) while using nuitka
 
-# def someFunctionThatReturnsDeletedValueViaReversed():
-#     class C:
-#         def __reversed__(self):
-#             nonlocal a
-#             del a
-#             return None
+def someFunctionThatReturnsDeletedValueViaReversed():
+    class C:
+        def __reversed__(self):
+            nonlocal a
+            del a
+            return None
+    a = 1
+    c = C()
 
-#     c = C()
+    reversed(c)
+    return a
 
-#     a = 1
-#     reversed(c)
-#     return a
-
-
-# try:
-#     someFunctionThatReturnsDeletedValueViaReversed()
-# except UnboundLocalError:
-#     print("OK, object reversed correctly deleted an item.")
-# else:
-#     print("Ouch.!")
+try:
+    someFunctionThatReturnsDeletedValueViaReversed()
+except UnboundLocalError:
+    print("OK, object reversed correctly deleted an item.")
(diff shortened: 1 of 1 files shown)


tests/basics/TrickAssignments32.py
```diff
@@ -510,7 +510,7 @@ def __int__(self):
             nonlocal a
             del a
 
-            return False
```
kayhayen: What warning were you talking about? I tended to purposefully no return booleans, as that must work as well.

tests/basics/TrickAssignments32.py
```diff
@@ -1047,4 +1047,172 @@ def __bytes__(self):
     print("Ouch.!")
 
 
+def someFunctionThatReturnsDeletedValueViaEq():
```
kayhayen: Can you do the same to `__lt__`, `__lte__`, `__gt__`, `__gte__`, `__neq__` (spelling ?) too?

Yours,
Kay

tests/basics/TrickAssignments32.py
ddatsko:
```
DeprecationWarning: __int__ returned non-int (type bool).  The ability to return an instance of a strict subclass of int is deprecated, and may be removed in a future version of Python.
  c = C()
```
This is the message I get when using both Nuitka and default Python interpreter.
```
Python 3.7.3
```

tests/basics/TrickAssignments32.py
ddatsko: Sure, I will do this

tests/basics/TrickAssignments32.py
ddatsko: @kayhayen , so, should I change this back or its OK?

kayhayen: I meant to say something about this, the test runner ignores warnings. The test code often does bad things, but Nuitka is not about compiling good code, but one that actually exists. You ought to use `./run_all.py filename` to execute the test and get a comparative diff. To create the test, you should use standard Python first, make that work, then compare.

kayhayen: And yes, please revert it, as it actually lowers test coverage.

ddatsko: OK, thank you for explanation

Issue #229, opened 2019-01-26
GSOC 2019: Make sure top 50 PyPI packages are supported


kayhayen:
Nuitka works with most software. The aim of this project is to make sure it's true for the top 50 packages on PyPI, by compiling and using their example codes.

In a first stage, you would identify and report the issues to the bug tracker, in a second stage develop tools that help to narrow down issues, e.g. what extension module fails to load precisely, even with a segfault happening, and put them to use and try to fix a few of the simpler issues.

Setting up these as automated tests would be the ultimate goal, so we can follow these top 50 packages with Nuitka over time and make sure they continue to work.

In the past it has happened e.g. that Jinja2 was breaking for Python3.7, and it would be cool to discover this immediately.

Skills: Python programming, pip installation, Linux and/or Windows installs of Python, one is good, both would be great.

aditya-hari:
Hey. 
 
This is something  that sounds really interesting to me and I'd love to be able to contribute to this. 
But I am completely new to Open source as a whole and have no idea how to even begin. I don't expect to be able to do something significant enough to qualify for GSoC or whatever but I'd love for this to be my first foray into open source and to gain experience.  
 
Since this and #230 both are labelled as easy in terms of difficulty, I think I should start here. 
 
But I don't know what "Linux and/or Windows installs of Python" means in terms of skills and I don't know how to begin.  
 
Could you please guide me to some resources or tell me how I should go about finding my bearings? I am competent with Python and consider myself to be a quick learner, I am sure I will be able to get up to speed soon.

kayhayen:
Welcome @aditya-hari and I agree this is not yet detailed enough. Actually today I was going to flesh this further out with pointers.

As for the skill, more of an asset. We would prefer you did this on those platforms, doing it on macOS would e.g. be too difficult for us to help you with. We do not care if you use Windows or Linux, however in my experience the later works better for the debugging that will be needed, i.e. in case of a crash of standalone compiled software.

As for more pointers, they are coming later today, or so I hope.

kayhayen:
So, first of all, we need to define what we mean by top 50 PyPI packages. For Python downloading the packages via `pip` is very widespread. So the first task would be to look at the top 100 and identify 50 packages that are even relevant to Nuitka. 

Of course `pip` itself, leading the list, we will skip that, nobody needs that compiled or uses it as a library in their software. But urllib3 is top 2, and very important piece of software that people rely on, so breaking it in Nuitka compilation would be bad.

[...]

PR #324, opened 2019-04-02
[TEST]Nuitka standalone test for Click PyPI package

tab1tha:
Click is a python package that supports arbitrary nesting
of commands, automatic help generation and lazy loading
of subcommands at runtime.

The Click package is tested by checking that click.option()
and click.echo() work as expected

This commit contributes to https://github.com/Nuitka/Nuitka/issues/229

Signed-off-by: tab1tha <tambetachere@gmail.com>

### PR Checklist


tests/standalone/ClickUsing.py
```diff
@@ -0,0 +1,16 @@
+import click
+
+# nuitka-skip-unless-imports: click
+
+@click.command()
+#@click.option("--count", default = 1, help = "Number of greetings.")
+#@click.option("--name", prompt = "Your name", help = "The person to greet.")
+
+def hello(count = 5, name = "Tabi"):
```
xennygrimmato: In the method signature, there should be no spaces between the variable name, equals symbol and the value.

tests/standalone/ClickUsing.py
```diff
@@ -0,0 +1,16 @@
+import click
+
+# nuitka-skip-unless-imports: click
+
+@click.command()
+#@click.option("--count", default = 1, help = "Number of greetings.")
```
xennygrimmato: We should ideally have some options configured. Any specific reason to comment them out?

tests/standalone/ClickUsing.py
```diff
@@ -0,0 +1,16 @@
...
+#@click.option("--name", prompt = "Your name", help = "The person to greet.")
+
+def hello(count = 5, name = "Tabi"):
+	"""Simple program that greets NAME for a total of COUNT times."""
+	for _ in range(count):
+		click.echo("Hello, %s!" % name)
+
+
+if __name__ == '__main__':
+	hello()
```
xennygrimmato: The way click should be tested is by passing some options as CLI args while invoking the program.
Example: `python ClickUsing.py --count=3`. Can you think of a way to simulate that behaviour here?

Our goal here should be to check if click can parse those args correctly and allocate correct values to the options.

tests/standalone/ClickUsing.py
tab1tha: Those options require that the user input his/her name and the value of count.When I ran it that way, Travis CI failed with an error that indicated that no output was received: since "name" and "count" had not been set.
Hence, I went ahead to put the values of "name" and "count" in the test such that the hello() function will have its desired parameters.

tests/standalone/ClickUsing.py
tab1tha: I'll do this and commit.

tests/standalone/ClickUsing.py
kayhayen: May I suggest that you enhance the test runner, to allow passing of arguments. You could change it to using `optparse` (or do you want to try click?) and instead of making sys.argv manual `in` tests, it would e.g. accept --with-command-args="...." and that would be used in `run_all.y`.

@tab1tha Do you know what I am talking about, I mean the binary behind / called in `nuitka.tools.testing.Common.compareWithCPython`. Do you think you can do this? It's actually a shame.

I guess you would just convert the "extra flags" in that function to proper command line options of the tool, and not have to change much other uses at all.

That way, we would test click properly, as well as command line handling for standalone binaries itself. I don't think anything really covers the correctness and compatibility of `sys.argv` in Nuitka for that matter.

tests/standalone/ClickUsing.py
tab1tha: I think the next commit solves this. Please check if it does.

tests/standalone/ClickUsing.py
tab1tha: Thanks for the suggestion @kayhayen . When I read further into click's documentation, I found out that it has its own runner!. It is called CliRunner and it invokes commands as command line scripts. I used it, in the new commit and it works . Is this satisfactory?

kayhayen: @tab1tha I don't think it is useful to use `clirunner` in the test.. We would like to see you manage to improve and extend the test runner in the described way. Do you think you can do it? If you want to use `click` instead of `optparse` for the argument handling in the Nuitka test runner, fine, but we would need to see it.

tests/standalone/ClickUsing.py
```diff
@@ -0,0 +1,19 @@
+import click
+from click.testing import CliRunner
+
+# nuitka-skip-unless-imports: click, CliRunner
```
kayhayen: I think this will not work, probably should be `click.testing.CliRunner`

tests/standalone/ClickUsing.py
```diff
@@ -0,0 +1,19 @@
...
+
+def test_hello_world():
```
kayhayen: Nothing executes these kinds of functions unless they are called.

tests/standalone/ClickUsing.py
```diff
@@ -0,0 +1,19 @@
...
+	'''This function uses the CliRunner.invoke()
+	method which runs the command line script in
+	isolation and captures the output in both bytes and binary data'''
+	@click.command()
+	@click.argument('name')
+	def hello(name):
+		click.echo('Hello %s!'%name)
+
+	runner = CliRunner()
+	result = runner.invoke(hello,['Peter'])
```
kayhayen: That is not interesting. I think @xennygrimmato told you already, we would like you to work on the `sys.argv` of the compiled program, making it use click methods to execute command line argument dependent behavior. The ``CliRunner` seems to have no point. We didn't compile it, so why bother.

tab1tha:
In the new commit, I have succeeded to simulate the `python ClickUsing.py --count=3` as described by @xennygrimmato . It is able to parse multiple commits at once.Is this satisfactory?

tab1tha:
I am now working on integrating command line argument parsing into the test runner.Can I open another issue and pull request for this? It is out of the scope of this Pull request.

tab1tha:
@kayhayen, @xennygrimmato  I have implemented your suggestions in the new commit; command line argument parsing in the runner and it works when run on ClickUsing.py. Also, the arguments can be passed in the test while invoking the program.
I was not quite sure,  however, whether to fit the function directly in `run_all.py` or to put it in `nuitka.tools.testing.Common`. I put it in `run_all.py`, is it okay here?.
Another thing is, I created a new function to prevent possibly breaking compatibility by changing `extra_flags` in `nuitka.tools.testing.Common.compareWithCPython` due to my not-yet-complete knowledge of the full scope of the project. Is this possibility real? If it is not, I can change the `extra_flags` as required.
I used a variable('commands`) to indicate the presence of command line arguments. I plan to document this if you confirm.
I look forward to your suggestions.

tests/standalone/run_all.py
```diff
@@ -301,6 +304,24 @@ def checkRequirements(filename):
     return (True, "")
 
 
+def ParseCommandLineArguments(filename):
```
xennygrimmato: Method names should follow snake_case.

tests/standalone/run_all.py
```diff
@@ -301,6 +304,24 @@ def checkRequirements(filename):
...
+    The command line arguments should be presented in the form of a
+    string assigned to a variable named commands. The assignment should be in the form: 
+	    commands='command1 command2=87 command3 --command4'
+	The commands may be positional or optional and may or may not be
+	assigned to values, as the case may be.
+	'''
+    for line in getFileContentByLine(filename):
+        if line.strip().startswith('commands'):
+            parser=argparse.ArgumentParser()
+            parser.parse_args(commands.split()) # At this point, the global namespace
```
xennygrimmato: Preferable formatting:
```
# Comment line
# Comment line
# Comment line
<python-statement>
```
In this case though, the comment can follow after the statement, but should not start on the statement's line itself.

tests/standalone/run_all.py
xennygrimmato: Also, where is this method called? I just want to understand its usage and try running it.

xennygrimmato: Specifically, I am interested in understanding why this method returns nothing and what's exactly happening here. I don't understand it fully.

tab1tha: > Also, where is this method called? I just want to understand its usage and try running it.

When I run the tests, the runner run_all.py automatically implements this method. That is what I observed when I used the runner on ClickUsing.py.

tests/standalone/run_all.py
tab1tha: okay. Thanks. I just learnt that now. it will be in my next commit

tests/standalone/run_all.py
tab1tha: > Specifically, I am interested in understanding why this method returns nothing and what's exactly happening here. I don't understand it fully.

parser.parse_args() will take the arguments provided when the program is run and interpret them according to the arguments that were added to the ArgumentParser object( in the code of the standalone test). So it assigns the CLI arg values to their corresponding variables and these assignments will exist in the global namespace and can be picked up by the standalone test. 
So it automatically returns these assignments to the global namespace. Hence there was no need to manually return anything.

tab1tha: > Method names should follow snake_case.

okay. I'll implement that in the next commit. Thank you for the correction.

tab1tha:
@tommyli3318 has been assigned to the issue that this pull request addresses. Hence I will close this pull request and halt my work on it. However, I am available to work on it and/or answer any questions pertaining to it as need be.

PR #336, opened 2019-04-05
Quality: Fixed the TypeError in autoformat 

### LINKED ISSUE #334: Autoformat of Nuitka source code gives TypeError and UnicodeDecodeError
I'm trying to autoformat my code. On executing without any flags I get `TypeError`
```
$ ./bin/autoformat-nuitka-source 
Working on: ['bin', 'nuitka']
Consider bin/autoformat-nuitka-source: Traceback (most recent call last):
  File "./bin/autoformat-nuitka-source", line 34, in <module>
    main()
  File "./nuitka/tools/quality/autoformat/__main__.py", line 100, in main
    autoformat(filename, git_stage=False, abort=options.abort)
  File "./nuitka/tools/quality/autoformat/Autoformat.py", line 332, in autoformat
    output_file.write(old_code)
TypeError: a bytes-like object is required, not 'str'

```

I have also tried to autoformat more than one dir, I get `UnicodeDecodeError`.
```
$ ./bin/autoformat-nuitka-source nuitka testsWorking on: ['nuitka', 'tests']
  File "./bin/autoformat-nuitka-source", line 34, in <module>
    main()
  File "./nuitka/tools/quality/autoformat/__main__.py", line 94, in main
    scanTargets(positional_args, (".py", ".scons", ".rst", ".txt"))
  File "./nuitka/tools/quality/ScanSources.py", line 74, in scanTargets
    for value in addFromDirectory(positional_arg, suffixes, blacklist):
  File "./nuitka/tools/quality/ScanSources.py", line 61, in addFromDirectory
    shebang = getShebangFromFile(fullpath)
  File "./nuitka/utils/Shebang.py", line 70, in getShebangFromFile
    return getShebangFromSource(f.readline())
  File "/home/bksahu/anaconda3/envs/nuitka/lib/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa2 in position 41: invalid start byte

```

**python -m nuitka --version**
0.6.4rc1
Python: 3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
Executable: /home/bksahu/anaconda3/envs/nuitka/bin/python
OS: Linux
Arch: x86_64

**Update**
After rebasing I'm getting `TypeError: write() argument must be str, not bytes`

bksahu:
fixed #334 

After rebasing I was getting another `TypeError` i.e. `TypeError: write() argument must be str, not bytes`

### PR Checklist

### DIFF
--- a/nuitka/tools/quality/autoformat/Autoformat.py
+++ b/nuitka/tools/quality/autoformat/Autoformat.py
@@ -326,14 +326,10 @@ def autoformat(filename, git_stage, abort):
     if git_stage:
         old_code = getFileHashContent(git_stage["dst_hash"])
     else:
-        old_code = getFileContents(filename)
+        old_code = getFileContents(filename, "rb")
 
-    if not isinstance(old_code, str):
-        with open(tmp_filename, "w") as output_file:
-            output_file.write(old_code.decode("utf-8"))
-    else:
-        with open(tmp_filename, "wb") as output_file:
-            output_file.write(old_code)
+    with open(tmp_filename, "wb") as output_file:
+        output_file.write(old_code)
 
     try:
         _cleanupWindowsNewlines(tmp_filename)
--- a/nuitka/tools/quality/autoformat/Autoformat.py
+++ b/nuitka/tools/quality/autoformat/Autoformat.py
@@ -328,7 +328,7 @@ def autoformat(filename, git_stage, abort):
     else:
         old_code = getFileContents(filename)
 
-    with open(tmp_filename, "w") as output_file:
+    with open(tmp_filename, "wb") as output_file:
         output_file.write(old_code)
 
     try:
--- a/nuitka/tools/quality/autoformat/Autoformat.py
+++ b/nuitka/tools/quality/autoformat/Autoformat.py
@@ -328,8 +328,12 @@ def autoformat(filename, git_stage, abort):
     else:
         old_code = getFileContents(filename)
 
-    with open(tmp_filename, "wb") as output_file:
-        output_file.write(old_code)
+    if not isinstance(old_code, str):
+        with open(tmp_filename, "w") as output_file:
+            output_file.write(old_code.decode("utf-8"))
+    else:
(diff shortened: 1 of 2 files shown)


kayhayen:
This was during a commit, right? Because I think the code is supposed to be a `str`, but `getFileHashContent` produces bytes I think. Just do the `if str is not bytes: ...decode("utf8")` dance there, because that is wrong. Right now `getFileContent` gives str, and `getHashContent` gives bytes, not a good idea.

That said, for non-utf8 code, we are going to have a hard time this way... and maybe, the `old_code` needs to be bytes, and when we add a mode to `getFileContent`, allowing us to say `rb` there, and then make sure we work on bytes all the time.

bksahu:
Yes, it was during a commit. I will try to submit a patch tomorrow morning.

kayhayen:
@bksahu You are compensating in your change for the difference bytes from git, unicode from filesystem, when you write the file. But I think that "old_code" is also later compared, and again that issue will be there.

```
        changed = False
        if old_code != getFileContents(tmp_filename):
            my_print("Updated.")
```

So I am thinking, instead of converting back and forth, make sure to write the file with `wb` and teach `getFileContents` to allow passing the mode, so you can make this and above `getFileContents(tmp_filename, "rb")`.

Always make sure to solve your data problems at the inception. So reading the files as "unicode" has no value in here, as we will have to know the encoding, and limit ourselves, so e.g. no test can have latin1 anymore. To copy, compare, we can work with bytes. Then black, isort, and friends can deal with it when reading from the file.

Do you get me there?

bksahu:
Yes. We want `old_code` to be in a single format i.e `Bytes` and **not** `Unicode` throughout the whole process.

bksahu:
@kayhayen Thank you for creating git hooks. It will save a lot of time. :relaxed:

PR #350, opened 2019-04-11
Created more construct tests

ddatsko:

### What does this PR do?
Add more construct tests for different functions and operations

### Why was it initiated?
One of first tasks of GSoC project I want to take up is about adding more micro-benchmarks, so I decided to write more construct tests.
###  Any relevant Issues?
@kayhayen should there be added more tests for testing comparison between 2 objects or changing object type(as I did with ```int``` and ```srt```) or even this was not needed?

### PR Checklist


kayhayen:
Can you think of a way to do this using Jinja?

ddatsko:
> Can you think of a way to do this using Jinja?

Do you mean writing a general template and just small pieces of code in every file?

kayhayen:
Well, yes, create these construct test cases via a Jinja template. They vary very little, don't they

ddatsko:
@kayhayen, can you review the code? The main idea of it is if the file with test starts with "{% extends" then we create a temporary file and call ```makeTestFromTemplate``` function that writes all the code into that temporary file. I just don't know if the structure of templates is OK... Maybe it would be better to make some variables common for every test and delete ```start``` block? And maybe there is a better way to determine whether the file is just a Python file or template rather than reading it? Also I haven`t a clear idea of what file format should have these templates like added ```OperationIntegerJinjaAdd.py```?
>Note: When I tested performances of ```OperationIntegerAdd.py``` and ```OperationIntegerJinjaAdd.py``` I got the same results

ddatsko:
I can`t find out why all the checks are not passed... When I ran all tests on my computer I got this:
```
+Traceback (most recent call last):
+  File "/home/denys/Nuitka/tests/standalone/CtypesUsing.dist/CtypesUsing.py", line 20, in <module>
+ImportError: No module named ctypes
 
Exit codes 0 (CPython) != 1 (Nuitka)
Error, outputs differed.
Error exit! 1
```

tests/benchmarks/constructs/layout
```diff
@@ -0,0 +1,40 @@
+#     Copyright 2019, Kay Hayen, mailto:kay.hayen@gmail.com
```
kayhayen: Put your name here.

tests/benchmarks/constructs/TypeChangingIntToStr.py
```diff
@@ -0,0 +1,42 @@
+#     Copyright 2019, Kay Hayen, mailto:kay.hayen@gmail.com
```
kayhayen: Question, are these already generated ones?

nuitka/tools/testing/measure_construct_performance/__main__.py
```diff
@@ -76,6 +78,12 @@ def main():
     if os.path.exists(test_case):
         test_case = os.path.abspath(test_case)
 
+    test_case_data = open(test_case).read()
```
kayhayen: That's very good.

tests/benchmarks/constructs/OperationBigIntegerMul.py
```diff
@@ -0,0 +1,43 @@
+#     Copyright 2019, Kay Hayen, mailto:kay.hayen@gmail.com
```
kayhayen: So I guess, you didn't change these ones over to your new mechanism

tests/benchmarks/constructs/OperationIntegerJinjaAdd.py
```diff
@@ -0,0 +1,17 @@
+{% extends "layout" %}
```
kayhayen: I think "layout" is a bad name, can you come up with some more descriptive. How about ``RepeatedFunctionCall.py.j2`` or so. And use the `.j2` suffix, so we can tell Jinja templates, even if there is no official suffix for these kinds of files, but this way, ``autoformat-nuitka-source`` could learn some basic corrections.

kayhayen:
As for your standalone test issue @ddatsko that is very interesting, and should be a issue raised. Please fill out the issue template, as it has to be something about your Python installation.

ddatsko:
@kayhayen , I changed all the tests to Jinja2 templates except 2 of them (I am going to rewrite them too) and changed autoformat to consider files of **.py.j2** extension. I should say that I ran all the old tests and all the new (with Jinja) ones and results of ```NUITKA GAIN```, ```BASE GAIN``` and ```RAW GAIN``` were almost equal (some of them had a difference of about 4 per cent but I discovered that the reason for it was that some of calledRepeatedly() functions returned not None, but some variables, so I didn`t pay attention to it). As for Travis checks, the reason for their failure was abscence of Jinja2 module:
```
nuitka/tools/testing/TestFromTemplateCreator.py:1 E0401 import-error  Unable to import 'jinja2'
```
Can we include Jinja to ```requirements-devel.txt``` or just configure Travis build?

ddatsko:
@kayhayen , I think, this pull request is ready to be merged (I was not able to rewrite 2 construct tests as their structure really differed from others). I ran all the tests again and the NUITKA_GAIN almost did not differ at the ones, created with Jinja2 and those, that were present before. And what should be done to Travis checks? Can You some way configure Travis to skip the file I wrote or add Jinja2 to requirements?

kayhayen:
@ddatsko Can you rebase this to current develop branch, it will fix the Jinja2 issue, also there are PyLint issues.

And I was thinking, and I kind of don't agree with the generated test files not also existing as full files, so I guess, what I would rather do is:

At the start of the test, simply go over the `.py.j2` files and create the many `.py` variant files from it, while having them in `.gitignore`. That way, the files can be looked at more easily, and edited by hand, if we want to, or need to, until we update the template file. Do you understand that?

ddatsko:
@kayhayen ,yeah, I think I got what you are talking about. As for Pylint issues, I'm really sorry that I haven`t fixed them, I will do this. I just don`t have a lot of time because of the end of the semester at university and preparing for exams. I will try to finish this PR as soon as possible

kayhayen:
Unlikely to be finished at this point.

PR #363, opened 2019-04-23
Qt theme for windows

katant:
Recreation of https://github.com/Nuitka/Nuitka/pull/256

### PR Checklist
- [ ] Correct base branch selected? `develop` for new features and bug fixes too. If it's

### DIFF
--- a/nuitka/plugins/standard/PySidePyQtPlugin.py
+++ b/nuitka/plugins/standard/PySidePyQtPlugin.py
@@ -181,6 +181,7 @@ if os.path.exists(guess_path):
             ]
 
             if os.name == "nt":
+                plugin_options.add("styles") # Qt Styles
                 # Those 2 vars will be used later, just saving some resources
                 # by caching the files list
                 qt_bin_dir = os.path.normpath(os.path.join(plugin_dir, "..", "bin"))


kayhayen:
So it seems "all" and "sensible,styles" should already work for you, can you try `--enable-plugin=qt-plugins=sensible,styles` and see if that does it without the code change? That would leave us with answering if including styles would be sensible or not. 

Does your application pick a style (how), or is this about avoiding a fallback to unstyled that is very ugly, if I recall it correctly?

redstoneleo:
I tried the command with [PyQt5WithPIL.py](https://gist.github.com/redstoneleo/fd260e89701912bb22b0533224d041fe)
`nuitka --standalone --mingw64 --remove-output --enable-plugin=qt-plugins=sensible,styles  PyQt5WithPIL.py --show-progress`
It solved the problem ! I think the meaning of `qt-plugins=sensible,styles` should be explained  in the documentation.

kayhayen:
Hello there,

>
> It solved the problem ! I think the meaning of qt-plugins=sensible,styles
> should be explained in the documentation.
>

I think it should just be part of "sensible", which is the default. I still
need to check out how large an effect it is. The `ImportError` and Qt
plugin discovery problem is something I am going to look into later. I have
access to a Windows machine, but too low on time right now.

But thanks on persisting in reporting on this issue, it seems easy now.

Yours,
Kay

>

kayhayen:
To merge this change, I would need to know what kind of impact including styles has one Windows and Linux in terms of distribution size, otherwise this could easily lead to creep in dependencies on KDE, Gnome, what not, and even for styles on Windows, how many are there, which effect does it have, do we want to include only one?

kayhayen:
Closing for lack of feedback.

PR #378, opened 2019-05-02
Tests: Add search mode "All"

### LINKED ISSUE #372: Tests: Add search mode that continues running tests after failure
It would be helpful if the test runner had an option to continue after the first error.  This would reduce the amount of test runs needed to determine which bits are working vs failing.

It can be really dumb, and just emit the error information immediately, and store a global failure status so it reports a non-zero exit status at the end.  I would rather kick off a build with all tests enabled and then come back an hour later and read a large log to identify all failures, rather than constantly restart the tests with different test dependencies present.

Ideally it reports the number of failures at the end, and even better would be reporting the error information at the end also.

I can understand avoiding creating too many features in a custom test runner.  Has there been an attempt to wrap the tests in an existing test runner?  Maybe a shell based test runner, like @bats-core?

bksahu:
Fixes #372 

### PR Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ Changelog.html
 doc/*.1
 doc/man-*.html
 doxygen-warnings.log
+venv_cpython/
--- a/nuitka/tools/testing/Common.py
+++ b/nuitka/tools/testing/Common.py
@@ -349,11 +349,6 @@ def compareWithCPython(dirname, filename, extra_flags, search_mode, needs_2to3):
     _removeCPythonTestSuiteDir()
 
     if result != 0 and result != 2 and search_mode.abortOnFinding(dirname, filename):
-        # if search_mode.isSearchModeAll():
-        #     search_mode.updateTotalErrors()
-        # else:
-        #     message = "Error exit!" + str(result)
-        #     search_mode.exit(message)
         search_mode.onErrorDetected("Error exit! %s" % result)
 
     if converted:
@@ -378,11 +373,7 @@ def checkCompilesNotWithCPython(dirname, filename, search_mode):
         result = 2
 
     if result != 1 and result != 2 and search_mode.abortOnFinding(dirname, filename):
-        if isinstance(search_mode, SearchModeAll):
-            search_mode.updateTotalErrors()
-        else:
-            my_print("Error exit!", result)
-            sys.exit(result)
+        search_mode.onErrorDetected("Error exit! %s" % result)
 
 
 def checkSucceedsWithCPython(filename):
--- a/nuitka/tools/testing/Common.py
+++ b/nuitka/tools/testing/Common.py
@@ -354,8 +354,7 @@ def compareWithCPython(dirname, filename, extra_flags, search_mode, needs_2to3):
         # else:
         #     message = "Error exit!" + str(result)
         #     search_mode.exit(message)
-        message = "Error exit!" + str(result)
-        search_mode.onErrorDetected(message)
+        search_mode.onErrorDetected("Error exit! %s" % result)
 
     if converted:
         os.unlink(path)
--- a/nuitka/tools/testing/Common.py
+++ b/nuitka/tools/testing/Common.py
@@ -15,6 +15,7 @@
 #     See the License for the specific language governing permissions and
 #     limitations under the License.
 #
+
 """ Common test infrastructure functions. To be used by test runners. """
 
 from __future__ import print_function
@@ -42,6 +43,7 @@ from nuitka.utils.FileOperations import (
 )
 
 from .SearchModes import (
+    SearchModeAll,
     SearchModeBase,
     SearchModeByPattern,
     SearchModeCoverage,
@@ -347,8 +349,11 @@ def compareWithCPython(dirname, filename, extra_flags, search_mode, needs_2to3):
     _removeCPythonTestSuiteDir()
 
     if result != 0 and result != 2 and search_mode.abortOnFinding(dirname, filename):
-        my_print("Error exit!", result)
-        sys.exit(result)
+        if isinstance(search_mode, SearchModeAll):
+            search_mode.updateTotalErrors()
+        else:
+            my_print("Error exit!", result)
+            sys.exit(result)
 
     if converted:
         os.unlink(path)
@@ -372,8 +377,11 @@ def checkCompilesNotWithCPython(dirname, filename, search_mode):
         result = 2
 
     if result != 1 and result != 2 and search_mode.abortOnFinding(dirname, filename):
-        my_print("Error exit!", result)
-        sys.exit(result)
+        if isinstance(search_mode, SearchModeAll):
+            search_mode.updateTotalErrors()
+        else:
+            my_print("Error exit!", result)
+            sys.exit(result)
 
 
 def checkSucceedsWithCPython(filename):
@@ -837,6 +845,7 @@ def checkReferenceCount(checked_function, max_rounds=10):
 
 
 def createSearchMode():
+    all_mode = len(sys.argv) > 1 and sys.argv[1] == "all"
     search_mode = len(sys.argv) > 1 and sys.argv[1] == "search"
     resume_mode = len(sys.argv) > 1 and sys.argv[1] == "resume"
     only_mode = len(sys.argv) > 1 and sys.argv[1] == "only"
@@ -844,8 +853,9 @@ def createSearchMode():
     coverage_mode = len(sys.argv) > 1 and sys.argv[1] == "coverage"
 
     if coverage_mode:
-
         return SearchModeCoverage()
+    elif all_mode:
+        return SearchModeAll()
     elif resume_mode:
         return SearchModeResume(sys.modules["__main__"].__file__)
     elif search_mode and start_at:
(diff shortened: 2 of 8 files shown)


tests/optimizations/run_all.py
```diff
@@ -108,12 +108,15 @@ def checkSequence(statements):
             print_arg, = getRole(statement, "value")
 
             if not isConstantExpression(print_arg):
-                sys.exit(
-                    "%s: Error, print of non-constant '%s'." % (
-                        getSourceRef(statement),
-                        getKind(print_arg)
+                if isinstance(search_mode, SearchModeAll):
```
kayhayen: One style complaint here. I think using `isinstance` is nearly always too involved with classes and their hierarchy. Instead, most often, a `isXYZ` or `shallABC` method that is asked for a feature or desired behavior better.

Also, in this concrete example, I dislike the `sys.exit()` being conditional. I think that is precisely what we want the search mode to decide.

Instead call `search_mode.exit(filename, message =...)`, the default implementation will then call `sys.exit(message)` whereas the all mode can pick it up, and add it to the list of errors occurred.

nuitka/tools/testing/SearchModes.py
```diff
@@ -173,3 +174,15 @@ def abortIfExecuted(self):
...
+
+class SearchModeAll(SearchModeBase):
+    def __init__(self):
+        SearchModeBase.__init__(self)
+        self.total_errors = 0
+
+    def updateTotalErrors(self):
+        self.total_errors += 1
+
+    def finish(self):
```
kayhayen: This probably should also `sys.exit()` with error code, so it's noticed by outside world.

kayhayen: See however first my comment below.

nuitka/tools/testing/Common.py
```diff
@@ -341,8 +343,11 @@ def compareWithCPython(dirname, filename, extra_flags, search_mode, needs_2to3):
     _removeCPythonTestSuiteDir()
 
     if result != 0 and result != 2 and search_mode.abortOnFinding(dirname, filename):
-        my_print("Error exit!", result)
-        sys.exit(result)
+        if search_mode.isSearchModeAll():
```
kayhayen: I would rather make this `search_mode.onErrorDetected(message)` and not use a test method.

When I said that test methods of the `isXY` kind are better, that was true. But even better, make the search mode decide what's going to happen entirely, and just inform it about facts.

bksahu: Sorry, I don't understand. Do you want me to store and display errors instead of just incrementing the error counter?

bksahu:
@kayhayen I have found a weird bug. During last commit `c01d046` somehow all the files I have changed has no executable permission due to which I am getting `permission denied` error while trying to run the tests. BTW before committing, I manually ran  `autoformat-nuitka-source` and `check-nuitka-with-pylint3`.

nuitka/tools/testing/Common.py
```diff
@@ -341,8 +343,13 @@ def compareWithCPython(dirname, filename, extra_flags, search_mode, needs_2to3):
...
 
     if result != 0 and result != 2 and search_mode.abortOnFinding(dirname, filename):
-        my_print("Error exit!", result)
-        sys.exit(result)
+        # if search_mode.isSearchModeAll():
+        #     search_mode.updateTotalErrors()
+        # else:
+        #     message = "Error exit!" + str(result)
+        #     search_mode.exit(message)
+        message = "Error exit!" + str(result)
```
kayhayen: Why not use the "%" formatting?

bksahu: Okay

bksahu:
@kayhayen So, should I close this PR?

kayhayen:
I was waiting for you to address my review comments, then merge it. The merge conflicts are probably from black formatting which I try to now apply to test runners as well. However, there can be an # isort:start trick missing in some,beware.

bksahu:
Okay. I think I have already addressed all your reviews but I will look into it again today. Also I am unable to resolve the merge conflict as it says "Only those with write access to this repository can merge pull requests."

bksahu:
@kayhayen I'm ready for a review.

nuitka/tools/testing/Common.py
```diff
@@ -347,8 +349,12 @@ def compareWithCPython(dirname, filename, extra_flags, search_mode, needs_2to3):
     _removeCPythonTestSuiteDir()
 
     if result != 0 and result != 2 and search_mode.abortOnFinding(dirname, filename):
-        my_print("Error exit!", result)
-        sys.exit(result)
+        # if search_mode.isSearchModeAll():
```
kayhayen: You probably meant to remove the commented out codes?

nuitka/tools/testing/Common.py
```diff
@@ -372,8 +378,11 @@ def checkCompilesNotWithCPython(dirname, filename, search_mode):
         result = 2
 
     if result != 1 and result != 2 and search_mode.abortOnFinding(dirname, filename):
-        my_print("Error exit!", result)
-        sys.exit(result)
+        if isinstance(search_mode, SearchModeAll):
```
kayhayen: Not using `onErrorDetected` here, but probably should too.

nuitka/tools/testing/SearchModes.py
```diff
@@ -82,10 +82,24 @@ def _match(cls, dirname, filename, candidate):
...
+    def isSearchModeAll(self):
+        # Virtual method, pylint: disable=no-self-use
+        return False
+
     def isCoverage(self):
         # Virtual method, pylint: disable=no-self-use
         return False
 
+    def onErrorDetected(self, message):
+        if self.isSearchModeAll():
```
kayhayen: Please use overloading of methods. This should simply be the `sys.exit` I figure, while the method in the all search mode would do what it does. You do not need `isSearchModeAll` in this case, which was the point of using a method.

tests/optimizations/run_all.py
```diff
@@ -33,12 +36,8 @@
...
     os.path.normpath(
-        os.path.join(
-            os.path.dirname(os.path.abspath(__file__)),
-            "..",
-            ".."
-        )
-    )
+        os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "..")
+    ),
 )
```
kayhayen: You are a victim here of the commit hook now covering `run_all.py`, but there are no isort:start to make sure the `nuitka` package can be imported. Please add `# isort:start` in much the same way I did in some other runners (can search for it), and it is my fault. I ought to have added these immediately to of all them.

tests/test-runners/run_all.py
```diff
@@ -39,9 +41,11 @@
 
 from nuitka.tools.testing.Common import getTempDir, my_print, setup
 from nuitka.utils.FileOperations import copyTree, listDir, removeDirectory
-
+from nuitka.tools.testing.SearchModes import SearchModeAll
 
 def main():
+    search_mode = createSearchMode()
```
kayhayen: This one is not using search modes at all yet, wrong to add it without a scan.

bksahu:
@kayhayen Should we have a testing module to test the "search modes"?

kayhayen:
No, I don't think that is needed. Us looking at them running will have to do. :)

.gitignore
```diff
@@ -33,3 +33,4 @@ Changelog.html
 doc/*.1
 doc/man-*.html
 doxygen-warnings.log
+venv_cpython/
```
kayhayen: The lack of trailing new line (not a problem), just curious, is that something that the commit hook doesn't do? Because I think it should.

tests/packages/run_all.py
```diff
@@ -19,51 +19,48 @@
...
+from nuitka.tools.testing.SearchModes import SearchModeAll
+
+# Find nuitka package relative to us.
+sys.path.insert(
+    0,
+    os.path.normpath(
+        os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "..")
+    ),
+)
+
```
kayhayen: I noticed you did add the isort:start correctly for one runner, but basically all that to that sys.path insertion need to do it, sorry, that is a bunch.

kayhayen:
Looks good aside those missing `# isort:start` markers that allow the isort to handle the sys.path extension to allow finding local `nuitka` from checkout gracefully. And of course, there is also still the method `isSearchModeAll` or so, which you stopped using, but has no right to exist. 

Thanks for the good work. I am almost tempted to repair the `isort:start` myself.

bksahu:
Give me 10 minutes.

.gitignore
bksahu: I think it doesn't

kayhayen: I prepared an update on factory, which makes it handle .gitignore files too.

tests/test-runners/run_all.py
```diff
@@ -22,6 +22,8 @@
 import os
 import sys
 
+from nuitka.tools.testing.SearchModes import SearchModeAll
```
kayhayen: Here is is actually harmful, and as unused ought to have been removed. The pylint doesn't cover these files yet. I am going to fix this myself though, no big deal,as Travis doesn't run these. Maybe it ought to though, I will explore that. And also adding the test runners to pylint coverage.

PR #399, opened 2019-05-22
Fix NumpyPlugin binaries path on case-sensitive file systems

johndoe46:
Numpy plugin breaks the path of binaries in case-sensitive file systems, giving a FileNotFoundError with option --plugin-enable=numpy while actually the .so file is really available.

Example:
Having numpy file at "/home/johndo/SOMEPROJECT/venv/lib/python3.7/site-packages/numpy/.libs/libopenblasp-r0-8dca6697.3.0.dev.so"

Gives such errror with nuitka:

FileNotFoundError: [Errno 2] No such file or directory: '/home/johndo/someproject/venv/lib/python3.7/site-packages/numpy/.libs/libopenblasp-r0-8dca6697.3.0.dev.so'

This patch fixes the issue so that only the base file name is lower-cased, leaving the rest of the path as-is.

### PR Checklist

### DIFF
--- a/nuitka/plugins/standard/NumpyPlugin.py
+++ b/nuitka/plugins/standard/NumpyPlugin.py
@@ -219,7 +219,7 @@ class NumpyPlugin(UserPluginBase):
             info("")
             info(" Copying extra binaries from 'numpy' installation:")
             for f in binaries:
-                bin_file = f[0].lower()  # full binary file name
+                bin_file = os.path.dirname(f[0]) + os.path.sep + os.path.basename(f[0]).lower()  # full binary file name
                 idx = bin_file.find("numpy")  # this will always work (idx > 0)
                 back_end = bin_file[idx:]  # => like 'numpy/core/file.so'
                 tar_file = os.path.join(dist_dir, back_end)


johndoe46:
Travis checks failed but I'm not sure what I can do about it... seems to be an issue with the CI system itself.

JorjMcKie:
I also saw this error and have it fixed in the `jorj` branch. So it should be on the way to develop when @kayhayen moves stuff to other branches.

JorjMcKie:
I am also getting those pylint errors since yesterday ...

johndoe46:
CI gives this:

2.6 is not installed; attempting download
Downloading archive: https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/16.04/x86_64/python-2.6.tar.bz2
0.12s$ curl -sSf -o python-2.6.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 404 Not Found

Same for python 3.3... so deprecated python versions for Travis maybe ? I'm not too versed in CI.

kayhayen:
I have seen this occur at times, but it repairs itself after a while, develop branch worked with 2.6 roughly 4 hours ago.

johndoe46:
Hello,

Should I re-submit a PR ?

nuitka/plugins/standard/NumpyPlugin.py
```diff
@@ -219,7 +219,7 @@ def considerExtraDlls(self, dist_dir, module):
             info("")
             info(" Copying extra binaries from 'numpy' installation:")
             for f in binaries:
-                bin_file = f[0].lower()  # full binary file name
+                bin_file = os.path.dirname(f[0]) + os.path.sep + os.path.basename(f[0]).lower()  # full binary file name
```
kayhayen: I think using `os.path.sep` and `+` seems not a good idea, but `os.path.join` is always preferable.

But I am wondering, why `lower` is used at all. Is this about filenames in the target to look nicer? I would say that `os.path.normcase` is a much better choice then.

kayhayen:
In my mind, we should probably have something that does the copying, and we shouldn't make plugins to the actual file copy.ing at all, so we can have a more common standard.

I think there was something about data files, where I adapted things, so now they only say what to do, and then it's done in `nuitka.freezer.Standalone.copyDataFiles` and that's cleaner I think. That then can use a best method to copy/create files, and can be enhanced universally.

What I think then it needed, for plugin developers, is the make transparent, what file is being copying from where, and why. The why could be the plugin that ordered it, and an optional string it produces.

For now I think, `normcase` should be employed here, and we make structural changes like this in a v2 refactoring. What is your take on this @JorjMcKie ?

JorjMcKie:
Sorry to say it is my fault that we are discussing this.
Indeed my motivation was to get a nicer printed output - and as usual I forgot Linux has this bloody case sensitivity in the file names (I heard that even faithful Linux disciples regret this design decision ...).

Anyway: mea culpa.

@kayhayen  - yes I will look into this option.
First impression: it may not be as easy as it seems, because already `tk-inter`, `numpy`, `scipy`, `matplotlib`, `torch`, `sklearn` look quite different there - beyond the obvious difference of the file / folder names.

JorjMcKie:
`normcase` is a sure thing - I should have thought of it in the first place.
I also could have sworn I had eliminated this bug in my recent `jorj` branch content ...

JorjMcKie:
Another electricity collapse, so some mor comments via smartphone.

The copying function itself (ie. shutil) should indeed never be part of the plugin.
There still some details need be clarified beyond this rule: in tk-inter I am copying 2 folders, but afterwards I am deleting one of the already copied sub folders again ... don’t yet know how to resolve things like this.

Von meinem iPhone gesendet

kayhayen:
@JorjMcKie I think I did the same for PyQt plugin, copy first, then deleted unwanted stuff. Very bad style, but easier, but slow disks hate you for that.

My proposal for this would be that for folders, we sort of yield source and destination paths, and a decision function, that will be called back to decide if a child folder should be entered, and if a contained file should be copied.

```
def decide(path):
   """ Return true or false for a path, maybe False, 'Some explanation to the user should be dig deeper'. """

yield source_path, dest_path, decide
``` 

I think for generated data files, I already do that kind of think as well.


I will look into the Jorj branch again. I think I merged content from one point, and might have missed updates since. Are you talking about main repo one, or your fork on GH still?

kayhayen:
@johndoe46 Want to go `os.path.normcase` in this one. I think that's all it takes, and I would glady hit merge, now that @JorjMcKie agrees. The restructuring is not blocking this at all. Just food for thought, that the best possible copy is going to do that. Or simply strike `lower()` that will also do.

kayhayen:
Stupid me, sorry for re-opening noise.

PR #403, opened 2019-05-23
support hinted-mode plugins extending NuitkaPluginBase

drzraf:
`Usr_Plugin` extends `NuitkaPluginBase` instead of `UserPluginBase`
See https://github.com/Nuitka/NUITKA-Utilities/commit/29d5fd2ea20af6bbe159029e51a925eda56e51a6#r33654001


kayhayen:
I don't know how, but it seems this changed made it into Nuitka already.

PR #407, opened 2019-05-26
Optimize the "all" built-in

bksahu:
Relevant Issue: #232 

# Todo
- [x] Run all world test without the optimization (done already)
- [x] Run all world test with only the C code replaced, basically using your C code only, and see what this finds. This is the point in time, where this has the best exposure. The node would return self, None, None and be good, i.e. definitely not do anything wrong.
- [x] Run all world test with the generic implementation (the one you got right now in your PR), and see what it does (any it finds bugs).
- [x] Look at cases to specialize and do that, again run all world test with that.


kayhayen:
As a next step, I want you to think about these:

`any([0]*10000)`
`all([0]*10000)`
`sum([0]*10000)`

What's notable about those, what will be need from the sequence repeat (that doesn't yet exist in that form, but will). Is there anything, that needs to be done 10000 times? Or is there a way to shortcut, and who should have that responsibility, and where to put it. For `any` and `all` alone, there is one solution in terms of optimization that could be employed, but doesn't fit `sum`, or at least so I am thinking. Maybe it does too.

bksahu:
I'm thinking of specialising that generic algorithm by adding it to `ConstantRefNodes`. But that won't be helpful in case of `sum`. However for sum we can do one thing that is to do same computation that we will do in case of `any` or `all`  but return 0 in place of `False`. We need to pass modes (eg. any, all or sum) as arguments to our method.

kayhayen:
I was thinking (but don't change things right now that way!), we might (later) move more of these kinds of iteration centric algorithms in a base class method of the iteration handles. Like you would do more of the `any` computation with default implementation there. And another one for `all` and `sum`.

And then, you overload to `all` implement shortcuts, e.g. `all(range(1,100000))` doesn't really need to use the generic algorithm that looks at elements, you just watch out for `0` to be in a range or not (or `xrange` for that matter). With `all` and ranges, stopping to check after 256 elements, can quickly become a reason, you do not get possible optimization.

And for `*` as a sequence multiplication, you would special handle `0` anyway, so ignoring that for now, we should make that node, one it exists (and I think it will), make that optimize to an empty sequence of the type.

But otherwise, say we only know it's non-zero, we might be able to optimize this for as long as we know that the truth check doesn't have a side effect: `any(some_list * 1000)` -> `any(some_list)` and `sum(some_list * 1000)` -> sum(some_list) * 1000` so those iteration handles, for some things, they can be more clever.

Then we have a spot, where the node or its iteration handle (they are friends anyway), do this, and it's much like what you were doing as `computeExpressionAny`, but not in the node, but in the iteration handle. When we moved the whole algorithm there, we can make it so, that iteration handles can plug and overload them.

If we pass this as a mode, or a separate method name, I think is a matter of how common you think the code is going to be. For `all` and `any`, things are going to be relatively similar, but not that easily similar without a lot of tests over a mode. Also I think it's not a lot of code you are repeating.

Also consider, @bksahu the any algorithm is as it stands, already relatively hard to grasp. Took us both a while to get where it is now. Making a function that does either `any` or `all` depending on a mode is not making it easier. And `sum` might make it catastrophic. So I am arguing here for those methods to be the not same, but different names, and different algorithm, even if `any` and `all` tend to have commonality, and `sum` also does looping, I think it doesn't justify that.

kayhayen:
But to me, and that is why I got convinced by your slot idea ultimately, the node for `*` and `+` do know how they can help to optimize `sum`, and should be tasked with it. And their changes are often algorithmic.

So e.g. `sum(a+b)` -> `sum(a) + sum(b)` is normally a good optimization for as long as `a+b` was understood well enough by the `+` node.

I think we have a lot of signs that we found a very good design there, @bksahu and I suggest we focus on finishing the iteration handle work in a V1.

Then we review what it is, and identify weakness, and things that would have made it easier, apply that, and then continue with `all` and a refined code. I have very good experience with that kind of approach.

Right now, I think, the iteration handles have one problem, that I also do not want us to deal with immediately. There is sequential iteration and random access, and we shoved it into one thing. Some of your constructors are suffering from increased complexity. And we get the problem of expression that we can do only one of those, but not the other.

I guess, we should have `SequentialIterationHandleBase` and `RandomAccessIterationBase`, and then e.g. a helper that turns one into the other. But I think we should discuss this after we finish the `any` work.

bksahu:
So, I have added the rest of the code along with the algorithm for `all` built-in. However, for now, I have disabled it until `any` node PR is merged as `IterationHandles` class is in there.

nuitka/nodes/BuiltinAllNodes.py
```diff
@@ -0,0 +1,112 @@
...
+        shape = value.getTypeShape()
+        if shape.hasShapeSlotIter() is False:
+            return makeRaiseTypeErrorExceptionReplacementFromTemplateAndValue(
+                template="'%s' object is not iterable",
+                operation="all",
+                original_node=value,
+                value_node=value,
+            )
+
+        if shape is ShapeTypeStr:
```
kayhayen: What about `bytes`, other types, consider the typical builtin shapes.

nuitka/nodes/BuiltinAllNodes.py
```diff
@@ -0,0 +1,112 @@
...
+            return (
+                wrapExpressionWithNodeSideEffects(
+                    new_node=makeConstantReplacementNode(constant=True, node=self),
+                    old_node=value,
+                ),
+                "new_constant",
+                "Predicted truth value of built-in all string type argument",
```
kayhayen: Always quote things, here e.g. 'all', so it's clear you don't mean "all arguments" or something like that.

nuitka/nodes/BuiltinAllNodes.py
```diff
@@ -0,0 +1,112 @@
...
+            if all_true is not None:
+                result = wrapExpressionWithNodeSideEffects(
+                    new_node=makeConstantReplacementNode(constant=all_true, node=self),
+                    old_node=value,
+                )
+
+                return (
+                    result,
+                    "new_constant",
+                    "Predicted truth value of built-in all argument",
```
kayhayen: See above.

nuitka/nodes/IterationHandles.py
```diff
@@ -49,6 +49,37 @@ def getNextValueTruth(self):
             return StopIteration
         return iteration_value.getTruthValue()
 
+    def getAllElementTruthValue(self):
+        all_true = True
+        count = 0
+        while True:
+            # This is a function that returns True, False, or None
```
kayhayen: I hope the doc string is saying this too. :-)

tests/basics/BuiltinsTest.py
```diff
@@ -565,4 +565,18 @@ def S():
...
+print(any({1:"One", 2:"Two"}))
+
+# Test for all builtins
+print(all([None, None, None]))
+print(all([None, 4, None]))
+print(all([]))
+print(all([0] * 20000))
+print(all([0] * 255))
+print("All with ranges:")
+print(range(1, 10000))
```
kayhayen: Not using `all` here, forgot to call it :-)

tests/basics/BuiltinsTest.py
```diff
@@ -565,4 +565,18 @@ def S():
...
+print(range(10000))
+print(range(2, 999, 4))
+print("All with strings:")
+print("")
```
kayhayen: No `all` call. Also other types that are always true too.

tests/basics/BuiltinsTest.py
```diff
@@ -565,4 +565,11 @@ def S():
...
-print(any({1:"One", 2:"Two"}))
\ No newline at end of file
+print(any({1:"One", 2:"Two"}))
+# Test for all builtins
+print(all([None, None, None]))
+print(all([None, 4, None]))
+print(all([]))
+print(all(range(20)))
+print(all([0] * 20000))
+print(all([0] * 255))
```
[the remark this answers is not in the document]
bksahu: I have moved both `any` and `all` tests to a `tests/optimization/CommonOptimizations.py`. I am planning to use it to write tests for future optimizations.

PR #419, opened 2019-06-06
Optimize the "abs" built-in

bksahu:
Relevant Issue: #232


nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -553,6 +553,17 @@ PyObject *BUILTIN_LEN(PyObject *value) {
     return PyInt_FromSsize_t(res);
 }
 
+PyObject *BUILTIN_ABS(PyObject *o) {
+    CHECK_OBJECT(o)
+
+    PyNumberMethods *m = o->ob_type->tp_as_number;
+    if (likely(m && m->nb_absolute))
```
kayhayen: Always use blocks, even for single statements. In C this avoids so many errors.

nuitka/codegen/CodeGeneration.py
```diff
@@ -548,6 +549,7 @@ def makeGlobalContext():
         "EXPRESSION_BUILTIN_REF": generateBuiltinRefCode,
         "EXPRESSION_BUILTIN_EXCEPTION_REF": generateExceptionRefCode,
         "EXPRESSION_BUILTIN_ANONYMOUS_REF": generateBuiltinAnonymousRefCode,
+        "EXPRESSION_BUILTIN_ABS_REF": generateBuiltinAbsCode,
```
kayhayen: There should be no `REF`. That part of the name stands for reference, and makes we are having some kind of name lookup. This is not the case here. That probably means you also misnamed the operation class.

nuitka/nodes/BuiltinAbsNodes.py
```diff
@@ -0,0 +1,41 @@
...
+#
+""" Built-in abs nodes. """
+
+from nuitka.specs import BuiltinParameterSpecs
+
+from .ExpressionBases import ExpressionBuiltinSingleArgBase
+
+
+class ExpressionBuiltinAbs(ExpressionBuiltinSingleArgBase):
+    kind = "EXPRESSION_BUILTIN_ABS"
```
kayhayen: I would have expected this one in `OperatorNodes`. There is going to be a lot of commonality with other unary operations. `|a|` in math is an unary operation, just not one for which there is Python syntax, i.e. the built-in is the only way to say it. That doesn't mean it shouldn't be where the others are.

nuitka/nodes/shapes/BuiltinTypeShapes.py
```diff
@@ -123,6 +123,10 @@ def getTypeName():
     def hasShapeSlotBool():
         return False
 
+    @staticmethod
```
kayhayen: Unfortunately Github diff sucks here and doesn't show the classes, so I am not checking these at this time. Arguably, we need to create test cases for Nuitka that give coverage for these. I have made mistakes in this myself in released versions in the past.

kayhayen: BTW: I node I can enable side by side diff, I am not doing now though.

tests/basics/BuiltinsTest.py
```diff
@@ -524,3 +524,9 @@ def __complex__(self):
     print("Oops, must not happen.")
 else:
     print("Oops, must not happen.")
+
+
+print("Tests for abs():")
+print(abs(-1000000**10))
+print(abs(len([1, 2, 3])))
```
kayhayen: That `len()` will statically be optimized to `3` and make the test case relatively boring.

You would need to cover at least one exception case, static and one that Nuitka doesn't see through.

nuitka/nodes/BuiltinAbsNodes.py
```diff
@@ -0,0 +1,41 @@
...
+
+    builtin_spec = BuiltinParameterSpecs.builtin_abs_spec
+
+    def computeExpression(self, trace_collection):
+        return self.getValue().computeExpressionAbs(
+            abs_node=self, trace_collection=trace_collection
+        )
+
+    def mayRaiseException(self, exception_type):
+        value = self.getValue()
```
kayhayen: What does `abs(float("nan"))` do?

kayhayen: Also, if a class overloaded `__abs__` then it could raise in the slot.

I think you need a function like `mayRaiseExceptionAbs()` in expression nodes, that is pessimistic, for all but a few things. But now that I think of it, `TypeShapeFloatDerived` is based on slots and may work with this approach. 

This may need more though from my side. I think I have had `mayRaiseExceptionBool` for some time, and only recently added the slot to type shapes. These might tie together like that or be redundant, or the base class implementation ought to use the type shape, but other parts may not.

nuitka/codegen/CodeGeneration.py
kayhayen: Ah, I see you did not match the names in the `abs` node and here, if they do not, this will simply not have an effect, and compiling to C should give an error for everything not statically optimized.

nuitka/nodes/shapes/BuiltinTypeShapes.py
bksahu: In a nutshell, I have set `True` only for `int`, `float` and `complex` related shape types.

nuitka/nodes/BuiltinAbsNodes.py
bksahu: > What does `abs(float("nan"))` do?

It returns `nan`. Should I do something to deal with it?

tests/basics/BuiltinsTest.py
bksahu: I have added the exception case. I'm not sure about static and one that Nuitka doesn't see through cases.

bksahu:
I think the pylint error in `FileOperations.py` are in develop also.

bksahu:
@kayhayen the pylint error in travis is unrelated to this PR

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -561,7 +561,11 @@ PyObject *BUILTIN_ABS(PyObject *o) {
         return m->nb_absolute(o);
     }
 
-    return PyErr_Format(PyExc_TypeError, "bad operand type for abs(): '%.200s'", PyString_AsString(o));
+    #if PYTHON_VERSION >= 300
+        return PyErr_Format(PyExc_TypeError, "bad operand type for abs(): '%.200s'", Py_TYPE(o));
+    #else
+        return PyErr_Format(PyExc_TypeError, "bad operand type for abs(): '%.200s'", PyString_AsString(o));
```
kayhayen: I don't belive that, Python never gives object `str` for exception message. Is this from the CPython original code? I would for these always check them first.

kayhayen:
yes, for the pylint stuff, rebase to current develop, I messed up, there is a false alarm of pylint only on travis, that sneaked past me.

nuitka/build/static_src/CompiledCodeHelpers.c
bksahu: Originally in the CPython implementation it was `return type_error("bad operand type for abs(): '%.200s'", o)`

kayhayen: I wouldn't be surprised of `type_error` did do the `Py_TYPE`  on `o`, does it? Since it's a frequent pattern, they probably created a function that does it.

bksahu: Yes, You are right. And I have checked for `CPython27` and there also they are doing `Py_Type` on `o`. But when I compile with `Py27`, I get the following warning:

```
abs_test.build/CompiledCodeHelpers.c: In function ‘BUILTIN_ABS’:
abs_test.build/CompiledCodeHelpers.c:565:46: warning: format ‘%s’ expects argument of type ‘char *’, but argument 3 has type ‘struct _typeobject *’ [-Wformat=]
         return PyErr_Format(PyExc_TypeError, "bad operand type for abs():
```

kayhayen: I am nearly sure t also looks up the `->tp_name` of that `PyTypeObject *`, which is indeed a string, and what is normally output in error messages.

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -590,6 +590,21 @@ PyObject *BUILTIN_ANY(PyObject *value) {
...
+PyObject *BUILTIN_ABS(PyObject *o) {
+    CHECK_OBJECT(o)
+
+    PyNumberMethods *m = o->ob_type->tp_as_number;
+    if (likely(m && m->nb_absolute)) {
+        return m->nb_absolute(o);
+    }
+
+    #if PYTHON_VERSION >= 300
+        return PyErr_Format(PyExc_TypeError, "bad operand type for abs(): '%.200s'", Py_TYPE(o));
```
kayhayen: Just use '%s'.

kayhayen: Should be more like `Py_TYPE(o)->tp_name` which is a C string.

kayhayen:
Let me know once nightshift turned green on this latest pushes.

PR #427, opened 2019-06-12
Extend Cryptodome(x), Add coincurve support

hclivess:
Added support for coincurve
Included libsecp256k1.dll data file required for coincurve to run
Added support for SHA384, SHA512 in Cryptodome(x)

I tested these changes locally, I am not completely sure if the ImplicitImports part is required

### DIFF
--- a/nuitka/plugins/standard/DataFileCollectorPlugin.py
+++ b/nuitka/plugins/standard/DataFileCollectorPlugin.py
@@ -35,6 +35,7 @@ known_data_files = {
     "scrapy": ((None, "VERSION"),),
     "requests": (("certifi", "../certifi/cacert.pem"),),
     "importlib_resources": ((None, "version.txt"),),
+    "coincurve": ((None, "./libsecp256k1.dll"),),
     "boto": ((None, "./endpoints.json"),),
     "moto": (
         (None, "./ec2/resources/instance_types.json"),
--- a/nuitka/plugins/standard/ImplicitImports.py
+++ b/nuitka/plugins/standard/ImplicitImports.py
@@ -506,6 +506,10 @@ class NuitkaPluginPopularImplicitImports(NuitkaPluginBase):
                 yield crypto_module_name + ".Hash._SHA224", True
             elif full_name == crypto_module_name + ".Hash.SHA256":
                 yield crypto_module_name + ".Hash._SHA256", True
+            elif full_name == crypto_module_name + ".Hash.SHA384":
+                yield crypto_module_name + ".Hash._SHA384", True
+            elif full_name == crypto_module_name + ".Hash.SHA512":
+                yield crypto_module_name + ".Hash._SHA512", True
             elif full_name == crypto_module_name + ".Hash.MD5":
                 yield crypto_module_name + ".Hash._MD5", True
             elif full_name == crypto_module_name + ".Protocol.KDF":
@@ -518,6 +522,10 @@ class NuitkaPluginPopularImplicitImports(NuitkaPluginBase):
             yield "pycparser.lextab", True
         elif full_name == "passlib.hash":
             yield "passlib.handlers.sha2_crypt", True
+
+        #Support for coincurve
+        elif full_name == "coincurve":
+            yield "coincurve._libsecp256k1._dll", True
 
     # We don't care about line length here, pylint: disable=line-too-long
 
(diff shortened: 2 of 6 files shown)


nuitka/plugins/standard/DataFileCollectorPlugin.py
```diff
@@ -35,6 +35,7 @@
     "scrapy": ((None, "VERSION"),),
     "requests": (("certifi", "../certifi/cacert.pem"),),
     "importlib_resources": ((None, "version.txt"),),
+    "coincurve": ((None, "./libsecp256k1.dll"),),
```
kayhayen: When a module needs a DLL, that should only be the case, when it is not linked against it, otherwise Nuitka will normally detect static DLL usage.

If this is a case of a dynamically opened DLL you ought to add it to the short handling in `considerExtraDlls` in the same file.

nuitka/plugins/standard/ImplicitImports.py
```diff
@@ -722,4 +730,4 @@ def decideCompilation(self, module_name, source_ref):
             if module_name == unworthy_namespace or module_name.startswith(
                 unworthy_namespace + "."
             ):
-                return "bytecode"
```
kayhayen: Stray new line change at end of file (that the commit hook is not yet dealing with), please remove it while you change the file again.

nuitka/plugins/standard/DataFileCollectorPlugin.py
hclivess: Thank you, I will take a look at this "considerExtraDlls". Sadly, it is not bundled automatically.

nuitka/plugins/standard/ImplicitImports.py
hclivess: Hi, looks like this was automatically done by PyCharm, I'll take a look.

nuitka/plugins/standard/DataFileCollectorPlugin.py
hclivess: I tested this, but I have no idea what is the correct format, any hints?

```
        #Support for coincurve
        if full_name == "coincurve":
            coincurve_dll_path = locateDLL("libsecp256k1.dll")
            dist_dll_path = os.path.join(dist_dir, os.path.basename(coincurve_dll_path))
            return ((coincurve_dll_path, dist_dll_path, None),)

        return ()
```

kayhayen:
Just a quick node, locateDLL wants the name of a DLL, not the filename.

So `libsecp256k1.dll` is definitely wrong in adding that `.dll` suffix, and it likely makes it fail. Also the `lib` prefix is suspicious, but could be true, but very unusual, typically it's used on Linux systems, so either `libsecp256k1` or `secp256k1` ought to be better choices.

Also, it can be that locateDLL is not looking in the package directory, which would be something that needs a fix, if it's hidden there.

hclivess:
Thank you very much, I will try compiling with your suggestions and see if it's included

hclivess:
I tested both `libsecp256k1` and `secp256k1` and neither were detected. Maybe I should ask coincurve to change something in their repo? This is where it happens: https://github.com/ofek/coincurve/tree/master/coincurve

I think they are actually downloading the library from an external source, so I can do that instead:

https://github.com/ofek/coincurve/blob/master/setup.py#L36

kayhayen:
No please don't duplicate anything, just make a working installation work. I think I can take it from there, and extend locateDLL to also search in a package directory, which is what will be needed. So far it has only searched in system paths, but in this case, that will be needed.

You will have to wait a bit though, and make sure I have proper instructions to set up the test virtualenv. I am still away from Windows machine for a week. Then this should be very easy for me once I find the time.

Yours,
Kay

hclivess:
I will use my "hacked" version for now, so no hurry. Thank you very much for help with this. I hope this helps improve Nuitka and perhaps attracts more users.

ofek:
@kayhayen Hello there! Any update on this?

nuitka/plugins/standard/DataFileCollectorPlugin.py
kayhayen: I missed this unfortunately. It looks wrong to provide the ".dll" suffix, that will make it non-portable, also it seems you might want to check for the `coincurve._libsecp256k1._dll` module, for enhanced clarity. Otherwise it looks good to me. Can you rebase the PR? if that's too painful, I would do it, I probably caused the troubles with my rebases of develop. Doing `git rebase upstream/develop` and skipping conflicting commits except your own ones will work though.

kayhayen:
Closing for lack of progress.

PR #430, opened 2019-06-13
Add support for SHA384, SHA512

hclivess:
I'am supplying this change individually instead. For some reason, GitHub still thinks I removed and added one extra line ( `yield crypto_module_name + ".Hash._SHA256", True`)


kayhayen:
The whitespace warnings could be avoided if you install the pre-commit hook, it does the automatic formatting, so Travis would like it. Merging it anyway and manually fixing it up now.

PR #440, opened 2019-06-23
Automating PyPI Wheel Pytest

tommyli3318:
### What does this PR do?
Creates a script automating PyPI Wheel tests that can be extended to test multiple PyPI packages.
Normal pytest results of each package will be compared to its nuitka-built .whl counterpart.
Testing is done to ensure that nuitka is building the wheel correctly for each PyPI package. If the pytests pass/fail in the same way, that means Nuitka built the wheel properly. Else if the tests differ, then something is wrong. Virtualenv is used to create a clean environment with no outside pollution.

GSoC2019

### PR Checklist


tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,123 @@
...
+        # get normal pytest results
+        with withVirtualenv("venv_%s" % package_name) as venv:
+            venv.runCommand(
+                commands=[
+                "call Scripts/activate",
+                "git clone https://github.com/%s/%s.git" % (package_name,package_name),
+                "python -m pip install pytest",
+                "cd %s" % package_name,
+                "python -m pip install -r dev-requirements.txt",
+                # "python -m pytest --disable-warnings",
```
tommyli3318: `python -m pytest --disable-warnings` (currently commented out) runs the normal pytest correctly, but the output is directed to the shell and I don't know of a way to redirect it

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,123 @@
...
+            # need to use subprocess to save stdout
+            # currently doesn't work due to import errors
+
+            os.chdir(os.path.join(base_dir, "venv_%s" % package_name, package_name))
+            process = subprocess.Popen(
+                    args="python -m pytest --disable-warnings",
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                )
+            stdout, stderr = process.communicate()
```
tommyli3318: Using subprocess I can redirect stdout and save it as a string, this is also what `distutils/run_all.py` does. However, pytest gets import errors when ran with subprocesss

tommyli3318:
@kayhayen this is far from being done but I added comments to show you what I meant by "`subprocess` being buggy with pytest when used inside the `withVirtualenv` context manager"

tests/PyPI-pytest/run_all.py
tommyli3318: and I need to save it as a string in some way to be able to compare it to the output of the nuitka wheel pytest

tests/PyPI-pytest/run_all.py
tommyli3318: Is it maybe because the virtualenv isn't activated here?

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,178 @@
...
+
+    exit_code_stderr = compareOutput(
+        "stderr",
+        stderr,
+        w_stderr,
+        ignore_warnings=True,
+        ignore_infos=True,
+        syntax_errors=True,
+    )
+    '''
```
tommyli3318: I think I need some pointers on how to use `compareOutput` before I can move on. It compares cpython output to nuitka output but in my case I'm trying to compare the regular pytest with the nuitka wheel pytest?

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,178 @@
...
+    except Exception as exceptObj:
+        print(exceptObj)
+        continue
+
+
+    os.chdir(base_dir)
+    # remove venv directory
+    # currently not removing the directories?
+    # removeDirectory(os.path.join(base_dir, "venv_%s" % package_name), ignore_errors=False)
+    # removeDirectory(os.path.join(base_dir, "wheel_venv_%s" % package_name), ignore_errors=False)
```
tommyli3318: `removeDirectory` is failing to remove the directories, giving a `PermissionError`, my guess is it has to do with the virtualenvs?

tommyli3318:
@kayhayen Just updated with a new commit, the code now gets stdout and stderr from both the uncompiled pytest and wheel pytest. I left some comments, please review

tests/PyPI-pytest/run_all.py
kayhayen: The `PermissionError` happens a lot on Windows, after you ran a binary.

You cannot delete a binary that has been executed, because you do not know when it really exits, and what happens a lot too, is that Virus checkers work too.

I have excluded my Buildbot working stage from `Windows Defender` and that makes it even faster. The issue is transient, is it?

tests/PyPI-pytest/run_all.py
kayhayen: Yes, I think you are, but py.test output also contains stack traces, and these will often contain Python code, or object names, that you will benefit from being made comparable.

I suggest we retry having a conference call. I will ping you about it.

tests/PyPI-pytest/run_all.py
tommyli3318: I see, is there a way around it? Right now I am manually deleting the virtualenv directories after each run but this deleting should be automated.

nuitka/tools/testing/Virtualenv.py
```diff
@@ -37,6 +37,7 @@ class Virtualenv(object):
     def __init__(self, env_dir):
         self.env_dir = os.path.abspath(env_dir)
 
+    """
```
tommyli3318: I've commented out the entire `runCommand` for testing purposes, but everything still runs correctly for some reason...

nuitka/tools/testing/Virtualenv.py
```diff
@@ -37,6 +37,7 @@ class Virtualenv(object):
     def __init__(self, env_dir):
         self.env_dir = os.path.abspath(env_dir)
```
tommyli3318: I've commented out the entire `runCommand` for testing purposes, but everything still runs correctly for some reason...

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,211 @@
...
+                    "cd ../%s" % package_name,
+                    "python -m pip install -r dev-requirements.txt",
+                    "python setup.py bdist_wheel",
+                ]
+            )
+            
+
+            dist_dir = os.path.join(venv.getVirtualenvDir(), package_name, "dist")
+
+            uncompiled_stdout, uncompiled_stderr = venv.runCommand(
```
tommyli3318: `runCommand` is called from here. I tried to change up runCommand to use subprocess and have it return stdout,stderr but my changes seemed to have no effect. I then commented out the entire `runCommand` method and for some reason, this file still ran correctly?

tommyli3318:
@kayhayen the bdist_wheel part works properly now. However, I ran into something weird when changing up the `runCommand` method to use subprocess and return stderr,stdout. Refer to the comments above. Thank you

tests/PyPI-pytest/run_all.py
tommyli3318: From what I can see, the `withVirtualenv` context manager yields an object of the `Virtualenv` class, and the `runCommand` method is being called here. So how come my changes don't seem to take effect?

tommyli3318:
@kayhayen Updated again, made changes to the Virtualenv class to include a new method `runCommandWithOutput` that uses subprocess to return `stdout` and `stderr`. 
`removeDirectory` still doesn't work, even after disabling windows defender and adding the range loop to retry. Issuing `ls -a`, I can see that there is a `.git` directory that is left over in each package, I think this may be causing permission issues. I've tried using `shutil.rmtree`, no luck on that either.

I will get to `compareOutput` soon.

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+# currently trying to start automating with urllib3, 
+#   will extend to other PyPI packages in the future
+
+packages = {
+    "urllib3" : {
+        "url": "https://github.com/urllib3/urllib3.git",
+        "requirements_file": "dev-requirements.txt",
+        "uncompiled_whl": "urllib3-1.25.3-py2.py3-none-any.whl",
```
tommyli3318: @kayhayen Are these file names going to change? Right now I have them hardcoded in, not sure if this will hold up as things get updated

tommyli3318:
Okay, `compareOutput` now works too.
For the urllib3 package, the output is now

```
Creating virtualenv for quick test:
--- python (stdout)
+++ nuitka (stdout)

@@ -12,7 +12,7 @@

 test\test_exceptions.py .............                                    [  9%]
 test\test_fields.py ...............                                      [ 11%]
 test\test_filepost.py ...........                                        [ 11%]
-test\test_no_ssl.py ..                                                   [ 12%]
+test\test_no_ssl.py .F                                                   [ 12%]
 test\test_poolmanager.py .........................                       [ 14%]
 test\test_proxymanager.py ...                                            [ 14%]
 test\test_queue_monkeypatch.py .                                         [ 14%]
@@ -47,5 +47,29 @@

 test\with_dummyserver\test_socketlevel.py .............................. [ 98%]
 ......................                                                   [100%]

-=========== 839 passed, 456 skipped, 113 warnings in 49.27 seconds ============
+================================== FAILURES ===================================
+__________________ TestImportWithoutSSL.test_import_urllib3 ___________________

+self = <test.test_no_ssl.TestImportWithoutSSL object at 0xxxxxxxxx>
+
+    def test_import_urllib3(self):
+>       import urllib3  # noqa: F401
+
+test\test_no_ssl.py:86:
+_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
+<frozen importlib._bootstrap>:983: in _find_and_load
+    ???
+<frozen importlib._bootstrap>:967: in _find_and_load_unlocked
+    ???
+<frozen importlib._bootstrap>:668: in _load_unlocked
+    ???
+_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
+
+spec = ModuleSpec(name='urllib3', loader=<class 'nuitka_module_loader'>)
+
+>   ???
+E   KeyError: 'urllib3'
+
+<frozen importlib._bootstrap>:638: KeyError
+====== 1 failed, 838 passed, 456 skipped, 113 warnings in 47.74 seconds =======
+
---------------------------------------------------------------------------------
exit_stdout: 1 exit_stderr: 0
Error, outputs differed.
```

tommyli3318:
Currently, the entire test only uses one `virtualenv`. Would it be a better idea to use separate ones for uncompiled and compiled?

tommyli3318:
Added `removeTimeDiff` function to ignore the time differences of pytest

tommyli3318:
I've color-coded the result outputs for each package. Now also prints a colored summary of all tests at the end. Will get to caching soon

kayhayen:
@tommyli3318 The caching refers to git clone mostly, make a test if it exists, then `git fetch`, `git reset --hard origin/branch_name` otherwise fresh clone.

tommyli3318:
@kayhayen got it, caching is finished :)

tommyli3318:
@kayhayen got it, caching is finished :)

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,381 @@
...
+            "pytz",
+            "pyyaml", # invalid command 'bdist_nuitka'
+        ):
+            if search_mode.abortIfExecuted():
+                break
+            continue
+
+        # clone package if not there, else update it
+        if not package_name in os.listdir(cache_dir):
+            os.system("git clone %s %s" % (details["url"], package_name))
```
kayhayen: Never do `not x in y`, always use `x not in y`. more readable, faster, only better. I think PyLint may complain about it, not sure.

kayhayen: Always check return codes of os.system to be 0 and abort otherwise.

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,381 @@
...
+            "numpy",
+            "pytz",
+            "pyyaml", # invalid command 'bdist_nuitka'
+        ):
+            if search_mode.abortIfExecuted():
+                break
+            continue
+
+        # clone package if not there, else update it
+        if not package_name in os.listdir(cache_dir):
```
kayhayen: Check directory existance with `os.path.isdir()` or `os.path.exists`

tests/PyPI-pytest/run_all.py
kayhayen: Can you try to use --shallow clone, --no-tags and potentially other bandwidth, disk size using measures. We do not need history, not all tags, and not all branches, I think there is a flag for that too. Using it, will make it do less transfers. Mind you, we want to run this a lot and upstreams can be wasteful with their tags, branches, etc.

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,381 @@
...
+            os.system("git clone %s %s" % (details["url"], package_name))
+        else:
+            os.system("cd %s && git fetch && git reset --hard origin" % package_name)
```
kayhayen: Also error check here, and if it fails, do the full git clone. I think that means you ought to do it first, potentially even without even checking for existance of the dir, the `cd` will fail in that case.

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,381 @@
...
+# isort:start
+
+from nuitka.tools.testing.OutputComparison import compareOutput
+from nuitka.tools.testing.Virtualenv import withVirtualenv
+from nuitka.utils.FileOperations import removeDirectory
+from nuitka.tools.testing.Common import createSearchMode, my_print
+from nuitka.utils.AppDirs import getCacheDir
+import nuitka
+
+class bcolors:
```
kayhayen: Nice and easy, like it.

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,381 @@
...
+
+    my_print(
+        bcolors.YELLOW,
+        "TOTAL NUMBER OF PACKAGES TESTED: %s" % len(results),
+        bcolors.ENDC
+    )
+
+    num_failed = sum(y or z for _,y,z in results)
+
+    my_print(
```
kayhayen: I wonder, if `my_print` should not learn about colors, with a `color = "yellow"` parameter style, what is your take on that, would make it re-usable.

nuitka/tools/testing/Virtualenv.py
```diff
@@ -48,9 +48,47 @@ def runCommand(self, commands):
...
+                setup_commands = [". bin/activate"] + setup_commands
+
+            setup_command = " && ".join(setup_commands)
+            assert os.system(setup_command) == 0, setup_command
+
+
+            if type(popen_args) in (str, unicode):
+                popen_args = popen_args.strip().split()
+
+            if popen_args[0] == "python":
```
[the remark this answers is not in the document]
tommyli3318: Seems like this was from an old commit which has already been changed.

nuitka/utils/FileOperations.py
```diff
@@ -248,12 +248,11 @@ def removeDirectory(path, ignore_errors):
 
     def onError(func, path, exc_info):
         # Try again immediately, ignore what happened, pylint: disable=unused-argument
-        try:
-            func(path)
-        except OSError:
-            time.sleep(0.1)
-
-        func(path)
+        for _i in range(10):
```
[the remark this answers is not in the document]
tommyli3318: Not anymore, I've reverted the changes

tests/PyPI-pytest/run_all.py
tommyli3318: got it

tommyli3318: Got it!

tests/PyPI-pytest/run_all.py
tommyli3318: Ok

tests/PyPI-pytest/run_all.py
tommyli3318: Sounds good.

tommyli3318:
@kayhayen Ok, took all your suggestions and added `style` keyword argumnet to `my_print` :)

kayhayen:
@tommyli3318 Can you fix the Travis issues? I am considering the merge or use this, but it should also work with Python2 for that.

tommyli3318:
@kayhayen Fixed Travis issues :)

nuitka/Tracing.py
```diff
@@ -1,62 +1,88 @@
...
-def my_print(*args, **kwargs):
-    """ Make sure we flush after every print.
-
-    Not even the "-u" option does more than that and this is easy enough.
-    """
-
-    print(*args, **kwargs)
-
-    flushStdout()
+#     Copyright 2019, Kay Hayen, mailto:kay.hayen@gmail.com
```
kayhayen: Suspicious how this whole file is changed, are you using the commit hook for autoformat as described in the developer manual? This should be caused by you using Windows new lines or misconfigured git that switches line endings at commit. Please check that out.

nuitka/tools/testing/OutputComparison.py
```diff
@@ -202,7 +222,7 @@ def compareOutput(
     diff = difflib.unified_diff(
         makeDiffable(out_cpython, ignore_warnings, ignore_infos, syntax_errors),
         makeDiffable(out_nuitka, ignore_warnings, ignore_infos, syntax_errors),
-        "{program} ({detail})".format(program=os.environ["PYTHON"], detail=kind),
+        "{program} ({detail})".format(program=os.environ["PYTHON"] if "PYTHON" in os.environ else "python", detail=kind),
```
kayhayen: Use `os.environ.get` to achieve a default value. However, why is it not set, I think that `setup` function we use for testing, is something you also ought to use and then it's guarantueed to be set.

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,551 @@
+#!/usr/bin/env python
+#     Copyright 2019, Kay Hayen, mailto:kay.hayen@gmail.com
```
kayhayen: Put your name here.

nuitka/tools/testing/OutputComparison.py
tommyli3318: not sure, it wasn't set in my os.environ so this was a quick fix

tommyli3318:
Updated with autoformatting

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,472 @@
...
+            continue
+
+        if package_name in ("pycparser", "numpy"):
+            reportSkip("Not yet supported, see Issue #477", ".", package_name)
+            if search_mode.abortIfExecuted():
+                break
+            continue
+
+        if package_name in ("pytz",):  # indirect usage of distutils
+            reportSkip("Not yet supported", ".", package_name)
```
kayhayen: Issue number?

tests/PyPI-pytest/run_all.py
```diff
@@ -0,0 +1,472 @@
...
+            if search_mode.abortIfExecuted():
+                break
+            continue
+
+        if package_name in (
+            "google-auth",  # bdist_nuitka fails AttributeError: single_version_externally_managed
+            "jinja2",  # ModuleNotFoundError: No module named 'jinja2.tests'
```
kayhayen: Seems we found a case, where the namespace trick will be needed, jinja2.

nuitka/tools/testing/OutputComparison.py
```diff
@@ -202,7 +226,9 @@ def compareOutput(
     diff = difflib.unified_diff(
         makeDiffable(out_cpython, ignore_warnings, ignore_infos, syntax_errors),
         makeDiffable(out_nuitka, ignore_warnings, ignore_infos, syntax_errors),
-        "{program} ({detail})".format(program=os.environ["PYTHON"], detail=kind),
+        "{program} ({detail})".format(
```
kayhayen: One important thing, *never* add defaults where you lack the knowledge to invent them. In this function, how can you know what has been used, it may not be `python` at all.

Luckily, and not by chance, there is no way that this is needed, after calling setup, which every test suite does, it will be set. I think you didn't do that initially, which broke assumptions like this. This should be removable.

kayhayen:
I guess, we can merge this, and let me attempt to run this test. For nightshift CI, one of the concerns would be that it uses the network for git, and as such is vulnerable, but what can we do otherwise, this will have to be that way.

The changes requested are minor, and I will merge after you do them, with more to come. I believe the Jinja2 needs the environment variables hack, which we should focus on next.

tommyli3318:
Can you explain the `Jinja2` environment variables hack? Why is the `ModuleNotFoundError` even occurring in the first place? 
Package `rsa` recently switched to Poetry in [this commit](https://github.com/sybrenstuvel/python-rsa/commit/fcf5b7457c70426a242b17db20dd4e34e1055f69) and removed their `setup.py`, what should we do in that case?
Package `pytz` also recently updated their repo, they moved their `setup.py` into their `src` folder and I can no longer run `setup.py bdist_*` due to an `AssertionError`. Looks like it is their problem. I will hold off on making an issue until this is fixed

tommyli3318:
@kayhayen 
Ok, should be ready for merge now.

PR #442, opened 2019-06-24
Optimize built-in "max" and "min"

bksahu:
### What does this PR do?
This is a WIP. Optimizes "max" and "min" built-in

#232 

### PR Checklist


nuitka/optimizations/OptimizeBuiltinCalls.py
```diff
@@ -911,6 +914,116 @@ def wrapEvalBuiltin(source, globals_arg, locals_arg, source_ref):
...
+            name="max_call",
+            source_ref=source_ref,
+        )
+
+        call_args = node.getCallArgs()
+
+        assert call_args.getIterationLength() >= 2
+
+        max_arg_variables = [
+            provider.allocateTempVariable(temp_scope=None, name="max_arg_%d" % i)
```
bksahu: @kayhayen All the tests are falling for `max` with all constant arguments. However when I tried to compile the same works just fine. Reading the traceback I found that we get an `AssertionError` https://github.com/Nuitka/Nuitka/blob/2d1759dbc83d7387fbce56bf6b4871bab755ba64/nuitka/nodes/NodeBases.py#L747

which makes sure that no duplicate variables are allocated. As we are passing `name="max_args_%d" % i` we are not violating it either.  I can't figure out why it failing the test while working fine when I try to compile the same test cases.

kayhayen: Ah my bad, the `outline_body` should be the one doing the allocation, otherwise the next `max` usage will do it.

The outline body has a temporary variable scope of its own, so 2 outline bodies can repeat.

nuitka/optimizations/OptimizeBuiltinCalls.py
```diff
@@ -911,6 +914,114 @@ def wrapEvalBuiltin(source, globals_arg, locals_arg, source_ref):
     )
 
 
+def max_extractor(node):
+    @calledWithBuiltinArgumentNamesDecorator
+    def wrapMaxBuiltin(source, source_ref):
```
bksahu: PyLint is complaining due to the unused variable `source`. However, it is required in the Paramer Specs to pass argument in the wrap function:
https://github.com/Nuitka/Nuitka/blob/45d791faedbcba5daf7b95412314d4f61a02f216/nuitka/specs/BuiltinParameterSpecs.py#L577

Should I supress the Pylint error?

PR #462, opened 2019-07-12
Optimize built-in "zip"

bksahu:
Relevant Issue: #232 

This is a work in progress.

# Todo
- [x] Run nightshift without the optimization
- [ ] Run nightshift with only the C code replaced, basically using your C code only.
- [ ] Run nightshift with the generic implementation.
- [ ] Look at cases to specialize and do that, again run all world test with that.


nuitka/optimizations/OptimizeBuiltinCalls.py
```diff
@@ -1000,6 +1015,188 @@ def wrapExpressionBuiltinExecCreation(
...
+        ]
+
+        # makeTryExceptSingleHandlerNode
+        statements += [
+            makeTryExceptSingleHandlerNode(
+                tried=assign_iter_statements,
+                exception_name="TypeError",
+                handler_body=makeRaiseExceptionExpressionFromTemplate(
+                    exception_type="TypeError",
+                    template="zip argument #%s must support iteration",
```
kayhayen: Why use a template with constant arguments, seems the value is known and `%d` would be better suited too, either way.

nuitka/optimizations/OptimizeBuiltinCalls.py
```diff
@@ -1000,6 +1015,188 @@ def wrapExpressionBuiltinExecCreation(
...
+                source_ref=source_ref,
+            )
+        )
+
+        statements.append(StatementLoop(body=loop_body, source_ref=source_ref))
+
+        statements.append(
+            StatementReturn(
+                ExpressionTempVariableRef(
+                    # TODO: Create that tmp_result
```
kayhayen: Without solving this TODO, it will crash (trigger an assertion with --`debug`) at run time, because the temp variable won't be assigned.

nuitka/optimizations/OptimizeBuiltinCalls.py
kayhayen: There may be a function that uses no template, but similarily does it.

nuitka/optimizations/OptimizeBuiltinCalls.py
```diff
@@ -1000,6 +1015,188 @@ def wrapExpressionBuiltinExecCreation(
...
+                    exception_type="TypeError",
+                    template="zip argument #%s must support iteration",
+                    template_args=makeConstantRefNode(
+                        constant=str(i + 1), source_ref=source_ref
+                    ),
+                    source_ref=source_ref,
+                ),
+                source_ref=source_ref,
+            )
+            for i, _ in enumerate(zip_arg_variables)
```
kayhayen: I have used `range(len())` in similar cases and I prefer it.

kayhayen (changes requested):
Almost there. :-)

nuitka/optimizations/OptimizeBuiltinCalls.py
bksahu: I think I have resolved this TODO in Line 1113. But it's still crashing

bksahu: And I'm getting `core dump`

kayhayen: Allocating a variable, merely creates an object, it does nothing with it. You need to assign a value to that variable for it to be assigned. I am not referring to `tmp_result` as a Python level variable in your re-formulation. I am referring to it as the run time temporary variable of the outline, which without a `StatementAssignVariable` is going to be very crash happy if referenced before or without it.

bksahu:
> Isn't that what `enumerate` is for? The use of 'i' and 'values[i]' going over the `len` is less readable, although of course correct.

I thought this way is more efficient than the `enumerate` way because `enumerate` is not optimized yet. I will change it.

bksahu:
> I recall checking and seeing that `zip()` for Python3 is not a constant like `[]`, but still a generator, so just remove the special case. It might be invisible unless used outside of type checks or print statements of the repr, but it's still true. 

Yes, you are right `zip()` in Python3 returns a `zip_object` rather than `[]`. That's why in my special case I have returned `zip()`. 

https://github.com/Nuitka/Nuitka/pull/462/files#diff-fbba5af83ad563967aff7b03a581677fR1183

kayhayen:
The enumerate is on the Nuitka compile time level, we don't care about performance there. You are right that there is no built-in emumerate node, and I think we should definitely strive to get at it, as it's going to be cruical for loop optimization.

kayhayen:
Also, I am unclear which branch you mean, but the code where you create a constant from `zip()` will refuse to do so, because it is not a compile time constant. Just put it to the test and see how your 0 args special class will fail. And if it does not, render me scared. :)

kayhayen:
Checkout this:

>  File "../../nuitka/nodes/ConstantRefNodes.py", line 996, in makeConstantRefNode
>    assert False, (constant, constant_type)
> AssertionError: (<zip object at 0x7fac6ced8408>, <class 'zip'>)

https://nightshift.nuitka.net/#/builders/176/builds/15

There you have it...

nuitka/nodes/BuiltinZipNodes.py
```diff
@@ -0,0 +1,242 @@
...
+        # Code generation expects this to be there.
+        statements.append(StatementGeneratorReturnNone(source_ref=self.source_ref))
+
+        generator_body.setBody(
+            makeStatementsSequence(
+                statements=statements, allow_none=False, source_ref=self.source_ref
+            )
+        )
+
+        return outline_body, "new_expression", "Lowered zip built-in to generator."
```
kayhayen: This should return an iterator node, not the iterated value.

kayhayen: `computeExpressionIter1` is expected to potentially change the iterator node iteration value, e.g. a list is lowered to a tuple, but returned an iterable value that is not iterated, is going to cause issues.

nuitka/optimizations/OptimizeBuiltinCalls.py
Omar111Omar: Hi guys

Omar111Omar: It's okay

Omar111Omar: Please I want to know How share my projects on GitHub?

PR #480, opened 2019-08-03
Added MetaClass of FunctionType

Nimishkhurana:


nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -25,6 +25,38 @@
...
+}
+
+static PyMethodDef Meta_methods[] = {
+    {"type", (PyCFunction) Check_instance, METH_NOARGS,
+     "Return the name, combining the first and last name"
+    },
+    {NULL}  /* Sentinel */
+};
+
+static PyTypeObject MetaType = {
```
kayhayen: I think you need to make "type" a base class, the crash is probably related to you "object" derived meta class being used a meta class, but not really being one. This is probably cruicial, but ought to lead to an exception, alAlso you didn't `PyType_Ready` it before you used it as a metaclass.

Compile with `--debug` and also check the return codes in an assert statement. We don't do that yet, but that was probably stupid of me.

nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -570,7 +602,12 @@ PyTypeObject Nuitka_Function_Type = {
 #endif
 };
 
-void _initCompiledFunctionType(void) { PyType_Ready(&Nuitka_Function_Type); }
+void _initCompiledFunctionType(void) {
+    // PyModule_AddObject(Nuitka_Function_Type.m_module, "Meta", (PyObject *) &MetaType);
```
kayhayen: Like I said, there is no module for them. We could consider having one, but right now, I see no point in that. Maybe it could allow sharing implementations across Nuitka of same version with multiple extension modules once we have ABI compatibility over time. But that is more a comment to myself.

nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -570,7 +602,12 @@ PyTypeObject Nuitka_Function_Type = {
...
+     Nuitka_Function_Type.ob_type = &MetaType;
+    //  Py_TYPE(&Nuitka_Function_Type) = &MetaType;
```
kayhayen: The uncommented version is more explicit. The commented one can be removed I guess.

kayhayen:
Hello @Nimishkhurana did you have a chance to try out my suggestions yet?

Nimishkhurana:
Hello @kayhayen I wil try them now. I am busy with some academic work currently, so did not get enough time,

nuitka/build/static_src/CompiledFunctionType.c
Nimishkhurana: I have a doubt where can I `PyType_Ready` the MetaType. In the `_initCompiledFunctionType()` or should another `PyMODINIT_FUNC` function should be made.

kayhayen: Just "ready" it before you "ready" the function type. Later we can think of sharing it with others, but now we are not going for re-usable to other compiled type yet.

nuitka/build/static_src/CompiledFunctionType.c
Nimishkhurana: @kayhayen but that doesn't works. It gives error `error: ‘PyTypeObject {aka struct _typeobject}’ has no member named ‘ob_type’`.

kayhayen: Check out this: https://docs.python.org/3/c-api/typeobj.html

kayhayen:
Hello @Nimishkhurana how is it going? :-)

Nimishkhurana:
Hello @kayhayen, I had made some changes 2 months ago. Can you please review :-)

kayhayen:
Sorry for the delay, I had private issues and had to neglect Nuitka during the time. I am resuming now and will check this out this week.

kayhayen:
Can you try to rebase this, I would want to continue with this.

kayhayen:
See the original issue, the problem will be solved without a meta class, so this is obsolete now.

Issue #479, opened 2019-08-03
bdist_nuitka fails for packages containing py_modules only


tommyli3318:
`python setup.py bdist_nuitka` fails for all packages containing py_modules only, including PyPI packages such as             `decorator`, `ipaddress`, and `pyparsing`.

Platform: Windows
Python 3.7.0 (x32)
Nuitka 0.6.5rc3

I have created a distutils example to reproduce this error.
Steps to reproduce:
1. `git clone -b develop-pypi https://github.com/tommyli3318/Nuitka.git`
2. `cd tests/distutils/`
3. `python run_all.py only py_modules_only/`

Full Traceback:
```
Traceback (most recent call last):
  File "setup.py", line 34, in <module>
    scripts = ["runner"],
  File "C:\Users\Tommy\Nuitka\tests\distutils\venv_nuitka\lib\site-packages\setuptools\__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "C:\Users\Tommy\AppData\Local\Programs\Python\Python37-32\Lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Users\Tommy\AppData\Local\Programs\Python\Python37-32\Lib\distutils\dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "C:\Users\Tommy\AppData\Local\Programs\Python\Python37-32\Lib\distutils\dist.py", line 985, in run_command
    cmd_obj.run()
  File "C:\Users\Tommy\Nuitka\tests\distutils\venv_nuitka\lib\site-packages\wheel\bdist_wheel.py", line 192, in run
    self.run_command('build')
  File "C:\Users\Tommy\AppData\Local\Programs\Python\Python37-32\Lib\distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\Users\Tommy\AppData\Local\Programs\Python\Python37-32\Lib\distutils\dist.py", line 985, in run_command
    cmd_obj.run()
  File "C:\Users\Tommy\Nuitka\tests\distutils\venv_nuitka\lib\site-packages\nuitka-0.6.6rc1-py3.7.egg\nuitka\distutils\bdist_nuitka.py", line 58, in run
    self.main_package = self.compile_packages[0]
TypeError: 'NoneType' object is not subscriptable
```

kayhayen:
[...]

PR #483, opened 2019-08-10
Add support for py_modules_only compilation

tommyli3318:
# What does this PR do?
Create support handling the case of compiling packages containing py_modules only. Currently, Nuitka assumes that a main package always exists and runs into an exception when it is not the case.

Issue #479 
GSoC2019

# PR Checklist


tommyli3318:
@kayhayen Need some pointers here
With the changes, `python run_all.py only py_modules_only` runs into the following exception:

```
Nuitka:WARNING:Recursed to module 'py_modules_only' at 'C:\Users\Tommy\Nuitka\tests\distutils\py_modules_only\build\lib\py_modules_only.py' twice.
Nuitka:INFO:Included compiled module 'py_modules_only'.
error: [WinError 3] The system cannot find the path specified: 'C:\\Users\\Tommy\\Nuitka\\tests\\distutils\\py_modules_only\\build\\lib\\py_modules_only'
Creating virtualenv for quick test:
  File "run_all.py", line 257, in <module>
    main()
  File "run_all.py", line 169, in main
    commands=['cd "%s"' % case_dir, "python setup.py bdist_nuitka"]
  File "C:\Users\Tommy\Nuitka\nuitka\tools\testing\Virtualenv.py", line 51, in runCommand
    assert os.system(command) == 0, command
AssertionError: call scripts\activate.bat && cd "C:\Users\Tommy\Nuitka\tests\distutils\py_modules_only" && python setup.py bdist_nuitka
```

It seems that `py_modules_only` is assumed to be a directory? How should I approach this

kayhayen:
Well, I recommend you trace the command before executing it.

Telling Nuitka that a module should be included as a package, is bound for disaster. No wonder, it's not found, it's a file and not a directory.

On modules, use `--include-module` on packages use `include-package`, do not fix them. Probably the error messages from Nuitka are not good or not there? Not sure how robust Nuitka is against your errors there.

Yours,
Kay

tommyli3318:
@kayhayen I fixed it to use only `--include-module` on modules. However I am still getting the same error.

Traceback:
```
Nuitka:WARNING:Recursed to module 'py_modules_only' at 'C:\Users\Tommy\Nuitka\tests\distutils\py_modules_only\build\lib\py_modules_only.py' twice.
Nuitka:INFO:Included compiled module 'py_modules_only'.
error: [WinError 3] The system cannot find the path specified: 'C:\\Users\\Tommy\\Nuitka\\tests\\distutils\\py_modules_only\\build\\lib\\py_modules_only'
Creating virtualenv for quick test:
  File "run_all.py", line 257, in <module>
    main()
  File "run_all.py", line 169, in main
    commands=['cd "%s"' % case_dir, "python setup.py bdist_nuitka"]
  File "C:\Users\Tommy\Nuitka\nuitka\tools\testing\Virtualenv.py", line 51, in runCommand
    assert os.system(command) == 0, command
AssertionError: call scripts\activate.bat && cd "C:\Users\Tommy\Nuitka\tests\distutils\py_modules_only" && python setup.py bdist_nuitka
```

kayhayen:
Can you show the resulting Nuitka command then?

tommyli3318:
`
C:\Users\Tommy\Nuitka\tests\distutils\venv_nuitka\Scripts\python.exe -m nuitka --module --plugin-enable=pylint-warnings --output-dir=C:\Users\Tommy\Nuitka\tests\distutils\py_modules_only\build\lib --nofollow-import-to=*.tests --show-modules --remove-output --include-module=py_modules_only C:\Users\Tommy\Nuitka\tests\distutils\py_modules_only\build\lib\py_modules_only.py
`

kayhayen:
The `--include-module` on the main module is not needed, unlike with packages, which will not automatically be fully included, but only top module, there is no extra benefit from that.

I cannot decipher the error message, it might not be from Nuitka, is it? I guess it is true:

> error: [WinError 3] The system cannot find the path specified: 'C:\\Users\\Tommy\\Nuitka\\tests\\distutils\\py_modules_only\\build\\lib\\py_modules_only

However, what is trying to use it?

nuitka/distutils/bdist_nuitka.py
```diff
@@ -53,7 +53,15 @@ def run(self):
         # for .py_modules
 
         self.compile_packages = self.distribution.packages
-        self.main_package = self.compile_packages[0]
+        if self.compile_packages:
+            self.main_target = self.compile_packages[0]
+            self.py_modules = False
+        else:
```
kayhayen: Like I said, I am not convinced that not both are allowed to be used at the same time. Can you check what happens if one specifies both?

nuitka/distutils/bdist_nuitka.py
```diff
@@ -121,12 +129,16 @@ def _buildPackage(self, build_lib):
...
-            "--include-package=%s" % self.main_package,
             "--nofollow-import-to=*.tests",
             "--show-modules",
             "--remove-output",
         ]
 
+        if self.py_modules:
+            command += ["--include-module=%s" % self.main_target]
+        else:
+            command += ["--include-package=%s" % self.main_target]
```
kayhayen: Only use `--include-package` if the main target is one.

nuitka/distutils/bdist_nuitka.py
```diff
@@ -158,9 +170,7 @@ def _buildPackage(self, build_lib):
         else:
             # Delete the entire source copy of the module
             shutil.rmtree(
-                os.path.join(
-                    self.build_lib, self.main_package.replace(".", os.path.sep)
-                )
+                os.path.join(self.build_lib, self.main_target.replace(".", os.path.sep))
```
kayhayen: Ah ok, for packages this is a thing, but not for modules. For those just delete it as a file, and I guess, the `.py` must be protected from replacing maybe. Trace what do you here, and do the right thing.

But what also strikes me as wrong, is that not all the packages, and not all the module files are deleted, but just the main one. It should be done for all or none I guess.  We could really use Nuitka to learn to `--replace` when it compiles, i.e. delete the input files included. Doing it on the outside is kind of harder to do, but doable, so lets stick to it for now.

For now, this code will probably have issues when multiple packages are included, which is something people do, then I think things might survive. On the other hand, are not Python files all deleted anyway? Can you check if this delete here is even needed?

nuitka/importing/Importing.py
```diff
@@ -217,7 +217,13 @@ def findModule(importing, module_name, parent_package, level, warn):
     # We have many branches here, because there are a lot of cases to try.
     # pylint: disable=too-many-branches
 
-    assert type(module_name) is ModuleName
+    if type(module_name) is not ModuleName:
```
kayhayen: That's definitely not the way to do it. You either obey to its demand, or you convert to `ModuleName` doing `module_name = ModuleName(module_name)` but you don't micro manage types like this.

I also think you are calling it with `ModuleName` in above code. Please just remove this change.

tommyli3318: Without this change, I get an `AssertionError`
```
  File "C:\Users\Tommy\Nuitka\tests\distutils\venv_nuitka\lib\site-packages\nuitka-0.6.6rc3-py3.7.egg\nuitka\__main__.py", line 184, in <module>
    main()
  File "C:\Users\Tommy\Nuitka\tests\distutils\venv_nuitka\lib\site-packages\nuitka-0.6.6rc3-py3.7.egg\nuitka\__main__.py", line 177, in main
    MainControl.main()
  File "C:\Users\Tommy\Nuitka\tests\distutils\venv_nuitka\lib\site-packages\nuitka-0.6.6rc3-py3.7.egg\nuitka\MainControl.py", line 753, in main
    main_module = createNodeTree(filename=filename)
  File "C:\Users\Tommy\Nuitka\tests\distutils\venv_nuitka\lib\site-packages\nuitka-0.6.6rc3-py3.7.egg\nuitka\MainControl.py", line 130, in createNodeTree
    warn=False,
  File "C:\Users\Tommy\Nuitka\tests\distutils\venv_nuitka\lib\site-packages\nuitka-0.6.6rc3-py3.7.egg\nuitka\importing\Importing.py", line 220, in findModule
    assert type(module_name) is ModuleName
AssertionError
```

I have no idea why, because as you said, it is called with `ModuleName` already...

tommyli3318: I added debug code to print out `type(module_name)` before the assertion. The result is `<class 'str'>`. Somehow `ModuleName` was not applied correct to the str? Im confused

nuitka/distutils/bdist_nuitka.py
tommyli3318: There are no .py files in the build folder even without this delete for package example_1, that means this is unnecessary, correct?

nuitka/distutils/bdist_nuitka.py
tommyli3318: I've included a new distutils example `package_and_module`. Nuitka builds it without error but `ls package_and_module/build/lib/` outputs 

```
example1_package/  libsome_module.a  some_module.pyd*  some_module.pyi
```

Which seems to be missing the .pyd and .pyi files from the package.

tommyli3318:
@kayhayen 
Made changes and replied to your comments above, please review :). The script now builds `distutils/py_modules_only` without error, but needs to be tweaked to support the case of both package and module I think

kayhayen:
Instead of printing values at crash time, your go to solution should be to consider the callstack and see what it puts there. Your imaginary call stack, your usage of `findModule` is not the only one. You just encountered a regression of Nuitka develop itself. Please find it and make it a separate PR "fixup_import_module".

Seems we have no coverage in testing for that option until your test.

Yours,
Kay

kayhayen:
@JorjMcKie The regression I just noted, should also affect your hinter maybe? Does it not use `--include-module` command line, or is it purely doing its stuff via plugin calls.

JorjMcKie:
The hinted-mods.py user-plugin does this to adjust Nuitka compilation parameters:
* directly ``append`` entries to ´´options.options.plugins_enabled``
* directly ``append`` entries to ``options.recurse_modules``

So it does not (yet) touch `` options.include_modules`` or ``options.include_packages``.

kayhayen:
Ok, good to know, then it is not affected by the regression, and I need to panic rebase you :)

JorjMcKie:
Anytime, just say when 😊

Von meinem iPhone gesendet

tommyli3318:
> Please find it and make it a separate PR "fixup_import_module".

#487 created

kayhayen:
Good job, I really like how this converged to unify the two cases.

PR #495, opened 2019-08-18
Improve pypi automation

tommyli3318:
# What does this PR do?
Improve automated PyPI test suite

GSoC 2019

# PR Checklist

### DIFF
--- a/nuitka/tools/testing/OutputComparison.py
+++ b/nuitka/tools/testing/OutputComparison.py
@@ -82,7 +82,7 @@ def normalizeTimeDiff(outputStr):
     =================== 1059 passed, 8 warnings in x.xx seconds ===================
     """
 
-    matchObj = re.search(b"in [0-9]+.[0-9][0-9] seconds", outputStr)
+    matchObj = re.search(b"in [0-9]+.[0-9][0-9](s| seconds)", outputStr)
     if matchObj:
         return (
             outputStr[: matchObj.start()]
--- a/tests/PyPI-pytest/run_all.py
+++ b/tests/PyPI-pytest/run_all.py
@@ -177,7 +177,8 @@ def main():
                     commands=[
                         "python -m pip install -U %s"
                         % os.path.join(dist_dir, os.listdir(dist_dir)[0]),
-                        "python -c print(getattr(__import__('%s'),'__compiled__','__uncompiled_version__'))"
+                        # use triple quotes for linux
+                        """python -c "print(getattr(__import__('%s'),'__compiled__','__uncompiled_version__'))" """
                         % details.get("package_name", package_name),
                     ]
                 )
--- a/tests/PyPI-pytest/run_all.py
+++ b/tests/PyPI-pytest/run_all.py
@@ -33,6 +33,7 @@ Virtualenv is used to create a clean environment with no outside pollution.
 
 import os
 import sys
+import json
 
 # Find nuitka package relative to us.
 sys.path.insert(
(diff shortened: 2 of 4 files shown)


nuitka/tools/testing/OutputComparison.py
```diff
@@ -82,7 +82,9 @@ def normalizeTimeDiff(outputStr):
     =================== 1059 passed, 8 warnings in x.xx seconds ===================
     """
 
-    matchObj = re.search(b"in [0-9]+.[0-9][0-9] seconds", outputStr)
+    matchObj = re.search(b"in [0-9]+.[0-9][0-9]s", outputStr) or re.search(
```
kayhayen: Are you familiar wird regular expressions? Because that kind of `or` is now how you use them, instead you make am or inside the re, let me know if you need help with that.

tests/PyPI-pytest/run_all.py
```diff
@@ -383,7 +388,8 @@ def main():
                     commands=[
                         "python -m pip install -U %s"
                         % os.path.join(dist_dir, os.listdir(dist_dir)[0]),
-                        "python -c print(getattr(__import__('%s'),'__compiled__','__uncompiled_version__'))"
+                        # use triple quotes for linux
+                        """python -c "print(getattr(__import__('%s'),'__compiled__','__uncompiled_version__'))" """
```
kayhayen: if that is all it took, it makes me feel stupid, however it *ought *to be all it takes. :)

nuitka/tools/testing/OutputComparison.py
tommyli3318: `matchObj = re.search(b"in [0-9]+.[0-9][0-9](s| seconds)", outputStr)`
Something like that?

tommyli3318:
@kayhayen 

Sometimes pytest would output different amounts of equal signs for the summary (probably because one less "=" is printed to make space for the extra digit of time)

'''
-================ 1977 passed, 76 skipped, 21 xfailed in x.xx seconds =================
+================ 1977 passed, 76 skipped, 21 xfailed in x.xx seconds ================
'''

What would be a good way to normalize this? `re.findall(b"=+")`  doesn't seem like a good idea

kayhayen:
That's a really funny type of difference, I guess, it does `center` like this;

```
> " allo ".center(20, "=")
'======= allo ======='
```

So it could be on either side of the text, could it not be?

tommyli3318:
> So it could be on either side of the text, could it not be?

Actually no, it always prefers right-centered if it cannot be perfectly centered

```
>>> " allo ".center(9, "=")
'== allo ='
```

tommyli3318:
Oh no, what did I just do ): I `git fetch upstream` and `git rebase upstream/develop` and now the differences are messed up

tommyli3318:
Should I move the `gitClone` function to another file? Maybe `Common.py`?

nuitka/tools/testing/OutputComparison.py
kayhayen: Looks good to me, should work.

tests/PyPI-pytest/run_all.py
```diff
@@ -289,11 +129,9 @@ def main():
...
             "google-auth",  # bdist_nuitka fails AttributeError: single_version_externally_managed
             "jinja2",  # ModuleNotFoundError: No module named 'jinja2.tests'
-            "pandas",  # python setup.py egg_info fails
-            "pytz",  # AssertionError: zoneinfo files not found!
-            "setuptools",  # compiled __import__ check fails; pytest passes
-            "wheel",  # compiled __import__ check fails; pytest passes
-            "pytest",  # compiled __import__ check fails; should we even test this?
+            "pandas",  # ModuleNotFoundError: No module named 'Cython'
+            "pytz",  # need to 'make build'
+            "rsa",  # Now uses Poetry (no setup.py)
```
kayhayen: This poetry thing is something I encountered elsewhere too, can you look into how it could be plugged with Nuitka?

kayhayen:
Can you rebase it to current develop once more, then it will check the new doc with restlint automatically, so we can find errors in the formatting automatically?

tommyli3318:
> Can you rebase it to current develop once more, then it will check the new doc with restlint automatically, so we can find errors in the formatting automatically?

Ok

tests/PyPI-pytest/run_all.py
tommyli3318: they are using a `tox.ini` file instead of `setup.py` from what I see

kayhayen: I think `tox.ini` is just a way of making it test with multiple Python versions, but in the end, it should be still something like `py.test`, esp. with tox I think it's extremely common. What is in that tox.ini ?

kayhayen:
Thanks for making it perfect. :-)

Issue #473, opened 2019-07-29
PyQt5 standalone fails with `Incompatible library version` on macOS


melvyn2:
```bash
$ python3 -m nuitka --version
0.6.6rc1
Python: 3.7.4 (default, Jul  9 2019, 18:13:23) 
Executable: /usr/local/opt/python/bin/python3.7
OS: Darwin
Arch: x86_64
```
Python was installed with homebrew, and nuitka with `pip install https://github.com/Nuitka/Nuitka/archive/develop.zip`.

Test example:
```python
import sys
from PyQt5 import QtWidgets

def main():
    app = QtWidgets.QApplication(sys.argv)
    window = QtWidgets.QMainWindow()
    button = QtWidgets.QPushButton("Hello, PyQt!")
    window.setCentralWidget(button)
    window.show()
    app.exec_()

if __name__ == '__main__':
    main()
```
The program was saved as tmp.py and compiled with `xcrun -sdk macosx python3.7 -m nuitka --standalone --follow-imports --plugin-enable=qt-plugins tmp.py`. `xcrun` was required to provide certain header files. Here is what happens when the executable is then ran:
```
$ ./tmp.dist/tmp
Traceback (most recent call last):
  File "/Users/melvyn/tmp/./tmp.dist/tmp.py", line 2, in <module>
ImportError: dlopen(/Users/melvyn/tmp/./tmp.dist/PyQt5/QtWidgets.so, 2): Library not loaded: @rpath/QtWidgets.framework/Versions/5/QtWidgets
  Referenced from: /Users/melvyn/tmp/./tmp.dist/PyQt5/QtWidgets.so
  Reason: Incompatible library version: QtWidgets.so requires version 5.13.0 or later, but QtWidgets provides version 5.7.0
```
Here is the full project folder: 
[tmp.zip](https://github.com/Nuitka/Nuitka/files/3443017/tmp.zip)

kayhayen:
In your tmp.zip there is no QtWidgets file. Can you have a look at what `_detectBinaryPathDLLsMacOS` does in your case? It should resolve the @rpath and then Nuitka would include the file in the distribution folder.

Yours,
Kay

melvyn2:
In `tmp.dist/PyQt5` there is a `QtWidgets.so` file. I'm a bit new to Nuitka, how should I run `_detectBinaryPathDLLsMacOS` (as in, what arguments to pass)?

kayhayen:
[...]

PR #502, opened 2019-08-25
PyQt fix

melvyn2:
# What does this PR do?
Resolves @rpath-s properly when copying dependencies and linking them
This PR fixes issue #473
# PR Checklist

- [ ] ~~Ideally new or changed features have documentation updates.~~

### DIFF
--- a/nuitka/freezer/Standalone.py
+++ b/nuitka/freezer/Standalone.py
@@ -669,7 +669,7 @@ def _detectBinaryPathDLLsPosix(dll_filename):
     return sub_result
 
 
-def _detectBinaryPathDLLsMacOS(original_dir, binary_filename):
+def _detectBinaryPathDLLsMacOS(original_dir, binary_filename, keep_unresolved=False):
     result = set()
 
     process = subprocess.Popen(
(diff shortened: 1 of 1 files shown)


kayhayen:
Sorry for the delay in merging this, I need to check this out, hopefully later today

nuitka/freezer/Standalone.py
```diff
@@ -669,7 +669,7 @@ def _detectBinaryPathDLLsPosix(dll_filename):
     return sub_result
 
 
-def _detectBinaryPathDLLsMacOS(original_dir, binary_filename):
+def _detectBinaryPathDLLsMacOS(original_dir, binary_filename, keep_unresolved=False):
```
kayhayen: This keep_unresolved is not used, right? Probably has no place as we expect everything to resolve?

melvyn2: No, is it used to get the original `@rpath` for `fixupBinaryDLLPaths()`. If it is passed the resolved paths, then `install_name_tool` won't be able to make the changes.

kayhayen:
Good job, I am merging this. There is two small things that I want to change. First, I think a default argument is not good, I am going to make it explicit. And second, the preserving if the file mode, while making it writable, is something we have as a context manage somewhere in Nuitka. I will work to expand our Travis to use Python3 too (currently only Python2) before releasing it.

PR #514, opened 2019-09-18
fix FileNotFoundError on adding resources

namuyan:
A problem with heavy binaries and depends on user's system
https://github.com/Nuitka/Nuitka/issues/468

**caution!** I could not specify a reason.

### DIFF
--- a/nuitka/utils/WindowsResources.py
+++ b/nuitka/utils/WindowsResources.py
@@ -251,6 +251,9 @@ def copyResourcesFromFileToFile(source_filename, target_filename, resource_kind)
 
 
 def addResourceToFile(target_filename, data, resource_kind, lang_id, res_name):
+    import os
+    target_filename = os.path.abspath(target_filename)
+
     update_handle = _openFileWindowsResources(target_filename)
 
     _updateWindowsResource(update_handle, resource_kind, res_name, lang_id, data)


kayhayen:
The build error of Travis has no connection to you, seems to be there problem, I have seen this occur before.

kayhayen:
Did this really occur in a repeatable fashion, if yes, it could be about what you suggested, a wrong separator, but I doubt that, and you could confirm with a trace, I am pretty religious about `os.path.join` for years now.

Another theory would be if relative file names are not relative to `pwd` there, but to `PATH`, which may make some sense for these kind of functions, and you might not have '.` in your `PATH`. An absolute filename would circumvent that. I will study MS documentation. 

For now, the only harm in absolute paths would be in how they can easily become too long to talk about them with Windows, or make unicode enter the picture, although you are on Python3, and there it's not the issue normally.

kayhayen:
Quoting: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-beginupdateresourcea

> The binary file in which to update resources. An application must be able to obtain write-access to this file; the file referenced by pFileName cannot be currently executing. If pFileName does not specify a full path, the system searches for the file in the current directory.

Makes me consider that there is a lock. You know how deleting a binary on Windows often fails because its lock is still held despite it being finished, e.g. due to virus checkers. On my system, I have disabled Windows Defender for the directories I am working on with Nuitka. Maybe this is a race condition, where that new file gets looked at and then you can't open it.

kayhayen:
I noticed, that for Python3 we use the "W" function all the time, and for Python2 the "A" function all the type, the type check on filename does not contribute in any way.

```
    if type(filename) is str and str is not bytes:
        BeginUpdateResource = ctypes.windll.kernel32.BeginUpdateResourceW
    else:
        BeginUpdateResource = ctypes.windll.kernel32.BeginUpdateResourceA
```

However, the args type defined towards ctypes may be wrong too, they are not the same. The ctypes may do a conversion of not for unicode pointers, but for `LPVOID` it will not do it. Since unicode strings have internally different representations, going with more or less ASCII could be a problem there.

https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-beginupdateresourcew

So I am trying out to use "LPCWSTR" for this code:

```   
     BeginUpdateResource.argtypes = [ctypes.wintypes.LPVOID, ctypes.wintypes.BOOL]
```

It may not have been defined or something like this, going to check that out now. But since os.path.join can change the unicode internal representation, that might explain things.

Yours,
Kay

kayhayen:
```
+        BeginUpdateResource.argtypes = [ctypes.wintypes.LPCWSTR, ctypes.wintypes.BOOL]
     else:
         BeginUpdateResource = ctypes.windll.kernel32.BeginUpdateResourceA
+        BeginUpdateResource.argtypes = [ctypes.wintypes.LPCSTR, ctypes.wintypes.BOOL]

-    BeginUpdateResource.argtypes = [ctypes.wintypes.LPVOID, ctypes.wintypes.BOOL]
```

For me this works equally well. I think I remember now, why the type check makes sense, filenames can be as `bytes` too, then converting to unicode would only be trouble.

kayhayen:
I will merge your fix now, as it's a fix, but this could all be coming down to a null termination by chance or not. 

The `LPVOID` most probably did not make ctypes do that, but often it will be the case. But then, once I convinced myself that the above change works, I will try replace it with that.

PR #526, opened 2019-09-23
Distutils: Allow slashes in package names

jayvdb:
compile_packages should use '.' separators, however setuptools
allows '/', as seen in package websockets.


kayhayen:
There is another use of `self.compile_packages[0]` that might have to be adapted, not sure though, probably `compile_packages` should be normalized for use in a local variable. I will clean that up later, I intend to make it use multiple commands anyway for multiple packages.

Issue #527, opened 2019-09-23
Documention for build_with_nuitka with pbr


jayvdb:
`setup(build_with_nuitka=True,pbr=True)` does *not* work, however it looks like `setup(pbr=True,build_with_nuitka=True)` does work, at least for a few I have tried.

jayvdb:
It looks like pbr causes problems, because it builds the list of packages, and then returns them in an unusual order, and nuitka distutils plugin only consumes the first one.

```console
[   63s] Nuitka:INFO:Recurse to import 'requests_mock.contrib' from 'requests_mock/contrib'. (Module is exact match of 'requests_mock.contrib' instructed by user to recurse to.)
[   63s] Nuitka:INFO:Recurse to import 'requests_mock.contrib._pytest_plugin' from 'requests_mock/contrib/_pytest_plugin.py'. (Module is package content of 'requests_mock.contrib' instructed by user to recurse to.)
[   63s] Nuitka:INFO:Recurse to import 'requests_mock.contrib.fixture' from 'requests_mock/contrib/fixture.py'. (Module is package content of 'requests_mock.contrib' instructed by user to recurse to.)
[   63s] Nuitka:INFO:Recurse to import 'requests_mock' from 'requests_mock'. (Module is exact match of 'requests_mock' instructed by user to recurse to.)
[   63s] Nuitka:INFO:Recurse to import 'requests_mock.adapter' from 'requests_mock/adapter.py'. (Module is package content of 'requests_mock' instructed by user to recurse to.)
[   63s] Nuitka:INFO:Recurse to import 'requests_mock.compat' from 'requests_mock/compat.py'. (Module is package content of 'requests_mock' instructed by user to recurse to.)
[   63s] Nuitka:WARNING:Recursed to package 'contrib' at '/home/abuild/rpmbuild/BUILD/requests-mock-1.7.0/build/lib/requests_mock/contrib' twice.
[   63s] Nuitka:WARNING:Recursed to module '_pytest_plugin' at '/home/abuild/rpmbuild/BUILD/requests-mock-1.7.0/build/lib/requests_mock/contrib/_pytest_plugin.py' twice.
[   63s] Nuitka:WARNING:Recursed to module 'fixture' at '/home/abuild/rpmbuild/BUILD/requests-mock-1.7.0/build/lib/requests_mock/contrib/fixture.py' twice.
[...]

PR #530, opened 2019-09-24
WIP: distutils: Use shortest package name as base

jayvdb:
Related to https://github.com/Nuitka/Nuitka/issues/527

### DIFF
--- a/nuitka/distutils/bdist_nuitka.py
+++ b/nuitka/distutils/bdist_nuitka.py
@@ -80,7 +80,7 @@ class build(distutils.command.build.build):
         package, main_filename, finding = findModule(
             importing=None,
             module_name=ModuleName(
-                self.compile_packages[0]
+                min(self.compile_packages)
                 if self.compile_packages
                 else self.py_modules[0]
             ),


jayvdb:
Very likely this should be improved by splitting the names on `.` and `/`, deps eliminated, and then returning the first top level package name.

It doesnt solve the problem of multiple top-level packages, but there are not many of those.

kayhayen:
There is an issue, where I would like to make Nuitka use `libnuitka` for the static parts. Right now, it is not issuing multiple commands, because most often, one package is always used first, and the second package is more of a slave package, and that kind of limits me.

For repeated compilation, relying on `ccache` for shared code is relatively indirect, and not always working. I intend to do that for the next release cycle.

However, in the mean time, we could in fact, for these rare cases, just duplicate and produce multiple extension modules for now. Inclusion of all in one, will be a hack and show limitations at one or the other instance.

kayhayen:
Also, I think, `something.b`, `something`, `another`, maybe still should be resolved to `something` if we maintain the single top level stuff.

More like removing all sub-packages of packages that are also in the list, before taking the first one, that I would feel might work more often. I do believe, people put the namesake package first, even if not always first. Can you cite, which ones triggered you here? @jayvdb

jayvdb:
All packages using pbr have the modules in the packages in random order.  See https://github.com/Nuitka/Nuitka/issues/527

kayhayen:
This has become obsolete, because of #572 removing the need to select a main package, but instead compiling multiple top level packages as necessary. Therefore, this workaround is no longer useful

Issue #543, opened 2019-10-04
Ordered compilation


jayvdb:
This is likely a part of https://github.com/Nuitka/Nuitka/issues/534

when the modules are being compiled, the list is in an unusual order, or has a random element to it.

This makes it harder to read the build log because it jumps between modules of different packages.

I believe that the set of included modules is finalised before scons is called, so the order of compilation (and linking) can be optimised for readability and for deterministic/reproducible builds.

PR #544, opened 2019-10-04
SingleExe.scons: Compile modules in alpha order

jayvdb:
The build logs are easier to read when the modules are compiled
in a logical order, and probably this helps to achieve
reproducible builds.

Closes https://github.com/Nuitka/Nuitka/issues/543

### DIFF
--- a/nuitka/build/SingleExe.scons
+++ b/nuitka/build/SingleExe.scons
@@ -1596,7 +1596,7 @@ def discoverSourceFiles():
         return target_file
 
     # Scan for Nuitka created source files, and add them too.
-    for filename in os.listdir(source_dir):
+    for filename in sorted(os.listdir(source_dir)):
         # Only C files are of interest here.
         if not filename.endswith((".c", "cpp")) or not filename.startswith(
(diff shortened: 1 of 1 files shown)


kayhayen:
I am a huge fan of determinism.

Notice that jobs may still execute in parallel, but e.g. linker command will be stable, and I would assume this will more likely produce identical outputs.

Yours,
Kay

nuitka/build/SingleExe.scons
```diff
@@ -1596,7 +1596,7 @@ def discoverSourceFiles():
         return target_file
 
     # Scan for Nuitka created source files, and add them too.
-    for filename in os.listdir(source_dir):
+    for filename in sorted(os.listdir(source_dir)):
```
kayhayen: Good, but more as a note to self, the scons file should start to use `nuitka.utils.FileOperations` in the mid run, which does this already.

jayvdb: Also a note to this self, as I'll likely be poking around in here again.

PR #545, opened 2019-10-04
MetaPathBasedLoader: Python3.8 PyInterpreterState

jayvdb:
On Python 3.8, the definition of PyInterpreterState has been
moved to internal headers.

### DIFF
--- a/nuitka/build/static_src/MetaPathBasedLoader.c
+++ b/nuitka/build/static_src/MetaPathBasedLoader.c
@@ -34,6 +34,12 @@
 #endif
 
 #include "nuitka/prelude.h"
+// Include definition of PyInterpreterState, hidden since Python 3.8
+#if PYTHON_VERSION >= 380
+#define Py_BUILD_CORE
+#include "internal/pycore_pystate.h"
+#undef Py_BUILD_CORE
+#endif
 #include "nuitka/unfreezing.h"
 
 #ifdef _WIN32


jayvdb:
Example at https://build.opensuse.org/package/show/home:jayvdb:python38-static:packages-Nuitka/python-story-clang

(ignore the 'static' in the URL; static disabled to verify this patch)

I'm happy to have this sit in waiting until we get Python 3.8 test rig happening.  Happy to help get the test rig working on Python 3.8, but cant help much this coming week.

nuitka/build/static_src/MetaPathBasedLoader.c
```diff
@@ -34,6 +34,12 @@
 #endif
 
 #include "nuitka/prelude.h"
+// Include definition of PyInterpreterState, hidden since Python 3.8
+#if PYTHON_VERSION > 380
```
kayhayen: This should probably be >= 380, otherwise it ought to not have an effect before 3.8.1

kayhayen:
The last Python3.8 that worked for me was beta3 btw, on both Linux and Windows, I think. 

I have updated to rc1 on Windows, and did not encounter this, although very likely, I might not have done standalone mode, now that I am thinking of it. I am currently updating on Linux, where compilation will take a while though.

nuitka/build/static_src/MetaPathBasedLoader.c
jayvdb: Yes, correct. I needed to change it. Forgot to repush.

jayvdb:
@kayhayen , I was thinking it is better to keep this very local so we would find any other bits that depend on internals, and avoid polluting the global namespace.

kayhayen:
@jayvdb you have a valid point there, also because I am not yet even aware, why the meta path based loader needs that level of access and the rest does not.

nuitka/build/static_src/MetaPathBasedLoader.c
kayhayen: Ok, please do so then :-)

kayhayen:
So, the Python 3.8 test suite was pushed, and I am working a bit on it. It found a crash in `test_asyncgen` with Python 3.8 showing a failed assertion and Python 3.7 showing a corruption.

Executing new tests with old Python sometimes reveals unfixed bugs of CPython, but often also bad error handling. I will try and narrow these down. Also running the test suite with "all" to get a feeling for how good everything is, should be on the agenda. I am aiming at working on and off on 3.8 support until my private life becomes stable again. That mostly is busy work and can be interrupted at any point.

nuitka/build/static_src/MetaPathBasedLoader.c
jayvdb: Got back from a mini-holiday and re-pushed as is with this bit fixed.
Happy to move it `prelude.h` if you think it is best now that you've played around with py3.8 a bit more.

I see the tests at https://github.com/Nuitka/Nuitka-CPython-tests/tree/CPython38 , but with timestamps of Jun 19, 2017.

kayhayen: Ah, and moving it, no, you convinced me. I will merge once I hit standalone mode, or whatever triggers the need for me, right now however, I was not working on this, my private issues continue.

kayhayen:
I found the reason it crashed with 3.7 and even 3.6 (asyncgen go no earlier) and fixed it. This is why it's always good to go first for old version with new tests, and then new version with old tests. I am doing the 3.8 tests for 3.8 only as the last thing. Surely I do not like to mix new test bugs with changes in the Python implementation, but to stand on stable grounds first.

kayhayen:
So, I am merging this now, btw factory is a very nearly good 3.8 support then, once I rebase on this I suppose, got most of the stuff implemented. I setup my CI, and this is indeed needed for non-Windows CI, to get at the flags to use for `dlopen`.

Nuitka does its own shared library loading, so it can handle special needs.

Issue #571, opened 2019-11-11
Distutils needs to support distinct package namespaces


kayhayen:
In `bdist_nuitka` right now, we pick a top level package, and include all packages below it, that fails the case of #527 and also doesn't work well when 2 packages are provided by one wheel.

For a wheel with packages names to be included:

```
[
    "something.a"
    "something.b"
    "something",
    "helper",
]
```

The result should not be one command, including them all, but instead, with multiple top level packages picked (here `something` and `helper`, multiple commands should be issued

```
# 2 commands 
--module something --include-package=something --include-package=something.a --include-package=something.b
--module helper 

-> something.pyd
-> helper.pyd
```

kayhayen:
This should solve many issues that @jayvdb has raised for good. Tommy has worked on this in GSoC 2019 and knows what to do.

kayhayen:
Once we have something workable, I guess that @jayvdb can put it through his OBS of all openSUSE again.

kayhayen:
This was implemented for 0.6.6 by @tommyli3318

PR #572, opened 2019-11-12
Distutils: Add support for distinct package namespaces

tommyli3318:
# What does this PR do?
Add `Distutils` support for wheels with multiple top level packages.

Address #571, #527, and makes a compromise for #530

# PR Checklist

### DIFF
--- a/nuitka/Tracing.py
+++ b/nuitka/Tracing.py
@@ -81,7 +81,7 @@ def my_print(*args, **kwargs):
 
         del kwargs["style"]
 
-        print(style, *(args + ("\033[0m",)), **kwargs)
+        print(*((style + args[0],) + args[1:] + ("\033[0m",)), **kwargs)
 
     else:
         print(*args, **kwargs)
--- a/nuitka/Tracing.py
+++ b/nuitka/Tracing.py
@@ -81,7 +81,9 @@ def my_print(*args, **kwargs):
 
         del kwargs["style"]
 
-        print(*((style + args[0],) + args[1:] + ("\033[0m",)), **kwargs)
+        print(style, end="")
+        print(*args, **kwargs)
+        print("\033[0m", end="")
 
     else:
         print(*args, **kwargs)
--- a/nuitka/distutils/bdist_nuitka.py
+++ b/nuitka/distutils/bdist_nuitka.py
@@ -27,6 +27,8 @@ import sys
 
 import wheel.bdist_wheel  # pylint: disable=I0021,import-error,no-name-in-module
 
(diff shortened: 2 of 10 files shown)


tommyli3318:
@kayhayen The current script passes all test cases in `distutils` (including the new one I added). I also ran some tests in `PyPI-pytest` and everything seems to work fine. Please let me know if my example test case is wrong.

nuitka/distutils/bdist_nuitka.py
```diff
@@ -42,6 +44,30 @@ def setuptools_build_hook(dist, keyword, value):
     dist.cmdclass["bdist_wheel"] = bdist_nuitka
 
 
+class py_package:
+    def __init__(self, module_name, related_packages=[]):
+        self.module_name = module_name  # string
+        self.related_packages = related_packages  # list
+
+    def __str__(self):
+        return "py_package(module_name=%s, related_packages=%s)" % (
```
kayhayen: Never repeat a class name in `__str__`, instead use `__class__.__name__`

nuitka/distutils/bdist_nuitka.py
```diff
@@ -77,78 +141,83 @@ def _build(self, build_lib):
...
+                    ):
+                        for val in value:
+                            command.append("%s=%s" % (option, val))
+                    else:
+                        command.append("%s=%s" % (option, value))
+
+            command.append(main_filename)
+
+            # added for clarity
+            my_print("Nuitka:INFO:Building: %s" % to_build, style="yellow")
```
kayhayen: Uh, looks like we are imitating the log package, not sure about that at all. Doesn't the color suffice?

nuitka/Tracing.py
```diff
@@ -81,7 +81,7 @@ def my_print(*args, **kwargs):
 
         del kwargs["style"]
 
-        print(style, *(args + ("\033[0m",)), **kwargs)
+        print(*((style + args[0],) + args[1:] + ("\033[0m",)), **kwargs)
```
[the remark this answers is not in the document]
kayhayen: Ah, so this is to address the leading spaces that I saw, and never understood.

You don't have to force yourself into doing everything in one print statement, also, image something does print something that is not a string, say an object, then `+` would fail.

You can just do `print(style, end="")` before the real print, and then `print("\033[0m", end="") ` afterwards, that is probably easier?

tommyli3318:
Updated, still passes all test cases in `distutils`

tommyli3318:
@kayhayen Can you review this some time :)

Issue #580, opened 2019-11-30
Error in checking if Python belongs to a conda environment


gioboske:
Python and Nuitka version:
```
nuitka: 0.6.5 to 0.6.6rc7 (both master and develop branch)
Python: 3.7.x (conda)
OS: Darwin
Arch: x86_64
```
To check if Python belongs to a conda environment, in general it is checked if "Anaconda" is included in `sys.version` string. From Python version 3.7 This lead to wrong results as explained in the comments of [this](https://stackoverflow.com/a/21282816/2809027) stackoverflow answer.
Instead of:
```
"Anaconda" in sys.version
```
the following should works for all supported Python version:
```
os.path.exists(os.path.join(sys.prefix, 'conda-meta'))
```

**This issue is probably related to #499.**
Maybe fixing it could lead to an easier fix of #499 (in place of PR #579).

I plan to report the outcome of some tests in the next few days.

gioboske:
Definitively related to #95.

kayhayen:
That's great, I was unaware of this. I saw your PR and will merge it. It conflates the static linking stuff, but like I said, that should be possible to decide.

We probably ought to make this detection result in a `conda_mode` towards `SingleExe.scons` code, and then just do static linking for these, but only if the Conda-CC is used. If asked to use any other, we could disable conda_mode, resulting in the performance loss.

My main concern with Conda-CC is that I was having erros with it on Windows, where gcc-5 is probably just a little big too old to not have some flukes. But this is not about Windows at all.

Thanks for your contributions, very much appreciated! I will look into this tomorrow I hope.

gioboske:
Closing the issue, since PR #581 is merged.
Since this issue was limited to conda env check for Python 3.6+, opening a new issue (#595) to address for the new plan to make static linking and using conda-cc mandatory on non-Windows (as outlined by @kayhayen [here](https://github.com/Nuitka/Nuitka/pull/581#issuecomment-569212442) in PR #581).

kayhayen:
The release 0.6.8 was just made and contains the correction.

Issue #499, opened 2019-08-22
"hello world" on macOS Mojave and miniconda - dyld: Library not loaded


gioboske:
Hello all,
first try with miniconda installation of Nuitka, with a simple hello-world script on macOS Mojave.
Here the key info:
- Python and Nuitka version:
```
nuitka: 0.6.5
Python: 3.7.4 (default, Aug 13 2019, 15:17:50) 
OS: Darwin
Arch: x86_64
```
- Tested both with Python3 and Nuitka installed from conda-forge and with Python3 from miniconda main and Nuitka installed from PyPI.

- Compiled with:
`python -m nuitka hello.py`
tried also with `--standalone`

- hello.py source from Nuitka documentation:
```
def talk(message):
    return "Talk " + message

def main():
    print(talk("Hello World"))

if __name__ == "__main__":
    main()
```

- No error at compile time (also checked with `--verbose` and `--show-scons`)

- Always get the following error at runtime:
```
dyld: Library not loaded: @rpath/libpython3.7m.dylib
  Referenced from: /Users/gioboske/dev/nuitka/hello.dist/./hello
  Reason: image not found
Abort trap: 6
```
Am I missing something obvious?
Thank you.

kayhayen:
There was a fix in 0.6.5 driven by Travis testing of macOS, with corrupted paths. Can you confirm if in hello.dist there is a file named `libpython3.7m.dylib`. I think there is somebody else working on a PR which fixes up handling of `@` keywords, but that was for specific Qt installations, but maybe hello world is affected with Python3.7 already.

But at least for MiniConda this could be a case of `nuitka.PostProcessing` failing to copy the shared library ner the executable.

Should it be missing, you can verify by not using standalone mode and placing the DLL next to the binary when executing it. If that works, I will know what to do then.

gioboske:
Hi Kay,
the library `libpython3.7m.dlyb` is **not included** in hello.dist when compiling with`--standalone` option (I paste below the output of `ls`, maybe it helps).

Unfortunately, the same error persists also without `--standalone` option, manual copying the library next to `hello.bin`.

Let me know if you need some additional info (I'm not a skilled developer, but I'll do my best!).

```
_asyncio.so		_curses.so		_pickle.so		binascii.so
_bisect.so		_curses_panel.so	_posixsubprocess.so	fcntl.so
_blake2.so		_datetime.so		_queue.so		grp.so
_bz2.so			_dbm.so			_random.so		hello
[...]

Issue #95, opened 2018-08-07
Static linking of Python library for programs not supported


kayhayen:
This hasn't worked so far.

Currently the issue seems to be that Py_GetArgcArgv collides with the symbols in a "libpythonXY.a" coming from "Modules/main.c".

Wo do not use that, and cannot use it. So the values would be wrong, while there are Python modules with semi wide usage, which rely on these functions to work correctly.

The idea therefore is that we can find a "libpython.a" that would be used, and turn it into a (cached) form with "main.o" removed from the archive.

What will be needed is to statically compile a CPython from source. Do not use --enable-shared doing that. Then examine the resulting "libpython.a" and try to remove "main.o" and see if linking to a program then works with Nuitka without any duplicate symbols errors. 

The way caching is done can be seen in "makeWindowsMingGW64Lib" which gets an appdirs provided directory "nuitka_cache" from the outside. So it would only do that once then.

This should be a simple task. I invite others to try it for themselves and create a PR to include in Nuitka.

niothiel:
FWIW: I was able to get this solution to work, as I am a user of `pyenv`. Adding more context for documentation purposes.

## Environment
pyenv version: 1.2.6
OSX version: 10.13.6
Python version: pyenv-installed 3.6.6
Nuitka version: 0.5.33rc4 (factory install as of `Tue Aug 14 19:38:20 EDT 2018`)
clang:
```
Apple LLVM version 9.1.0 (clang-902.0.39.1)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

## Steps taken
1. Install python statically via pyenv: `pyenv install 3.6.6`
1. Find all libpython references:
    ```
    $ cd $(pyenv root)/versions/3.6.6
    $ pwd
    /Users/xxxxxx/.pyenv/versions/3.6.6
    $ find . -name 'libpython*'
    ./lib/python3.6/config-3.6m-darwin/libpython3.6m.a
    ./lib/libpython3.6m.a
    ```
1. Remove `main.o`
  `ar -d lib/python3.6/config-3.6m-darwin/libpython3.6m.a main.o`
  **NOTE:** In my testing, only the config directory needs to have `main.o` removed, not the other one.
1. Try compiling a simple file: `python -m nuitka hello.py`
1. Execute file successfully:
    ```
    $ ./hello.exe
    Hello world!
    ```

## Results
[...]

Issue #516, opened 2019-09-20
Standalone includes hardcoded external modules meaning executable can't be shared


brendanator:
* Nuitka version, full Python version and Platform (Windows, OSX, Linux ...)

```
$ python -m nuitka --version
0.6.5
Python: 3.7.4 (v3.7.4:e09359112e, Jul  8 2019, 14:54:52)
Executable: /Users/brendan/.local/share/virtualenvs/test/bin/python
OS: Darwin
Arch: x86_64

$ uname -a
Darwin 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64
```

* How did you install Nuitka and Python (pip, anaconda, deb, rpm, from source,
  what is a virtualenv ...), this is very important usually.
    - Python from https://www.python.org
    - nuitka with pipenv

* If possible please supply a [Short, Self Contained, Correct, Example](http://sscce.org/)
  that demonstrates the issue i.e a small piece of code which reproduces
  the issue and can be run with out any other (or as few as possible)
  external dependencies.

```
$ cat main.py
import random

$ python -m nuitka --standalone main.py

$ ./main.dist/main

$ sudo mv /Library/Frameworks/Python.framework/Versions/3.7 /Library/Frameworks/Python.framework/Versions/3.6

$ ./main.dist/main
ERROR:root:code for hash md5 was not found.
Traceback (most recent call last):
  File "/tmp/example/main.dist/./hashlib.py", line 244, in <module>
  File "/tmp/example/main.dist/./hashlib.py", line 113, in __get_builtin_constructor
ValueError: unsupported hash type md5
ERROR:root:code for hash sha1 was not found.
Traceback (most recent call last):
  File "/tmp/example/main.dist/./hashlib.py", line 244, in <module>
  File "/tmp/example/main.dist/./hashlib.py", line 113, in __get_builtin_constructor
ValueError: unsupported hash type sha1
ERROR:root:code for hash sha224 was not found.
Traceback (most recent call last):
  File "/tmp/example/main.dist/./hashlib.py", line 244, in <module>
  File "/tmp/example/main.dist/./hashlib.py", line 113, in __get_builtin_constructor
ValueError: unsupported hash type sha224
ERROR:root:code for hash sha256 was not found.
Traceback (most recent call last):
  File "/tmp/example/main.dist/./hashlib.py", line 244, in <module>
[...]

PR #581, opened 2019-11-30
Update check for conda envs to support Python > 3.6 (issue #580)

gioboske:
# What does this PR do?
Update the check command for conda envs, to support Python > 3.6. Checking if "Anaconda" is in sys.version string, lead to wrong results for Python ver. > 3.6, as explained in [this](https://stackoverflow.com/a/21282816/2809027) stackoverflow answer.
This is also related to static linking of libpython (in fact only available for conda on non-windows OS).

Related to issue #580 and #95.
Maybe also related to issue #499.

# PR Checklist

- [x] ~~Ideally new or changed features have documentation updates.~~ Not needed

### DIFF
--- a/nuitka/Options.py
+++ b/nuitka/Options.py
@@ -361,7 +361,7 @@ def shallClearPythonPathEnvironment():
 
 
 def shallUseStaticLibPython():
-    """ *bool* = derived from sys.version
+    """ *bool* = derived from `sys.prefix` and `os.name`
 
     Notes:
         Currently only AnaConda on non-Windows can do this.
--- a/nuitka/Options.py
+++ b/nuitka/Options.py
@@ -369,7 +369,7 @@ def shallUseStaticLibPython():
 
     # For AnaConda default to trying static lib python library, which
     # normally is just not available or if it is even unusable.
-    return "Anaconda" in sys.version and os.name == "nt"
+    return os.path.exists(os.path.join(sys.prefix, 'conda-meta')) and not os.name == "nt"
 
 
 def shallTreatUninstalledPython():
--- a/nuitka/PythonVersions.py
+++ b/nuitka/PythonVersions.py
@@ -149,7 +149,7 @@ def isUninstalledPython():
         system_path = os.path.normcase(buf.value)
         return not getRunningPythonDLLPath().startswith(system_path)
 
-    return "Anaconda" in sys.version or "WinPython" in sys.version
+    return os.path.exists(os.path.join(sys.prefix, 'conda-meta')) or "WinPython" in sys.version
 
 
 def getRunningPythonDLLPath():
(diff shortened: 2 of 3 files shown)


nuitka/Options.py
```diff
@@ -354,15 +354,15 @@ def shallClearPythonPathEnvironment():
...
+    """ *bool* = derived from `sys.prefix` and `os.name`
 
     Notes:
         Currently only AnaConda on non-Windows can do this.
     """
 
     # For AnaConda default to trying static lib python library, which
     # normally is just not available or if it is even unusable.
-    return "Anaconda" in sys.version and os.name == "nt"
+    return os.path.exists(os.path.join(sys.prefix, 'conda-meta')) and not os.name == "nt"
```
kayhayen: Yes, that is clearly much better, however did a `not` snuggle in there in that condition? I assume, you want to do it not just for Windows, which is what this code did. I therefore think it might break Windows.

I was defensive about this, as static Python didn't work unless you used the precisely same gcc version than Anaconda did. I was e.g. unlucky with gcc-6 and gcc-8 on Linux, where it is compiled with gcc-7, one was saying too new bytecode, the other was crashing when using it to link with cryptic errors.

kayhayen:
This would require us to force users to only use `conda-cc` when using AnaConda.

Arguably that is an option. I think I need more thoughts on this. An Anaconda mode clearly could be considered, where it is added automatically to the PATH when compiling. I am assume you are working with it, with an env activated that sets CC.

Nuitka also works without that, but it's questionable if it should. Again, need some time to think through the implications. Getting errors from using the bytecode link library will be painful, but performance will be better.

nuitka/Options.py
gioboske: Maybe I missed something: as far as I understood, static linking is not feasible in Windows OS, so the `not os.name == "nt"` condition...

gioboske:
Totally agree with you: more thoughts on this are needed.
By the way, since static linking (and more in general Anaconda support and optimisations) are somehow related to #95, #499, #516, #580 and probably more, I plan to make a better issues' report for conda Python 3.6/3.7 on macOS/Linux.
Maybe I'll open a new issue, reporting different combination on conda Python version and platform (for now, macOS and Linux only), also testing `--module` and `--follow-imports` options (not only `--standalone` as previously done).
It will take me some time, since I've only basic knowledge of compilation, linking and Scons, but I'll try.
I've rebased this PR onto current `develop` and start testing.

Thank you @kayhayen!

nuitka/Options.py
kayhayen: You are right, I need to check how this came to be, the doc comment also states that. You just correct a mistake there, great. With this error, the feature was disabled. Let me take the time to check the history there before getting back to this.

kayhayen:
I am delaying this until 0.6.7 and will make 0.6.6 without it, I think we will indeed make static linking and using conda-cc mandatory on non-Windows. On Windows the old gcc is a pain, and doesn't do all the stuff as well as it should.

As this is a massive change, I would not want to put it on the path of the huge already 0.6.6, but I will deal with this right afer release, or so the plan. Nuitka then ought to be able to complain its way out of a minimal conda environment until it works and refuse to use other compilers without warning about it.

gioboske:
Rebased on current develop (0.6.7rc1).

Issue #589, opened 2019-12-16
Possibly broken--python-flag=-O flag


deajan:
* Nuitka version, full Python version and Platform (Windows, OSX, Linux ...)

Running on Python 3.7.4 x64 on Windows 10 1903.

  python -m nuitka --version
```
0.6.5
Python: 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 20:34:20) [MSC v.1916 64 bit (AMD64)]
Executable: c:\Python37-64\python.exe
OS: Windows
Arch: x86_64

```
* How did you install Nuitka and Python (pip, anaconda, deb, rpm, from source,
  what is a virtualenv ...), this is very important usually.

Generic `python -m pip install nuitka`

Hello Kay, I noticed from earlier (docs)[https://nuitka.net/posts/nuitka-release-0525.html] that the `--python-flag=-O` should remove docstrings from compilation.

I've tried a couple of times with standard, standalone to compile and than seach for a docstring in the resulting executable.

Here's my minimal code:
```
def fn():
    """
    This is the string I am finding in the nuitka executable
    """
    print('Hello, World !')

fn()
```

Compiled with MingW via `c:\Python37-64\python.exe -m nuitka --standalone --python-flag="-O" docstring.py`

When I use the strings.exe binary that comes with mingw, I can find the docstring:
(btw, using busybox grep because of... well windows ;))

`
C:\GIT\EvilWorldDominationProject\PythonSRC\test_programs\docstring.dist>c:\MINGW\mingw64\bin\strings.exe docstring.exe | c:\tools\busybox.exe grep "This is the string"
`
`
This is the string I am finding in the nuitka executable
`

Just wanted to report that in hope it's not a regression.
Kind regards.

kamgha:
This was a bug that got fixed for Nuitka 0.6.1.1. https://nuitka.net/posts/nuitka-this-week-12.html Please try `--python-flag=-OO`. According to the CPython 3.7 documentation, `-O` will not remove docstrings, but `-OO` does.

kayhayen:
@kamgha spot on, thanks!

Do you think, we need a better help output that mentions all flags supported. I think "-OO" is missing right now, and other flags too.

```
                if part in ("-S", "nosite", "no_site"):
                    _python_flags.add("no_site")
                elif part in ("static_hashes", "norandomization", "no_randomization"):
                    _python_flags.add("no_randomization")
[...]

PR #591, opened 2019-12-20
Fix --python-flags=no_docstrings without implying no_asserts

kamgha:
Derived from issue #589.


kayhayen:
Perfect, thanks for your contribution, that makes a lot of sense to discard this misleading aliases and make them do what they say only

PR #622, opened 2020-02-11
Fix, mac ctypes.util does not have _get_soname().

prantaaho:
# What does this PR do?

On macOS Python ctypes.util does not contain _get_somename() protected method: https://github.com/python/cpython/blob/3.7/Lib/ctypes/util.py#L70 Due to this Nuitka SharedLibraries:locateDLL fails on macOS.

I stumbled upon this when I was trying to create standalone build from QtQuick Hello World example. 

# PR Checklist

### DIFF
--- a/nuitka/utils/SharedLibraries.py
+++ b/nuitka/utils/SharedLibraries.py
@@ -41,6 +41,7 @@ def localDLLFromFilesystem(name, paths):
 
 
 def locateDLL(dll_name):
+    # This function is a case driven by returns, pylint: disable=too-many-return-statements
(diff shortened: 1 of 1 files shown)


prantaaho:
So the PyLint doesn't like it: ``R0911 too-many-return-statements locateDLL Too many return statements``. Any preferences how to refactor it?

kayhayen:
In other cases, I have used this:

`    # This function is a case driven by returns, pylint: disable=too-many-return-statements`

In that case, the function must return only, not assign a return value, and then return.

Issue #612, opened 2020-01-29
bdist_nuitka produces wrong output for a nested namespace package


ra-kete:
* Nuitka version, full Python version and Platform (Windows, OSX, Linux ...)

```
$ python3 -m nuitka --version
0.6.7
Python: 3.6.8 (default, Aug 20 2019, 17:12:48)
Executable: /usr/bin/python3
OS: Linux
Arch: x86_64
```

* How did you install Nuitka and Python (pip, anaconda, deb, rpm, from source,
  what is a virtualenv ...), this is very important usually.

```
$ python3 -m pip install nuitka
```

# Issue

I am trying to compile a Python package of mine to an extension module and package it into a wheel using Nuitka. The special thing about my case seems to be that this package resides inside an implicit namespace package nested two levels deep. The compile runs fine, the `.so` file gets created too, but when I install that wheel, my package cannot be imported.

## Minimal example

Consider the following package layout:

```
|- a/b/pkg/__init__.py
|- setup.py
```

`a.b.pkg` is the package I want to compile to an extension module. `a.b` is the implicit namespace package (notice the lack of `__init__.py`s).

These are the contents of `setup.py`:

```python
from setuptools import setup

setup(
    name="pkg",
    packages=["a.b.pkg"],
)
```

I now invoke `bdist_nuitka`:

```
$ python3 setup.py bdist_nuitka
running bdist_nuitka
running build
running build_py
creating build
creating build/lib
creating build/lib/a
creating build/lib/a/b
creating build/lib/a/b/pkg
copying a/b/pkg/__init__.py -> build/lib/a/b/pkg
Building: PyPackage(module_name=a.b.pkg, related_packages=['a.b.pkg'])
Nuitka:INFO:Recurse to import 'a.b.pkg' from 'a/b/pkg'. (Module is exact match of 'a.b.pkg' instructed by user to recurse to.)
Nuitka:INFO:Recurse to import 'a.b' from 'a/b'. (Containing package of recursed module 'a.b.pkg'.)
Nuitka:INFO:Recurse to import 'a' from 'a'. (Containing package of recursed module 'a.b'.)
Nuitka:INFO:Included compiled module 'a'.
Nuitka:INFO:Included compiled module 'a.b'.
Nuitka:INFO:Included compiled module 'a.b.pkg'.
Nuitka:INFO:Included compiled module 'pkg'.
installing to build/bdist.linux-x86_64/wheel
running install
running install_lib
[...]

PR #646, opened 2020-03-10
Distutils nested namespaces

ra-kete:
# What does this PR do?

Fixes support of the build_nuitka distutils command for implicit namespace packages nested more than one level deep.

This bug was reported in  #612.

### DIFF
--- a/nuitka/distutils/DistutilCommands.py
+++ b/nuitka/distutils/DistutilCommands.py
@@ -24,6 +24,7 @@ import distutils.command.install  # pylint: disable=I0021,import-error,no-name-i
 import os
 import subprocess
 import sys
+from distutils import dir_util
 
 import wheel.bdist_wheel  # pylint: disable=I0021,import-error,no-name-in-module
 
@@ -230,6 +231,18 @@ class build(distutils.command.build.build):
 
                     if fullpath.lower().endswith((".py", ".pyw", ".pyc", ".pyo")):
                         os.unlink(fullpath)
+
+            # If the Python module has more than one parent package (e.g.
+            # 'a.b.mod'), the compiled module will be in 'a.b/mod.so'. Move it
+            # to 'a/b/mod.so', to make imports work.
+            if package and "." in package:
+                compiled_package_path = os.path.join(build_lib, package)
+                assert os.path.isdir(compiled_package_path), compiled_package_path
+
+                parts = package.split(".")
+                fixed_package_path = os.path.join(build_lib, *parts)
+                dir_util.copy_tree(compiled_package_path, fixed_package_path)
+                dir_util.remove_tree(compiled_package_path)
 
             os.chdir(old_dir)
 
--- a/nuitka/distutils/DistutilCommands.py
+++ b/nuitka/distutils/DistutilCommands.py
@@ -24,11 +24,11 @@ import distutils.command.install  # pylint: disable=I0021,import-error,no-name-i
 import os
 import subprocess
 import sys
-from distutils import dir_util
 
 import wheel.bdist_wheel  # pylint: disable=I0021,import-error,no-name-in-module
 
 from nuitka.tools.testing.Common import my_print
+from nuitka.utils.FileOperations import copyTree, removeDirectory
 
 
 def setupNuitkaDistutilsCommands(dist, keyword, value):
@@ -241,8 +241,8 @@ class build(distutils.command.build.build):
 
                 parts = package.split(".")
                 fixed_package_path = os.path.join(build_lib, *parts)
-                dir_util.copy_tree(compiled_package_path, fixed_package_path)
-                dir_util.remove_tree(compiled_package_path)
+                copyTree(compiled_package_path, fixed_package_path)
+                removeDirectory(compiled_package_path, ignore_errors=False)
 
             os.chdir(old_dir)
 
--- a/tests/distutils/run_all.py
+++ b/tests/distutils/run_all.py
@@ -57,7 +57,7 @@ from nuitka.utils.FileOperations import removeDirectory
 
 
 def main():
-    # Complex stuff, pylint: disable=too-many-branches,too-many-locals,too-many-statements
+    # Complex stuff, pylint: disable=too-many-locals,too-many-statements
 
     python_version = setup(needs_io_encoding=True)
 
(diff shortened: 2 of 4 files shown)


nuitka/distutils/DistutilCommands.py
```diff
@@ -231,6 +232,18 @@ def _build(self, build_lib):
...
+            # If the Python module has more than one parent package (e.g.
+            # 'a.b.mod'), the compiled module will be in 'a.b/mod.so'. Move it
+            # to 'a/b/mod.so', to make imports work.
+            if package and "." in package:
+                compiled_package_path = os.path.join(build_lib, package)
+                assert os.path.isdir(compiled_package_path), compiled_package_path
+
+                parts = package.split(".")
+                fixed_package_path = os.path.join(build_lib, *parts)
+                dir_util.copy_tree(compiled_package_path, fixed_package_path)
```
kayhayen: We have `nuitka.utils.FileOperation.copyTree` for that. Arguably importing distutils cannot be all that wrong in this module, but still.

nuitka/distutils/DistutilCommands.py
```diff
@@ -231,6 +232,18 @@ def _build(self, build_lib):
...
+                dir_util.remove_tree(compiled_package_path)
```
kayhayen: Is this not really a move then?

tests/distutils/run_all.py
```diff
@@ -83,10 +83,12 @@ def main():
...
             my_print("Consider distutils example:", filename)
 
-            if python_version < "3":
-                if filename == "example_3":
-                    reportSkip("Skipped, only relevant for Python3", ".", filename)
-                    continue
+            if python_version < "3" and filename == "example_3":
+                reportSkip("Skipped, only relevant for Python3", ".", filename)
+                continue
+            if python_version < "3.3" and filename == "nested_namespaces":
```
kayhayen: Just a minor quibble, we don't support "3.2" anymore, and therefore "<3" and "<3.3" are really the same thing to us, so you can spare the distinction.

kayhayen:
The PyLint issue would not be raised, if you used the autoformat from the commit hook, or execute it manually, it's called `./bin/autoformat-nuitka-source`. It does isort and black, and more, and avoids these nasty pylint stuff about imports entirely, or so I think.

nuitka/distutils/DistutilCommands.py
ra-kete: Yes, it is a move. This was just the most succinct way to write this. `shutil.move` doesn't cut it, since the details are a bit unusual:

- I want to move all contents of `compiled_package_path` into `fixed_package_path`
- I want to create `fixed_package_path`, if it doesn't already exist

Is there something in `nuitka.utils.FileOperation` that does this?

tests/distutils/run_all.py
ra-kete: Ah, good to know. I will change this.

ra-kete:
I tried running the commit hook, but for some reason it also introduced unrelated changes in `nuitka/distutils/DistutilCommands.py` and I didn't want to touch more code than necessary, since this is only a small bugfix.

nuitka/distutils/DistutilCommands.py
ra-kete: I moved to `FileOperation` instead of `dir_util`, but there isn't a function there to do a "move all the contents of this directory into this directory" either, so it's still a copy + remove.

kayhayen:
Thanks for doing it. What sometimes happens, but not now, or at least so I am thinking, is that I merge things, which are not subject to autoformat (yet) and then other changes happen, and only later this happens, but I think this is not the case here. At least it looks like it (on current develop).

> 
> ./bin/autoformat-nuitka-source nuitka/distutils/
> Working on: ['nuitka/distutils/']
> Consider nuitka/distutils/DistutilCommands.py: OK.
> Consider nuitka/distutils/__init__.py: OK. 

So those unrelated changes are probably just what black does, or it would be a bug in how isort works, which I think can have different ideas about 3rd party if not everything is installed. Anyway, if you have something to share on that front, I would appreciate it, but it won't matter too much.

ra-kete:
There were two instances where black complained about too long lines in `DistutilCommands.py`, but looks like they're fixed now. And isort seemed unable to figure out that `nuitka` is a first-party package, it tried to group it with `wheel` imports. Couldn't figure out how to resolve this, so I still had to commit with `--no-verify` in the end. Maybe just add `nuitka` to isort's `known_first_party` list?

kayhayen:
That's sad @ra-kete please help me figure that out. Can you try the current factory branch, running `./bin/autoformat-nuitka-source` and let me know if it's better?

On thing that also crossed my mind, is that the version of `isort` and `black` might not be the ones specified on the `requirements-devel.txt`. Can you confirm that? It might be worth adding version checks to the tool to discover this automatically.

ra-kete:
Sure. Here is what I did:

- Do a clean clone into a new directory
- Create a venv: `python3 -m venv .venv; source .venc/bin/activate.fish`
- Install dev requirements: `pip install -r requirements-devel.txt`
- Checkout `factory` branch and run `./bin/autoformat-nuitka-source`

Here is the git diff after autoformatting:

```diff
diff --git a/nuitka/distutils/DistutilCommands.py b/nuitka/distutils/DistutilCommands.py
index 59b0af2d..5ef28029 100644
--- a/nuitka/distutils/DistutilCommands.py
+++ b/nuitka/distutils/DistutilCommands.py
@@ -26,7 +26,6 @@ import subprocess
 import sys

 import wheel.bdist_wheel  # pylint: disable=I0021,import-error,no-name-in-module
-
 from nuitka.tools.testing.Common import my_print
 from nuitka.utils.FileOperations import copyTree, removeDirectory

diff --git a/nuitka/tools/testing/run_nuitka_tests/__main__.py b/nuitka/tools/testing/run_nuitka_tests/__main__.py
index 3e284434..2d56402e 100755
--- a/nuitka/tools/testing/run_nuitka_tests/__main__.py
+++ b/nuitka/tools/testing/run_nuitka_tests/__main__.py
@@ -95,7 +95,6 @@ The plugins tests, execute these to check if Nuitka handles its own plugin
 interfaces, e.g. user plugins, etc. fine. Default is %default.""",
     )

-
     parser.add_option(
         "--skip-optimizations-tests",
         action="store_false",
diff --git a/nuitka/tree/ReformulationContractionExpressions.py b/nuitka/tree/ReformulationContractionExpressions.py
index 1575eaf5..28c6a936 100644
--- a/nuitka/tree/ReformulationContractionExpressions.py
+++ b/nuitka/tree/ReformulationContractionExpressions.py
@@ -244,7 +244,10 @@ def buildGeneratorExpressionNode(provider, node, source_ref):

     # Some of the newly allowed stuff in 3.7 fails to set the async flag.
     if not is_async and python_version >= 370:
-        is_async = detectFunctionBodyKind(nodes=node.generators)[0] in ("Asyncgen", "Coroutine")
+        is_async = detectFunctionBodyKind(nodes=node.generators)[0] in (
+            "Asyncgen",
+            "Coroutine",
+        )

     if is_async:
         code_body = ExpressionAsyncgenObjectBody(
```

I thought it might be the versions of `isort` and `black` too, but since they are pinned in the `requirements-devel.txt` that can't be it, right?

PR #656, opened 2020-03-22
Add Python 3.8 to Travis

gliptak:


gliptak:
@kayhayen please review

kayhayen:
Sorry for the delay. I am mostly switching to Github actions recently, so I wasn't so sure about this, but is clear cut and definitely useful.

Issue #662, opened 2020-04-05
binary_directory in getBinaryDirectoryHostEncoded() lacks an implementation for OpenBSD


juanfra684:
Hi, I'm working on an OpenBSD package for Nuitka. It builds fine but the standalone binaries generated abort in `getBinaryDirectoryWideChars`. Looks like ` binary_directory` in `getBinaryDirectoryHostEncoded()` (`build/static_src/CompiledCodeHelpers.c`) lacks an implementation for OpenBSD.

```plain
$ python3.7 -m nuitka --version
0.6.7
Python: 3.7.7 (default, Mar 30 2020, 15:25:59) 
Executable: /usr/local/bin/python3.7
OS: OpenBSD
Arch: amd64
```

kayhayen:
Hello,

can you check if the factory branch works better for you. I do not have access to an openBSD machine, but this should be better and I willing to work on it.

https://nuitka.net/doc/factory.html

Yours,
Kay

juanfra684:
Thanks for the quick reply. I checked your recent changes in `CompiledCodeHelpers.c`, unfortunately that's not going to work on OpenBSD. It doesn't have `KERN_PROC_PATHNAME`.

juanfra684:
I asked in our internal chat and Robert Nagy gave me this code. http://ix.io/2gZF

kayhayen:
Excuse me for arguing with the demo program. 

I am not convinced that `argv[0] is the path to the binary in all cases, it depends on how it's forked and could be an arbitrary value. Therefore the Chromium patch is very much based on that kind of luck to work. My guess that `KERN_PROC_PATHNAME` is apparently untrue.

Also tokens need not be an array, only a `char *` is return by `strtok` and only element 0 is used. The split to `:` is confusing to me, maybe uses an implementation detail. Makes me wonder of course, what happens with paths that contain that character.

Also the scan for open files, seems to only do that the split to `:` value is returned only if a main program file is found. I guess, it's missing a match there. It would make sense to check if it's the same file I guess, but that part is not done. A program not having the program text open is probably unlikely, so this doesn't contribute to reliability, does it, and requiring it is not a help I think, if it's not used to check. So I would leave that out, or see it enhanced to make a check that compares the name or content derived from `argv[0]` and that file, and error out if they mismatch.

I don't think there is a way to get the filename names from a file handle, is there?

[...]

PR #665, opened 2020-04-06
Add OpenBSD to clang and posix platforms.

juanfra684:
# What does this PR do?

Clang is the default compiler on the most of platforms supported by OpenBSD. It also has gcc in the base system but the version is too old for the Nuitka requirements and will be deleted in the future.

The changes are needed to support OpenBSD and for the issue #662.

# PR Checklist

### DIFF
--- a/nuitka/build/SingleExe.scons
+++ b/nuitka/build/SingleExe.scons
@@ -157,6 +157,8 @@ noelf_mode = getBoolOption("noelf_mode", False)
 
 macosx_target = sys.platform == "darwin"
 
+openbsd_target = sys.platform.startswith("openbsd")
+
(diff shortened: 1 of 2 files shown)


kayhayen:
Thanks, I take it, this is not enough for standalone mode, but acceleration works with it now?

juanfra684:
I think so but I'm not 100% sure. I ran `'$PYTHON ./tests/run-tests --no-other-python --skip-reflection-test --skip-all-cpython-tests'` until it crashed after a lot of time, which I guess is due the issue #662.

kayhayen:
Yes, that would be standalone mode.

Issue #674, opened 2020-04-15
Suggestion regarding `nuitka/tools/release/Documentation.py`


oz123:
Hi, 
While working on packaging nuitka for gentoo, I noticed that the [Documentation.py](https://github.com/Nuitka/Nuitka/blob/8b97b891385b5d07c466ce7d0f0c8d71987f6202/nuitka/tools/release/Documentation.py) script contains
many of the following  `assert` ...`os.system` blocks:

```
assert (
            os.system(
                "convert -resize 32x32 doc/Logo/Nuitka-Logo-Symbol.svg ../nikola-site/files/favicon.ico"
            )
            == 0
        )
```

Those can be replaced with `subprocess.check_call` which raises an exception if the process exists with a code other than 0.

This change can reduce many lines of code.
Before I rush to create a patch, I was wondering if you are interested in merging such a change.

Thank you for your work on Nuitka.

kayhayen:
Hello, this is very welcome proposal. In the tooling, I have been lazy, and not adopted subprocess, but remained with this very low skill code.

Not to mention, that asserts, since they can be disabled, must never execute any actual stuff....

PR #678, opened 2020-04-18
Remove assertion of exist status 0 in favor of check_call

oz123:
`subprocess.check_call` will raise an exception if a command fails.
This shortens the code in this module significantly and delegated
the work to subprocess.check_call.

As discusses in #674.

### DIFF
--- a/nuitka/tools/release/Documentation.py
+++ b/nuitka/tools/release/Documentation.py
@@ -36,23 +36,30 @@ def optimize_pngs(pngList):
 def makeLogoImages():
     basePathLogo = "doc/Logo/Nuitka-Logo-%s"
 
-    for logo in ("Vertical", "Symbol", "Horizontal"):
-        cmd = "convert -background none %s.svg %s.png" % (basePathLogo, basePathLogo)
-        subprocess.check_call((cmd % (logo, logo)).split())
+    try:
+        for logo in ("Vertical", "Symbol", "Horizontal"):
+            cmd = "convert -background none %s.svg %s.png" % (basePathLogo,
+                                                              basePathLogo)
+            subprocess.check_call((cmd % (logo, logo)).split())
+    except OSError:
+        sys.exit("Could not execute convert. Is it installed?")
 
-    optimize_pngs([basePathLogo % item for item in
-                  ("Vertical", "Symbol", "Horizontal")])
+    try:
+        optimize_pngs([basePathLogo % item for item in
+                      ("Vertical", "Symbol", "Horizontal")])
+    except OSError:
+        sys.exit("Could not execute optipng. Is it installed?")
 
     if os.path.exists("../nikola-site"):
         cmd = "convert -resize %s doc/Logo/Nuitka-Logo-Symbol.svg %s"
-        for icon, size in {"../nikola-site/files/favicon.ico": "32x32",
-                           "../nikola-site/files/favicon.png": "32x32",
-                           "../nikola-site/files/apple-touch-icon-ipad.png": "72x72",
-                           "../nikola-site/files/apple-touch-icon-ipad3.png": "144x144",
-                           "../nikola-site/files/apple-touch-icon-iphone.png": "57x57",
-                           "../nikola-site/files/apple-touch-icon-iphone4.png": "114x114",
-
-                           }:
+        for icon, size in {
+                "../nikola-site/files/favicon.ico": "32x32",
+                "../nikola-site/files/favicon.png": "32x32",
+                "../nikola-site/files/apple-touch-icon-ipad.png": "72x72",
+                "../nikola-site/files/apple-touch-icon-ipad3.png": "144x144",
+                "../nikola-site/files/apple-touch-icon-iphone.png": "57x57",
+                "../nikola-site/files/apple-touch-icon-iphone4.png": "114x114",
+                }:
             subprocess.check_call((cmd % (icon, size)).split())
 
 
@@ -82,15 +89,17 @@ def makeManpages():
         os.mkdir("man")
 
     def makeManpage(python, suffix):
-        cmd =  ["help2man", "-n", "'the Python compiler'", "--no-discard-stderr",
-                "--no-info", "--include doc/nuitka-man-include.txt",
-                "'%s ./bin/nuitka'" % python, ">doc/nuitka%s.1" % suffix]
+        cmd = ["help2man", "-n", "'the Python compiler'",
+               "--no-discard-stderr",
+               "--no-info", "--include doc/nuitka-man-include.txt",
+               "'%s ./bin/nuitka'" % python, ">doc/nuitka%s.1" % suffix]
         subprocess.check_call(cmd)
 
         cmd[-1] = ">doc/nuitka%s-run.1" % suffix
         subprocess.check_call(cmd)
 
-        for manpage in ("doc/nuitka%s.1" % suffix, "doc/nuitka%s-run.1" % suffix):
+        for manpage in ("doc/nuitka%s.1" % suffix,
+                        "doc/nuitka%s-run.1" % suffix):
             with open(manpage) as f:
                 manpage_contents = f.readlines()
             new_contents = []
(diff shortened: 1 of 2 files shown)


nuitka/tools/release/Documentation.py
```diff
@@ -28,68 +28,65 @@
...
+    """
+    cross version program detection.
+
+    Whenever Nuitka decides to drop Python versions older than 3.3
+    the code in the else block can be removed.
+    """
+    def real(f):
+        def wrap(*args):
+            cmd = ""
+            if sys.version_info > (3, 3):
```
kayhayen, line 41: That is probably not a good way to compare. You probably mean `>=` here.

nuitka/tools/release/Documentation.py
```diff
@@ -28,68 +28,65 @@
 from nuitka.utils.FileOperations import getFileList
 
 
+def which(prog):
```
kayhayen, line 31: We already have `nuitka.utils.Execution.getExecutablePath` to which some of these changes could be considered an improvement for sure.

nuitka/tools/release/Documentation.py
```diff
@@ -28,68 +28,65 @@
...
+    """
+    cross version program detection.
+
+    Whenever Nuitka decides to drop Python versions older than 3.3
```
kayhayen, line 35: For Nuitka this practically Python2 then, we don't support 3.2 anymore, and 3.1 and 3.0 were never considered.

nuitka/tools/release/Documentation.py
```diff
@@ -28,68 +28,65 @@
...
+    the code in the else block can be removed.
+    """
+    def real(f):
+        def wrap(*args):
+            cmd = ""
+            if sys.version_info > (3, 3):
+                import shutil
+                cmd = shutil.which(prog)
+            else:
+                _, cmdname = os.path.split(prog)
```
kayhayen, line 45: This looks like `os.path.basename`, but I don't know why accept paths as input, seems unused.

nuitka/tools/release/Documentation.py
```diff
@@ -28,68 +28,65 @@
...
+
+                for path in os.environ["PATH"].split(os.pathsep):
+                    cmd_ = os.path.join(path, cmdname)
+                    if os.path.isfile(cmd_) and os.access(cmd_, os.X_OK):
+                        cmd = cmd_
+            if cmd:
+                f(*args)
+            else:
+                print("Could not find such program %s" % prog)
```
kayhayen, line 54: That is better as `sys.exit` with a string argument. Errors in stdout are not good.

nuitka/tools/release/Documentation.py
```diff
@@ -28,68 +28,65 @@
...
+                sys.exit(1)
+        return wrap
+    return real
+
+
+@which('optipng')
+def optimize_pngs(pngList):
+    cmd = "optipng -o2 %s.png"
```
kayhayen, line 62: Why not use this form of calling:

```
    subprocess.check_call(
        [
            "git",
            "update-index",
            "--cacheinfo",
            "%s,%s,%s"
            % (diff_entry["dst_mode"], new_object_hash, diff_entry["src_path"]),
        ]
    )
```

The shell need not be used to locate the program, the first argument is searched in `PATH` automatically I think. We only use the function to locate binaries, only for other purposes, not to execute it. Also, this means your quoting is perfect, because it's not needed. The shell will do funny things with spaces, etc. never a good idea to use it, unless you are lazy and want e.g. shell patterns.

nuitka/tools/release/Documentation.py
```diff
@@ -167,23 +160,20 @@ def makeManpage(python, suffix):
...
-                    else r"C:\Python27_32\Scripts\rst2pdf.exe"
-                ),
-                "args": " ".join(args),
-                "document": document,
-            },
-            shell=True,
-        )
-        == 0
-    ), document
+    subprocess.check_call(
```
kayhayen, line 165: I guess, I was lazy here. But obviously, I should simply demand that rst2pdf be in the path when running the tool, and not make up of it like that.

nuitka/tools/release/Documentation.py
```diff
@@ -113,33 +110,29 @@ def checkRstLint(document):
     print("OK.")
 
 
+@which("help2man")
```
kayhayen, line 113: So is all this is doing to give a nicer error message of the first argument cannot be resolved? Not sure if that's worth it, a wrapper in `nuitka.utils.Execution` could use `try`/`except` on the process call error and output something nicer.

nuitka/tools/release/Documentation.py
oz123: I'm not sure what you mean here. Would you like me to use `nuitka.utils.Execution.getExecutablePath` here? Or change `nuitka.utils.Execution.getExecutablePath` ?

nuitka/tools/release/Documentation.py
oz123: That's correct. I just coppied old code of mine which was never reviewed. Thanks for that.

nuitka/tools/release/Documentation.py
oz123: Correct. I changed. this.

nuitka/tools/release/Documentation.py
oz123: I think this code is mostly used by packagers this - a clear error is quite a help. If this code was used by others codes, than try except is better.
You decide. If you want I can remove this decorator completely.

oz123:
> Sorry for the many comments, I hope it's OK.

Hi, That's no problem. Thanks for the through review.

nuitka/tools/release/Documentation.py
oz123: Yes, that can be simplified. I changed this call too.

nuitka/tools/release/Documentation.py
oz123: I can change this to the list format if you prefer.

nuitka/tools/release/Documentation.py
oz123: Yes, You are correct here.

nuitka/tools/release/Documentation.py
kayhayen: Use it, merging the enhancements, e.g. to use `shutil.which` where possible. But clearly not to duplicate the kind of code.

nuitka/tools/release/Documentation.py
kayhayen: I think, having a wrapper of `subprocess.checked_call` that catches the error in case of program not found, would be good enough. 

```
try: subprocess.checked_call(args)
except subprocess.SomeExceptionWeNeedToLoopUp as e:
   if e.something:
       sys.exit("Program args[0] not found!")
   else:
      raise
```

```
if os.path.sep not in args[0]:
    args=list(args)
    prog = getExecutablePath(args[0))
    if prog is None:
       sys.exit("Program not found!")
```

I think the later is better for extension, because I have places, where e.g. some tools are called, with `python -m prog` in preference to `prog` and I could merge that there. Also I like checks better than exception catches, even if Python is good at those.

I agree the Python call stack is barely useful to normal people, and having a proper system exit that explains an installation problem would be very nice. There is something about these tracebacks that turns off people's ability to continue reading. :-)

nuitka/tools/release/Documentation.py
oz123: You mean something like this?

```
def getExecutablePath(filename):
    """ Find an execute in PATH environment.
    """

    # Append ".exe" suffix  on Windows if not already present.
    if getOS() == "Windows" and not filename.lower().endswith(".exe"):
        filename += ".exe"

    if sys.version_info >= (3, 3):
        import shutil
        return shutil.which(filename)
    else:
        # Search in PATH environment.
        search_path = os.environ.get("PATH", "")

        # Now check in each path element, much like the shell will.
        path_elements = search_path.split(os.pathsep)

        for path_element in path_elements:
            path_element = path_element.strip('"')

            full = os.path.join(path_element, filename)

            if os.path.exists(full):
                return full
```

nuitka/tools/release/Documentation.py
oz123: Honestly, I don't know where you would put this wrapper for `check_call` ... I see there is already
a `check_output` for compat with Python 2.6. Would like to me to add those changes there?
Do you really still want to support Python2.6? It's been dead for quite a while...

kayhayen: Just to respond to this, Python 2.6 is used on enterprise systems still. For Nuitka it's not a big deal, because 2.7 is really only a 2.6 with the then best Python3 features backported to ease the migration, e.g. dict and set contractions, but nothing really important.

To Nuitka, Python is translated into a simpler language right after ast parsing, and the non-availability of higher language constructs doesn't make a problem there.

In terms of support effort, it's minimal, but test coverage is nice to have. I have not spent time on Python 2.6 for ages. Should I remove the support, very little code could go away.

As for making the changes there, you can go ahead. I am satisfied with your PR as it is though, this maybe is something I should do, as it affects many tools you don't care about. Let me know, if you want to persue this further. Otherwise I can pick up these ideas and use them.

oz123: Thanks for the reply. I think my latest push has all your suggestions. It's just a bit hard to really know how you would implement this. I would not mind if you took a look in the latest chnages and say if you are satisfied or not.

kayhayen: Thanks for the good job.

PR #717, opened 2020-05-20
fix UnicodeDecodeError

byehack:


kayhayen:
So this was good for documenting only, thanks for it. But the more involved fix is has come to factory by myself.

PR #764, opened 2020-06-25
Fixed bad diagnostic error, if user use old plugin option format like --enable-plugin=numpy=scipy

belonesox:
# What does this PR do?
Fixed bad diagnostic message — if user used old plugin options format, like 
«--enable-plugin=numpy=scipy» it gets error message
«Error, plugin options format changed. Use '--enable-plugin=numpy=scipy --help», and catched in bad recursion.


kayhayen:
Spurios errors in workflow for you too, only had that for factory so far.

kayhayen:
I might make this part of the next hotfix still, as the warning is not as helpful as it was intended to be when introduced in another hotfix. In that case, I would rebase develop with it.

PR #790, opened 2020-07-14
escape percent signs in TimerReport message

pipermerriam:

# What does this PR do?

When compiling a large application the scons step consistently took longer than 60 seconds.  This resulted in attempts to print out warning messages about the time the command took to run.

The construction of the `message` argument for the `TimerReport` class involved dumping some very large CLI commands.  In my case, one of the values contained a `%s`.  This resulted in the `TimerReport` throwing an exception when it attempts to construct the final formatted logger message.

To fix this, I first applied `repr(...)` to the argument that was being formatted into the message template, then did a `.replace('%', '%%')` to replace any percent signs that happen to be in the message with double percents in the final message template used to output the timing information.

# PR Checklist

      There are Travis and Github Actions tests that cover the most important
      cover enough.


kayhayen:
PR's against factory are moving targets. I will have to cherry-pick your commit.

As for the need of it, I guess, it's a sign of a broken interface there. But for now, this is an important fixup. I think on case insensitive platforms, similar names are broken with `%1`, `%2` .... suffixes, so this was very unlikely to be found by me. Thanks a lot for reporting and fixing it, that's going to spare me a lot of confusion later on.

kayhayen:
I am closing this, as I have cherry-picked your commit into factory.

PR #804, opened 2020-08-08
Update PySidePyQtPlugin.py

DerekRein:
added missing plugin family sqldrivers


kayhayen:
There is a command line options for the plugin to specify the plugins to include. The MySQL drivers are clearly not used by every Qt application, I dare say only a big minority, so they won't be added.

Maybe the plugin should be more talkative about what plugins it includes, and which ones not. But adding all to be included into "sensible" makes no sense, does it?

kayhayen:
Also, develop had issues from the latest pre-release that I had to rebase for, sorry about that, that's breaking your tests.

kayhayen:
Closing this, as it seems useless and I didn't receive any feedback on my last comments.

Issue #849, opened 2020-10-05
Hinted compilation with nuitka 0.6.9


cvpython:
I have installed nuitka 0.6.9.1 and use nuitka-utilities from the master branch.
When I try to compile with the hinted approach (`python get-hints.py script.py`) I have the following error:
```
Traceback (most recent call last):
  File "get-hints.py", line 43, in <module>
    from nuitka.utils.Utils import getSharedLibrarySuffix
ImportError: cannot import name 'getSharedLibrarySuffix' from 'nuitka.utils.Utils'
```

Are there nutika-utilities compatible with nuitka 0.6.9.1 ?

Here are some characteristics of my installation:
> 
> 0.6.9.1
> Python: 3.7.6 (default, Jan  8 2020, 19:59:22) 
> OS: Linux
> Arch: x86_64

danielmorgan0x1:
I am also using the latest version of Nuitka and had the same issue. I found the Utils.py file from 0.6.8 and found the getSharedLibrarySuffix function. I then added this to my Utils.py in 0.6.9.1 as follows:

`def getSharedLibrarySuffix():
    if python_version < 300:
        import imp

        result = None

        for suffix, _mode, module_type in imp.get_suffixes():
            if module_type != imp.C_EXTENSION:
                continue

            if result is None or len(suffix) < len(result):
                result = suffix

        return result
    else:
        import importlib.machinery  # pylint: disable=I0021,import-error,no-name-in-module

        result = None

        for suffix in importlib.machinery.EXTENSION_SUFFIXES:
            if result is None or len(suffix) < len(result):
                result = suffix

        return result`

After saving this and running get-hints.py it is working correctly!

kayhayen:
I have corrected the hinted compilation, should work again.

PR #857, opened 2020-10-14
Added getSharedLibrarySuffix to Utils

danielmorgan0x1:
What does this PR do?
#849 - Hinted compilation with nuitka 0.6.9
Adding the getSharedLibrarySuffix to the Utils.py fixes the issue raised in the above issue #849 

Adding the getSharedLibrarySuffix to the Utils.py fixes the issue raised in the above issue #849


kayhayen:
So this is obsolete after I changed the hinted compilation to import from the correct place.

Issue #852, opened 2020-10-07
matplotlib pyplot crashing on doc string


ma6254:
## report error

build is ok
but runtime error:
```
PS C:\soft\aa> .\release\main.dist\main.exe

Traceback (most recent call last):
  File "C:\soft\aa\release\MAIN~1.DIS\main.py", line 3, in <module>
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "C:\soft\aa\release\MAIN~1.DIS\matplotlib\__init__.py", line 107, in <module matplotlib>
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "C:\soft\aa\release\MAIN~1.DIS\matplotlib\rcsetup.py", line 29, in <module matplotlib.rcsetup>
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "C:\soft\aa\release\MAIN~1.DIS\matplotlib\colors.py", line 1169, in <module matplotlib.colors>
  File "C:\soft\aa\release\MAIN~1.DIS\matplotlib\cbook\deprecation.py", line 236, in deprecate
  File "C:\soft\aa\release\MAIN~1.DIS\inspect.py", line 631, in cleandoc
AttributeError: 'ellipsis' object has no attribute 'expandtabs'
PS C:\soft\aa>
```

## Environment

```
gcc.exe (Rev3, Built by MSYS2 project) 10.2.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO 
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Python 3.8.2
pip 20.2.3 from c:\users\mjc\appdata\local\programs\python\python38\lib\site-packages\pip (python 3.8)
```

## python code

```python
import math

from matplotlib import pyplot as plt

x = []
y = []
for k in range(100):
    x.append(k)
    y.append(math.sin(0.5*k))


[...]

PR #873, opened 2020-10-29
fix storing ellipsis as a docstring…

salembream:
# What does this PR do?
fix storing **ellipsis** `...` as a **docstring** when the object [e.g. a class] body is just an ellipsis. 
in standard it should be None, same as when using `pass` as the object body.

Yes #852 , 
here is the code which get mis-compiled, and cause **_deprecated_** decorator in **_matplotlib_** to raise exeption as it assumes the **docstring**  is a string [or None].

\'_matplotlib/colors.py_\'  , note: the `...` is from the code it self, **NOT** presenting omitted text.
```
@cbook.deprecation.deprecated('3.2', alternative='TwoSlopeNorm')
class DivergingNorm(TwoSlopeNorm):
    ...
```

# PR Checklist
#### Note: Unfortunately, I didn't do the tests, it's small direct fix and I don't have a ready Nuitka dev-environment.

      There are Travis and Github Actions tests that cover the most important
      cover enough.

### DIFF
--- a/nuitka/tree/TreeHelpers.py
+++ b/nuitka/tree/TreeHelpers.py
@@ -66,8 +66,9 @@ def extractDocFromBody(node):
             doc = body[0].value.s
             body = body[1:]
         elif getKind(body[0].value) == "Constant":  # python3.8
-            doc = body[0].value.value
-            body = body[1:]
+            if body[0].value.value is not Ellipsis:
+                doc = body[0].value.value
+                body = body[1:]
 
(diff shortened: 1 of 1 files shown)


kayhayen:
Travis Python3.4 seems broken by recent changes of mine, I am just going to remove 3.4 there.

I kind of suspect, that Ellipsis is not the only thing, in theory any non-string must be rejected. I am going to merge it for a hotfix, then try to make it more general, testing e.g. if integers do not pose the same problem.

PR #911, opened 2020-12-26
docs: fix simple typo, resourcs -> resources

timgates42:
There is a small typo in nuitka/utils/WindowsResources.py.

Should read `resources` rather than `resourcs`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md

### DIFF
--- a/nuitka/utils/WindowsResources.py
+++ b/nuitka/utils/WindowsResources.py
@@ -56,7 +56,7 @@ def getResourcesFromDLL(filename, resource_kinds, with_data=False):
         with_data - Return value includes data or only the name, lang pairs
 
     Returns:
-        List of resourcs in the DLL, see with_data which controls scope.
+        List of resources in the DLL, see with_data which controls scope.
 
     """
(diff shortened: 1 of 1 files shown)


kayhayen:
What is strange is what codespell didn't find it. I wonder if it checks the doc strings. I thought it would, but it doesn't complain about this. Yup, would have to be added to their dictionaries, this is one they don't catch: https://github.com/codespell-project/codespell

timgates42:
Thanks @kayhayen I hadn’t heard of codespell before, looks interesting

kayhayen:
Yes, the kind of PR you did here, is automatic to semi-automatic, and if you want, you can add this specific typo I made to their dictionaries in a PR. I learned from it originally in a PR that made a lot of spelling fixes, and then I automated it, and it's even in my PR, so if you care about spelling, that is the way to go, and on any given project, it's very advisable to run it at least occasionally.

PR #972, opened 2021-02-04
Upgrade insecure depends_url to HTTPS

zu3st:
The depends executable is downloaded via HTTP and executed without any signature checks,
allowing potential attackers to hijack the download process and execute malicious code.
Since dependencywalker.com offers TLS, we should use it.

### DIFF
--- a/nuitka/freezer/DependsExe.py
+++ b/nuitka/freezer/DependsExe.py
@@ -33,9 +33,9 @@ def getDependsExePath():
     directory for Nuitka.
     """
     if getArchitecture() == "x86":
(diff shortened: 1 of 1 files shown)


kayhayen:
Good idea, loving it.

kayhayen:
I checked, this was the only instance of this happening, in the source.

License, Debian copyright file format, etc. still reference non-HTTP urls, and of course the old bug tracker is still a lot in the changelog, despite being dead. Occasionally these would need an update too, I suspect esp. the license stuff has better templates now.

PR #1048, opened 2021-03-27
PySidePyQtPlugin: Fix dll check logic

jayvdb:


kayhayen:
Seems perfect, merging before tests finish as there is no real coverage of this one anyway apparently.

kayhayen:
I have included this in 0.6.13.2 hotfix as well.

PR #1049, opened 2021-03-27
fix: locateDLLFromFilesystem fallback in case of no exact match found

### LINKED ISSUE #1046: libuuid not found in Nuitka on Alpine
Hi! 

Minimal example to repeat the issue in docker:

```python
docker run --rm -it python:3.9-alpine sh

apk add libuuid
ls /lib/libuuid*
# /lib/libuuid.so.1      /lib/libuuid.so.1.3.0

pip install nuitka
# Successfully installed nuitka-0.6.13.1

python
from nuitka.utils.SharedLibraries import locateDLL
locateDLL("uuid")
# None                     # Expected: libuuid

from ctypes.util import find_library
find_library("uuid")
# None                     # Expected: libuuid
```

I have this [fix](https://github.com/anki-code/xonsh-portable-binary/blob/3ec06162ff75bd5d78ff17b1c0da74e0acb7dc73/xonsh-portable-musl-alpine.Dockerfile#L35-L39):

```python
#
# Fix Nuitka 0.6.10 + Python 3.6 + Alpine issue when libuuid is installed via `apk add libuuid`
# but `ctypes.util.find_library('uuid')` returns None instead of right path.
#
RUN sed -i 's|locateDLL("uuid")|"/lib/libuuid.so.1.3.0"|g' /usr/local/lib/python3.6/site-packages/nuitka/plugins/standard/ImplicitImports.py
```

Thanks!

anki-code:
Closes #1046

### DIFF
--- a/nuitka/utils/SharedLibraries.py
+++ b/nuitka/utils/SharedLibraries.py
@@ -47,9 +47,8 @@ def locateDLLFromFilesystem(name, paths):
     for path in paths:
         for root, _dirs, files in os.walk(path):
             for file in files:
-                if name in file and '.so' in file:
+                if name in file and ".so" in file:
                     return os.path.join(root, file)
-            
 
 
 _ldconfig_usage = "The 'ldconfig' is used to analyse dependencies on ELF using systems and required to be found."
--- a/nuitka/utils/SharedLibraries.py
+++ b/nuitka/utils/SharedLibraries.py
@@ -43,6 +43,13 @@ def locateDLLFromFilesystem(name, paths):
         for root, _dirs, files in os.walk(path):
             if name in files:
                 return os.path.join(root, name)
+    # Fallback in case of no exact match found
+    for path in paths:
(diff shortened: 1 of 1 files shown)


kayhayen:
Check out developer manual for a commit hook that autoformats code, avoids the pylint issue to begin with.

But this should not help, because ctypes won't resolve it, you said it's None, and therefore no name is known, i.e. you do not enter this code, or is there more changes coming?

kayhayen:
I guess, in case of "None" from the `find_library` in `locateDLL` you could use `"lib" + name` as the `dll_name` and that would make it working. It however looks strange to me, why that DLL is not found. Is there a tool this `find_library` uses on your OS, and that is not installed or something?

anki-code:
@kayhayen Thank you for your questions. Answer - https://github.com/Nuitka/Nuitka/issues/1046#issuecomment-808749618

Issue #980, opened 2021-02-09
errors using pyreadstat with --standalone


gdementen:
* Please check out if the develop version of Nuitka works better for you.

Did not test that as I got 0.6.12, which is not even tagged yet on Github (but does exist on pypi???) so I guess I am running develop already.

* Nuitka version, full Python version and Platform (Windows, OSX, Linux ...)

>nuitka --version
0.6.12
Python: 3.9.1 (default, Dec 11 2020, 09:29:25) [MSC v.1916 64 bit (AMD64)]
Executable: C:\soft\Miniconda3x64\envs\sas2csv-nuitka\python.exe
OS: Windows
Arch: x86_64

* How did you install Nuitka and Python

> conda create -n sas2csv-nuitka python pytz python-dateutil six
> activate sas2csv-nuitka
> pip install pandas # installs numpy too
> pip install pyreadstat
> pip install nuitka

* Many times when you get an error from Nuitka, your setup may be special

a program with just print("hello world") works fine when compiled with --standalone

* SSCCE (this is equivalent to test.py in the .zip file, sample.sas7bdat is also in the zip file):

[nuitka_testcase.zip](https://github.com/Nuitka/Nuitka/files/5951641/nuitka_testcase.zip)

```python
import pyreadstat

if __name__ == '__main__':
    df, meta = pyreadstat.read_sas7bdat('sample.sas7bdat')
    df.to_csv('sample.csv')
```
compiles fine using either --standalone or --onefile and ultimately produces the following traceback at runtime:

```python
Traceback (most recent call last):
  File "C:\Users\gdm\AppData\Local\Test\test\001~1.0\test.py", line 3, in <module>
    import pyreadstat
  File "C:\Users\gdm\AppData\Local\Test\test\001~1.0\pyreadstat\__init__.py", line 17, in <module pyreadstat>
  File "pyreadstat\pyreadstat.pyx", line 1, in init pyreadstat.pyreadstat
  File "pyreadstat\_readstat_parser.pyx", line 28, in init pyreadstat._readstat_parser
ModuleNotFoundError: No module named 'pandas'
```

Adding an explicit "import pandas" like below (test2.py in the zip file) gets me seemingly further...
```python
import pyreadstat
import pandas

if __name__ == '__main__':
    df, meta = pyreadstat.read_sas7bdat('sample.sas7bdat')
    df.to_csv('sample.csv')
```
... but ultimately crashes at runtime too with:

```python
Traceback (most recent call last):
[...]

PR #1072, opened 2021-04-20
pyreadstat

gdementen:
fixes using pyreadstat with --standalone (issue #980)

### DIFF
--- a/nuitka/plugins/standard/ImplicitImports.py
+++ b/nuitka/plugins/standard/ImplicitImports.py
@@ -1025,6 +1025,9 @@ class NuitkaPluginPopularImplicitImports(NuitkaPluginBase):
             yield "py._path.local"
         elif full_name == "pyreadstat._readstat_parser":
             yield "pandas"
+        elif full_name == "pyreadstat.pyreadstat":
+            yield "pyreadstat._readstat_writer"
+            yield "pyreadstat.worker"
         elif full_name == "cytoolz.itertoolz":
             yield "cytoolz.utils"
         elif full_name == "cytoolz.functoolz":


gdementen:
Note that this makes all my test scripts work (minimal.py and test.py) when using --standalone (or --onefile) but I am unsure whether or not there should be other imports added. I just had a look at pyreadstat cython files and they import a bunch of other stuff. I guess it works because pandas is required and it imports those too, but I wonder whether or not I should add those imports too. FWIW, here are the import blocks of those cython files:

pyreadstat/pyreadstat.pyx
```cython
import multiprocessing as mp

import pandas as pd

from readstat_api cimport readstat_parse_sas7bdat, readstat_parse_dta, readstat_parse_sav
from readstat_api cimport readstat_parse_por, readstat_parse_xport
from readstat_api cimport readstat_parse_sas7bcat
from readstat_api cimport readstat_begin_writing_dta, readstat_begin_writing_por, readstat_begin_writing_sav
from _readstat_parser cimport py_file_format, run_conversion
from _readstat_writer cimport run_write
cimport _readstat_parser, _readstat_writer
from copy import deepcopy
from worker import worker
```

pyreadstat/_readstat_parser.pyx
```cython
from cpython.datetime cimport import_datetime, timedelta_new, datetime_new
from cpython.exc cimport PyErr_Occurred
from cpython.object cimport PyObject
from libc.math cimport NAN, floor

#from datetime import timedelta, datetime
from collections import OrderedDict
import os
is_pathlib_available = False
try:
    from pathlib import Path
    is_pathlib_available = True
except:
    pass

import pandas as pd
import numpy as np
#from pandas._libs import Timestamp

from readstat_api cimport *

# necessary to work with the datetime C API
import_datetime()
```

pyreadstat/_readstat_writer.pyx
```cython
import os
import numpy as np
#cimport numpy as np
import pandas as pd
import datetime
import calendar
IF PY_MAJOR_VERSION >2:
    from datetime import timezone as _timezone
ELSE: # not available on python 2
    import pytz as _timezone
from libc.math cimport round, NAN
is_pathlib_available = False
try:
    from pathlib import Path
    is_pathlib_available = True
except:
    pass

from readstat_api cimport *
from _readstat_parser import ReadstatError, PyreadstatError
from _readstat_parser cimport check_exit_status
```

kayhayen:
I am going to include it in a hotfix that is coming

Issue #865, opened 2020-10-19
Too many modules will be optimized, caching is needed


torchsmoke:
When I compile my program the modules are optimized first. Unfortunately far too many are optimized. For example PyOpenGL although I import subprocessed. And that leads to the fact that I have to wait half an hour just to optimize all modules.
Solution 1: Nuitka only has to optimize all modules it needs.
Solution 2: Nuitka optimizes them all and stores them in a cache. Then next time only updated modules have to be optimized and saved again and only the needed ones have to be loaded from the cache. This would have the additional advantage that the compilation times are reduced considerably.
What do you think about the solutions?

kayhayen:
The second one is the plan, and exists as an issue already. Without it, Nuitka is practically unusable. Sorry for closing this as a duplicate though.

Ben-Faessler:
If you haven't started yet, I'd like to get started with the caching system. Can I help you?

kayhayen:
I did search a lot now, and shamefully I cannot find it. Many issues for performance mention the topic, but they don't work on solutions.

As far as caching goes, I started with something that #415 would greatly benefit from, namely the writing of used modules to a cache file when doing the demotion. Some modules, which are too large to compile as C code, still take long to compile in Nuitka. And the goal is to do it only once, store it in the cache, and next time we compile, to take the used modules, and create a bytecode module right away with the proper dependencies.

There is `pkg_resources` which includes a copy of pyparsing for whatever reason. And it won't work without it being there, even if rarely ever used. That module takes time to compile each time. Also standard library could be handled as well with full compilation for one time, and properly analyse dependencies, to find that e.g. sqlite is used or not. There is plenty of benefit from just that level of caching.

If you want to peruse that, the bytecode demotion code pointed to in #415 writes things, and the creation of the module from source code, ought to still let plugins have a go (they might modify it), and then then a bytecode module rather than the compile module.

Getting the caching key right will be a large part of the trick. The `sys.path` relevant contents is to be scanned, and I think I started that work. As that affects the used modules analysis. The hash of the sourcecode should be used, and of course Python version, etc.

With that being present, it ought to be possible to demote trouble makers, or whole packages from stdlib that you don't use and have things work fast.

Ben-Faessler:
If I understand that correctly, we need a cache that contains the .o files of the respective scripts and the hash of the source code. So we can actually save the .o for each script and use the hash (sha1?) Of the source code as the file name.

[...]

PR #1081, opened 2021-05-01
Bytecode caching

Ben-Faessler:
# What does this PR do?
I finished the bytecode caching
#865
# PR Checklist

      cover enough.

### DIFF
--- a/nuitka/optimizations/BytecodeDemotion.py
+++ b/nuitka/optimizations/BytecodeDemotion.py
@@ -112,7 +112,9 @@ def demoteCompiledModuleToBytecode(module):
     uncompiled_module = makeUncompiledPythonModule(
         module_name=full_name,
         filename=filename,
-        bytecode=marshal.dumps(bytecode) if not already_demoted else _loadBytecodeFromCache(_getCacheFilename(full_name, module_importables_hash, "dat")),
+        bytecode=marshal.dumps(bytecode) if not already_demoted else _loadBytecodeFromCache(_getCacheFilename(
+            full_name, module_importables_hash, "dat")
+        ),
         is_package=module.isCompiledPythonPackage(),
         user_provided=True,
         technical=False,
--- a/nuitka/optimizations/BytecodeDemotion.py
+++ b/nuitka/optimizations/BytecodeDemotion.py
@@ -66,14 +66,17 @@ def getModuleImportableFilesHash(full_name):
     return result_hash.hexdigest()
 
 
-def _getCacheFilename(full_name, module_importables_hash):
+def _getCacheFilename(full_name, module_importables_hash, extension):
     module_cache_dir = os.path.join(getCacheDir(), "module-imports-cache")
     makePath(module_cache_dir)
 
     return os.path.join(
-        module_cache_dir, "%s-%s.txt" % (full_name.asString(), module_importables_hash)
+        module_cache_dir, "%s-%s.%s" % (full_name.asString(), module_importables_hash, extension)
     )
 
+def _loadBytecodeFromCache(cache_filename):
+    with open(cache_filename, "rb") as bytecode_cache_file:
+        return bytecode_cache_file.read()
 
 def demoteCompiledModuleToBytecode(module):
     """Demote a compiled module to uncompiled (bytecode)."""
@@ -86,31 +89,38 @@ def demoteCompiledModuleToBytecode(module):
             "Demoting module %r to bytecode from %r." % (full_name, filename)
         )
 
-    source_code = module.getSourceCode()
+    module_importables_hash = getModuleImportableFilesHash(full_name)
+    if os.path.exists(_getCacheFilename(full_name, module_importables_hash, "txt")):
+        inclusion_logger.info("The module %r is already demoted" % (full_name))
+        already_demoted = True
 
-    # Second chance for plugins to modify source code just before turning it
-    # to bytecode.
-    source_code = Plugins.onFrozenModuleSourceCode(
-        module_name=full_name, is_package=False, source_code=source_code
-    )
+    else:
+        already_demoted = False
+        source_code = module.getSourceCode()
 
-    bytecode = compile(source_code, filename, "exec", dont_inherit=True)
+        # Second chance for plugins to modify source code just before turning it
+        # to bytecode.
+        source_code = Plugins.onFrozenModuleSourceCode(
+            module_name=full_name, is_package=False, source_code=source_code
+        )
 
-    bytecode = Plugins.onFrozenModuleBytecode(
-        module_name=full_name, is_package=False, bytecode=bytecode
-    )
+        bytecode = compile(source_code, filename, "exec", dont_inherit=True)
+
+        bytecode = Plugins.onFrozenModuleBytecode(
+            module_name=full_name, is_package=False, bytecode=bytecode
+        )
 
     uncompiled_module = makeUncompiledPythonModule(
         module_name=full_name,
         filename=filename,
-        bytecode=marshal.dumps(bytecode),
+        bytecode=marshal.dumps(bytecode) if not already_demoted else _loadBytecodeFromCache(_getCacheFilename(full_name, module_importables_hash, "dat")),
         is_package=module.isCompiledPythonPackage(),
         user_provided=True,
         technical=False,
     )
 
     used_modules = module.getUsedModules()
-
+    print(used_modules)
     uncompiled_module.setUsedModules(used_modules)
     module.finalize()
 
@@ -123,8 +133,10 @@ def demoteCompiledModuleToBytecode(module):
     if isTriggerModule(module):
         replaceTriggerModule(old=module, new=uncompiled_module)
 
-    module_importables_hash = getModuleImportableFilesHash(full_name)
+    if not already_demoted:
+        with open(_getCacheFilename(full_name, module_importables_hash, "txt"), "w") as modules_cache_file:
+            for module_name, _filename in used_modules:
+                modules_cache_file.write(module_name.asString() + "\n")
 
-    with open(_getCacheFilename(full_name, module_importables_hash), "w") as cache_file:
-        for module_name, _filename in used_modules:
-            cache_file.write(module_name.asString() + "\n")
+        with open(_getCacheFilename(full_name, module_importables_hash, "dat"), "wb") as bytecode_cache_file:
+            bytecode_cache_file.write(marshal.dumps(bytecode))
--- a/nuitka/optimizations/BytecodeDemotion.py
+++ b/nuitka/optimizations/BytecodeDemotion.py
@@ -34,7 +34,7 @@ from nuitka.nodes.ModuleNodes import makeUncompiledPythonModule
 from nuitka.plugins.Plugins import Plugins
 from nuitka.Tracing import inclusion_logger
 from nuitka.utils.AppDirs import getCacheDir
-from nuitka.utils.FileOperations import listDir, makePath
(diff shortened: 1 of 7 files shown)


nuitka/optimizations/BytecodeDemotion.py
```diff
@@ -66,14 +66,17 @@ def getModuleImportableFilesHash(full_name):
...
+def _getCacheFilename(full_name, module_importables_hash, extension):
     module_cache_dir = os.path.join(getCacheDir(), "module-imports-cache")
     makePath(module_cache_dir)
 
     return os.path.join(
-        module_cache_dir, "%s-%s.txt" % (full_name.asString(), module_importables_hash)
+        module_cache_dir, "%s-%s.%s" % (full_name.asString(), module_importables_hash, extension)
     )
 
+def _loadBytecodeFromCache(cache_filename):
```
kayhayen, line 77: Using `getFileContents` from `nuitka.utils.FileOperations` will respect locking. This might be important later down the road, if Nuitka runs in more processes, and avoids the code.

kayhayen (changes requested):
Outside of the print, which the CI will hate absolutely, this looks perfect to me.

Ben-Faessler:
I have fixed the pylint error, but I am not responsible for the Windows error.

kayhayen:
Annoyingly I had to click run for PRs now, I will look into the new results, the Windows error looked like the print only, so hopefully it's good. Onefile tests currently occasionally hang, I have stuff for that in the queue on factory, but it's not complete. Sorry about the sad state of you having to wait for me.

nuitka/optimizations/BytecodeDemotion.py
```diff
@@ -86,31 +88,40 @@ def demoteCompiledModuleToBytecode(module):
...
+            full_name, module_importables_hash, "dat")
+        ),
         is_package=module.isCompiledPythonPackage(),
         user_provided=True,
         technical=False,
     )
 
     used_modules = module.getUsedModules()
-
+    print(used_modules)
```
kayhayen, line 124: This `print` is causing the test failures. It happening on Windows is probably due to only it have a bytecode demoted module in tests scope.

Ben-Faessler:
Oh, that was just for debugging purposes. I must have forgotten to delete it. I will delete this with the following commits. I still tried to add an option that changes the location of the whole cache, so ccache too. Unfortunately the test failed with an error I can't explain.
```
Nuitka-Scons:INFO: Backend C compiler: /usr/lib/ccache/gcc-8 (gcc-8).                                                                                                                                                                         
AttributeError: 'NoneType' object has no attribute 'cache_dir':
  File "/usr/local/lib/python3.7/dist-packages/Nuitka-0.6.15rc5-py3.7.egg/nuitka/build/Backend.scons", line 1292:                                                                                                                             
    assume_yes_for_downloads=assume_yes_for_downloads,
  file "/usr/local/lib/python3.7/dist-packages/Nuitka-0.6.15rc5-py3.7.egg/nuitka/build/SconsCaching.py", line 173:                                                                                                                            
    ccache_dir = os.path.join(getCacheDir(), "ccache")
  File "/usr/local/lib/python3.7/dist-packages/Nuitka-0.6.15rc5-py3.7.egg/nuitka/utils/AppDirs.py", line 44:                                                                                                                                  
    if Options.getCacheDir() is not None:
  File "/usr/local/lib/python3.7/dist-packages/Nuitka-0.6.15rc5-py3.7.egg/nuitka/Options.py", line 426:                                                                                                                                       
    return options.cache_dir
```
Maybe you know what the problem is?

kayhayen:
@Ben-Faessler this looks strange:

>  File "/usr/local/lib/python3.7/dist-packages/Nuitka-0.6.15rc5-py3.7.egg/nuitka/utils/AppDirs.py", line 44:                                                                                                                                  
>    if Options.getCacheDir() is not None:

Could this be something you did, never was that like in that modules code, unless I am very mistaken.

kayhayen:
As an aside, for development, I recommend to do `python setup.py develop` which will make Nuitka register something that hooks your git checkout, that avoids a need to install. Alternatively, use the runners in the bin folder, they will set the `sys.path` to the code near them, but I think `develop` is the most clever way, esp. if you are in a virtualenv.

Ben-Faessler:
> @Ben-Faessler this looks strange:
> 
> > File "/usr/local/lib/python3.7/dist-packages/Nuitka-0.6.15rc5-py3.7.egg/nuitka/utils/AppDirs.py", line 44:
> > if Options.getCacheDir() is not None:
> 
> Could this be something you did, never was that like in that modules code, unless I am very mistaken.

Yes, that comes from me. I need that to manipulate the _cache_dir.

Ben-Faessler:
Should I commit the changes for the command line option? Then maybe you can tell what is wrong.

kayhayen:
Some parts of the code are used by the scons process and the nuitka process. Only nuitka has a functional nuitka.Options module, scons will not work with that. You need to not use options in there, but more on the outside, in code only used by the frontend, and not from Backend.scons for things to work.

kayhayen:
Yes, you can always push what you have got, I can then comment on it, and ask for changes.

kayhayen:
I think you are not yet using the cached stuff in the pushed code, or I am missing something. There needs to be some change in the tree building, where it's considered.

Ben-Faessler:
In my tests, Nuitka used the cached bytecode. However, only if nothing has changed in the folder where the programme to be compiled is located (deleting or adding files and folders). Meanwhile I have the same modules in the cache 5 times...

Ben-Faessler:
I just got an idea. Would it make sense if I move the caching in a separate module. So you could limit the BytecodeDemotion on their actual job, and it would be easier for example. implement a C code caching. As a name suggest "nuitka.utils.Cache"

kayhayen:
The way this works now, is that Nuitka in `nuitka.tree.Building` creates a compiled module and optimizes it and later when optimization is finished, it is replaced with a uncompiled module. Your cache kicks in too late, where it doesn't avoid this.

You need to make use of the cache here: https://github.com/Nuitka/Nuitka/blob/0c88a3c7fa0f79613d09b998d3151ace3601312e/nuitka/tree/Building.py#L992

There, you should if bytecode is the compilation mode decided, for an `UncompiledPythonModule` rather than a `CompiledPythonModule`, and then it will be great. Using the cache where you use it in terms of reading back has no point, although it probably demonstrates correctness, but you didn't achieve it to be avoiding the compilation step at all, outside of demoting more things of course, but that's for the option to control that.

kayhayen:
Regarding the location of things, I need think a bit. Do whatever feels right to you, it's all changeable later, but I think `nuitka.optimization` and `nuitka.tree` are better places, or `nuitka.Caching` maybe even, as this is probably something that they all are sharing, but not really a utility if you know what I mean, at least not outside of determining version hashes, etc.

Ben-Faessler:
Where do I get all the values for the parameters?
Code snippet from Recursion.py:

  module_name=module.getFullName(),
  filename=module_filename,
  bytecode=marshal.dumps(
      compile(
          source_code, module_filename, "exec", dont_inherit=True
          )
      ),
  is_package=module.isCompiledPythonPackage(),
  user_provided=True,
  technical=False,
  )

kayhayen:
I guess, you need to cache all required information. The bytecode isn't actually important to cache, you can just compile it again, but it being a package, is contextual knowledge. Can you give a better point, so I can tell why you have to do that? I thought you only have to replace this:

```
            result = CompiledPythonModule(
                module_name=module_name,
                is_top=is_top,
                mode=decideCompilationMode(is_top, module_name, source_ref),
                future_spec=None,
                source_ref=source_ref,
            )

```

There this is all available and you just need to change the class to be constructed, depending on the decided compilation mode, making that the condition of an if statement and checking the cache availability.

Ben-Faessler:
I thought I had to do something like that.
```
            mode = decideCompilationMode(is_top, module_name, source_ref)
            if mode == "bytecode":
                result = UncompiledPythonModule(
                    module_name=module_name,
                    filename=source_ref.getFilename(),
                    bytecode=
                    is_package=False,
                    user_provided=False,
                    technical=False,
                    )
            else:
                result = CompiledPythonModule(
                    module_name=module_name,
                    is_top=is_top,
                    mode=mode,
                    future_spec=None,
                    source_ref=source_ref,
                )
```

> There this is all available and you just need to change the class to be constructed, depending on the decided compilation mode, making that the condition of an if statement and checking the cache availability.

Where should I change that?

kayhayen:
That code lives in `nuitka.tree.Building` function `decideModuleTree`. Do not bother with `is_main` because the top level module must not be bytecode or else you get into trouble with the linkage. In fact, should `is_top` be set, you must not do it as well, which will matter for compiling in `--module` mode.

kayhayen:
In `buildModuleTree` when you got the uncompiled module, you definitely should not use the source code anymore, that code wants to add the nodes with the Python code to the compiled module, which was created empty. Just check if its a `isCompiledPythonModule` and then `and` that to checking if there is a source filename. That branch exists e.g. because extension modules are not having source either. Being bytecode module is the same, except you were passed in source filename, but now you know to not use it anymore.

Ben-Faessler:
> Some parts of the code are used by the scons process and the nuitka process. Only nuitka has a functional nuitka.Options module, scons will not work with that. You need to not use options in there, but more on the outside, in code only used by the frontend, and not from Backend.scons for things to work.

How else can I use it?

kayhayen:
> 
> 
> > Some parts of the code are used by the scons process and the nuitka process. Only nuitka has a functional nuitka.Options module, scons will not work with that. You need to not use options in there, but more on the outside, in code only used by the frontend, and not from Backend.scons for things to work.
> 
> How else can I use it?

You need to do it in code not shared between them, the appdata module clearly is too low level. You should only use it where it makes sense, or is it your intention to affect all appdata caching dirs there? Just don't do that I suppose. There are good reasons for using standard directories, and even if not, that's a separate issue.

Ben-Faessler:
> In `buildModuleTree` when you got the uncompiled module, you definitely should not use the source code anymore, that code wants to add the nodes with the Python code to the compiled module, which was created empty. Just check if its a `isCompiledPythonModule` and then `and` that to checking if there is a source filename. That branch exists e.g. because extension modules are not having source either. Being bytecode module is the same, except you were passed in source filename, but now you know to not use it anymore.

Can you please be more precise?

kayhayen:
I am manually merging this right now, rebasing and such, and making it harmless (needs an experimental flag), so it can be included in the next release, and then debug it for a bit, I think it ought to work almost.

kayhayen:
I am deciding against caching of the bytecode though, as it would be impacted by modes, e.g. `no_doc`, `no_annotations` and the like, for no good reason. Also plugins can interact with this better that way. This is assuming that bytecode compilation itself is fast, which I think is true.

This is only about caching the included modules of said module, to avoid the compilation of Python to figure those out.

kayhayen:
This is on factory, how to do proceed with this... I don't want you have this as rejected, when it fact it should be accepted, so I guess, I am going to merge it, and immediately undo it on github develop with a rebase of current factory, when I do the next pre-release.

It's still not working, but disabled by default, the only reason is that dependent modules are not imported when using the cache, so they are missing during optimization. That should be easy to overcome. But I would like to eliminate some duplication in `nuitka.Recursion` and `nuitka.Building` related to creating modules, as they both have to similar things.

kayhayen:
Ok, doing that, it's controlled by an experimental option I added to `nuitka.Caching` that deals with the presence of information, it's always false unless allowed to be used.

PR #1104, opened 2021-05-21
Nuitka-Python Integration with Windows Static Build

Maxwell175:

# What does this PR do?

Implements a static Windows build using the Nuitka-Python fork.

# PR Checklist

      cover enough.

### DIFF
--- a/nuitka/build/Backend.scons
+++ b/nuitka/build/Backend.scons
@@ -58,6 +58,7 @@ from nuitka.Tracing import (
     scons_logger,
     setQuiet,
 )
+from nuitka.utils.Utils import isNuitkaPython
 
 from .SconsCaching import enableCcache, enableClcache
 from .SconsCompilerSettings import (
@@ -829,7 +830,7 @@ def getWindowsPythonDLLPath():
 
 if win_target:
     # Some non standard Pythons on Windows have this.
-    if sys.version_info[0] > 2 and sys.implementation.name == "nuitkapython":
+    if isNuitkaPython():
         with open(os.path.join(python_prefix, "link.json")) as f:
             linkData = json.load(f)
             env.Append(LIBS=linkData["libraries"])
--- a/nuitka/build/Backend.scons
+++ b/nuitka/build/Backend.scons
@@ -829,7 +829,7 @@ def getWindowsPythonDLLPath():
 
 if win_target:
     # Some non standard Pythons on Windows have this.
-    if sys.implementation.name == "nuitkapython":
+    if sys.version_info[0] > 2 and sys.implementation.name == "nuitkapython":
         with open(os.path.join(python_prefix, "link.json")) as f:
             linkData = json.load(f)
             env.Append(LIBS=linkData["libraries"])
--- a/nuitka/build/Backend.scons
+++ b/nuitka/build/Backend.scons
@@ -41,6 +41,7 @@ __package__ = "nuitka.build"  # pylint: disable=redefined-builtin
 
 # isort:start
 
+import json
 import platform
 import subprocess
 
@@ -741,6 +742,7 @@ if python_debug:
 if static_libpython:
     env.Append(CPPDEFINES=["Py_NO_ENABLE_SHARED"])
 
+
 def detectHostMultiarch():
     pipe = SCons.Action._subproc(  # pylint: disable=protected-access
         env,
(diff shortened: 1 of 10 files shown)


nuitka/build/Backend.scons
```diff
@@ -825,8 +827,11 @@ def getWindowsPythonDLLPath():
 
 if win_target:
     # Some non standard Pythons on Windows have this.
-    if gcc_mode and static_libpython:
+    if static_libpython:
         env.Append(LIBS=[env.File(static_libpython)])
+        env.Append(LIBPATH=[os.path.join(python_prefix, 'libs')])
+
+        env.Append(LIBS=["User32.lib", "Advapi32.lib", "Ole32.lib", "OleAut32.lib", "Ws2_32.lib", "version.lib", "Iphlpapi.lib", "crypt32.lib", "Rpcrt4.lib", "Shlwapi.lib", "libcrypto.lib", "libssl.lib", "_asyncio.lib", "_bz2.lib", "ffi.lib", "_ctypes.lib", "_decimal.lib", "_elementtree.lib", "_hashlib.lib", "liblzma.lib", "_lzma.lib", "cabinet.lib", "msi.lib", "_msi.lib", "_multiprocessing.lib", "_overlapped.lib", "_queue.lib", "_socket.lib", "sqlite3.lib", "_sqlite3.lib", "_ssl.lib", "tcl86t.lib", "tk86t.lib", "_tkinter.lib", "_uuid.lib", "_zoneinfo.lib", "pyexpat.lib", "select.lib", "unicodedata.lib", "winsound.lib", "Winmm.lib"])
```
kayhayen, line 834: Would there be a way to tell that it's Nuitka-Python? These won't exist otherwise.

nuitka/utils/WindowsResources.py
```diff
@@ -155,6 +155,7 @@ def callback2(hModule2, lpType2, lpName2, wLang, _lParam):
 
 
 def _openFileWindowsResources(filename):
+    absFilename = os.path.abspath(filename)
```
kayhayen, line 158: Do you recall, why the `abspath` calls were necessary? I normally try to avoid them as they can only cause length issues and indetermism.

kayhayen (commented):
I will try this out. I think I would like to know if we can do without the `abspath`.

Also, what it tried to do with `ctypes` extension module for standard library path, might be an issue, although I hope it's that dynload folder that you won't have with Nuitka-Python.

Regarding the detection of Nuitka-Python, we should have something in `sys.version` that allows us to pass it as a mode to Scons as well.

kayhayen:
From the CI results, I would say, you need to consider the `#define ` you have for static Python, and probably even better for Nuitka-Python, such that it works with the other things, but that's not a priority. If you don't mind, I would work on this a bit to make it merge ready, just let me know what to look for in `sys.version`. I guess, this can be merged without having any impact at all on other things.

nuitka/utils/WindowsResources.py
Maxwell175: Without the abspath, it threw a file not found type of error.

nuitka/build/Backend.scons
Maxwell175: Well, in theory they are already in a check for if it is windows or not and Nuitka - Python is the only static windows version I believe. But nonetheless we can add a check based on platform.python_implementation() here.

nuitka/build/Backend.scons
```diff
@@ -825,7 +829,17 @@ def getWindowsPythonDLLPath():
 
 if win_target:
     # Some non standard Pythons on Windows have this.
-    if gcc_mode and static_libpython:
+    if sys.version_info[0] > 2 and sys.implementation.name == "nuitkapython":
```
kayhayen, line 832: The Python3 check seems redundent, also need to check `python_version` as passed from the outside, as Scons itself may be running with another Python, although that's going to be rare, but I would just remove the check here.

nuitka/build/Backend.scons
```diff
@@ -825,7 +829,17 @@ def getWindowsPythonDLLPath():
...
+        with open(os.path.join(python_prefix, "link.json")) as f:
+            linkData = json.load(f)
+            env.Append(LIBS=linkData["libraries"])
```
kayhayen, line 835: Nice. I like using json and doing it here much better.

nuitka/build/SconsInterface.py
```diff
@@ -163,7 +163,8 @@ def _setupSconsEnvironment():
         if os.getcwd() != os.getcwdu():
             os.chdir(getWindowsShortPathName(os.getcwdu()))
 
-    if Utils.isWin32Windows():
+    from nuitka.Options import shallUseStaticLibPython
```
kayhayen, line 166: This is not going to work well, Options need to be passed via command line, as the options object is not going to be there, you are getting some sort of default here. But it's already passed and available as the `static_libpython` variable, so just use that please.

nuitka/utils/WindowsResources.py
kayhayen: I have avoided absolute paths typically, because it's easy to run into length limits for no good reason. Make sure it's not a locking issue with Windows Defender, that is actually more often the case, but I trust you there.

nuitka/utils/WindowsResources.py
```diff
@@ -164,7 +165,7 @@ def _openFileWindowsResources(filename):
 
     BeginUpdateResource.restype = ctypes.wintypes.HANDLE
 
-    update_handle = BeginUpdateResource(filename, False)
+    update_handle = BeginUpdateResource(absFilename, False)
```
kayhayen, line 168: Just a minor style issue, can you name it `fullpath` as that is what I used more often for this kind of variable in other codes in Nuitka.

nuitka/build/Backend.scons
Maxwell175: The reason the check is there is that  sys.implementation.name does not exist on 2.7 and it caused the CI to fail. So I will just change the check to use python_version

nuitka/build/SconsInterface.py
Maxwell175: Please clarify where I can find the `static_libpython` variable from here.

kayhayen: Oh, that is me confused, I somehow thought this is in the scons file.

kayhayen:
Merging, thanks for your hard work on this.

Issue #769, opened 2020-06-26
Nuitka + GTK GI: Fail, because some initializations not called when Nuitka


belonesox:
# Nuitka + GTK GI: Fail, because some initializations not called when Nuitka

GI =  Global Introspection (GTK+GDK) 

Lets consider simple (minimal for reproduce the problem) python3 file:

```python
import gi
import cairo
gi.require_version('Gtk', '3.0')
gi.require_version('Gdk', '3.0')
from gi.repository import Gtk
Gtk.init()
from gi.repository import Gdk

image = cairo.ImageSurface.create_from_png("image.png")
region = Gdk.cairo_region_create_from_surface(image)

print('All OK')
```

It should work on regular linux python3
if installed package
* ``python3-gi-cairo`` on (Debian/Ubuntu) 
* ``python3-gobject-base`` on Fedora

But if we compile it on nuitka, 
we got ``KeyError: 'could not find foreign type Surface`` on line 

```  
  region = Gdk.cairo_region_create_from_surface(image)
```

I did some research. 
The exception occured on 
``pygi_struct_foreign_convert_to_g_argument`` (see
https://gitlab.gnome.org/GNOME/pygobject/-/blob/3.36.1/gi/pygi-foreign.c#L111 or can be similar error «Couldn't find foreign struct converter for» on ``pygi_struct_foreign_lookup_by_name`` 
(see https://gitlab.gnome.org/GNOME/pygobject/-/blob/3.36.1/gi/pygi-foreign.c#L74)

This is because when used  regular python3 we 
call 
``PYGLIB_MODULE_START(_gi_cairo, "_gi_cairo")``
( https://gitlab.gnome.org/GNOME/pygobject/-/blob/3.36.1/gi/pygi-foreign-cairo.c#L617 )
where called a lot of ``pygi_register_foreign_struct`` to register
all cairo types as foreigh.

Long stack from https://gitlab.gnome.org/GNOME/pygobject/-/blob/3.36.1/gi/pygi-foreign-cairo.c#L619 when regular python3 called:
```
/usr/lib64/python3.8/site-packages/gi/_gi_cairo.cpython-38-x86_64-linux-gnu.so(+0x385b) [0x7f83808b185b]
/usr/lib64/python3.8/site-packages/gi/_gi_cairo.cpython-38-x86_64-linux-gnu.so(+0x38fc) [0x7f83808b18fc]
[...]

PR #1111, opened 2021-05-28
Include Gtk overrides

zegervdv:
# What does this PR do?
Add the implicit imports of overrides for Gtk etc.

I'm not sure if this is the way to go?
If so, probably some more imports need to be added or made more selectively.

In an attempt to provide a fix for #769.

# PR Checklist

      cover enough.


kayhayen:
I think you ought to simply add these to `nuitka/plugins/standard/ImplicitImports.py` where there are already a few for the `gi` module.

zegervdv:
This should be better then?

kayhayen:
Part of 0.6.15.1 hotfix release.

PR #1131, opened 2021-06-18
onefile icons linux: look for versioned icon file first

bearsh:
for look for an icon for pythonMAJOR.MINOR.xpm, then pythonMAJOR.xpm
then python.xpm.
on some systems (e.g. ubuntu) python.xpm does not exist.


kayhayen:
Thanks, that is a nice addition.

Issue #1154, opened 2021-07-13
Onefile application on windows does not work anymore once singed


bearsh:
On Windows, signing a application appends the necessary information at the end of the file in a _Certificate Table / Security directory_. Additionally a reference to this table is added to the optional pe header.

The current bootstrap code looks at the very end of the file to get the start address of the embedded container which contains the actual application. Once an additional table is present at the end of the file, this lookup fails.

Proposed solution:
- use the function form `imagehlp.h` (or add a simple PE parser) to determine if the file is singed (cert table at the end)
- if a certificate table is existing, travers backward to get the embedded image start address (maybe an additional padding is necessary when placing the address as the certificate table is quadword-aligned)

What I don't know is if the certificate table is the only piece of data which may be appended to the file by an external tool. if there are other data which gets appended in the real world, the proposed solution may fail again.... But it would fail anyway

PR #1167, opened 2021-07-22
add support for singed onefile-application on windows

bearsh:
parse the pe header using MapAndLoad() system function
to find out whether the application has a trailling certificate
rather then assume that the start position of the container is at the
very end of the file.

Close: #1154

### DIFF
--- a/nuitka/build/Onefile.scons
+++ b/nuitka/build/Onefile.scons
@@ -688,6 +688,9 @@ target = env.Program(result_exe, source_files + source_targets)
 # Avoid dependency on MinGW libraries.
 if win_target and gcc_mode and not clang_mode:
     env.Append(LINKFLAGS=["-static-libgcc"])
+# on Windows link against imagehlp to support signed applications detection
+if win_target:
+    link_libraries.append("imagehlp")
 
 # Avoid IO for compilation as much as possible, this should make the
 # compilation more memory hungry, but also faster.
--- a/nuitka/build/static_src/OnefileBootstrap.c
+++ b/nuitka/build/static_src/OnefileBootstrap.c
@@ -44,6 +44,7 @@
 #if defined(_WIN32)
 #include <Shlobj.h>
 #include <windows.h>
+#include <imagehlp.h>
 
 #ifndef CSIDL_LOCAL_APPDATA
 #define CSIDL_LOCAL_APPDATA 28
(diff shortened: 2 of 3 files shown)


nuitka/build/static_src/OnefileBootstrap.c
```diff
@@ -638,7 +654,34 @@ int main(int argc, char **argv) {
...
+       the overcome this limitation, use the windows function MapAndLoad()
+       to parse the PE header. The header contains information whether
+       a signature is present and at which address the first signature
+       start. so we can use that address to find the start position value */
+    DWORD cert_table_addr = 0;
+
+    PSTR exe_filename_a = convertPcwstrToPstr(exe_filename);
+    if (exe_filename_a) {
+        LOADED_IMAGE loaded_image;
+        if (MapAndLoad(exe_filename_a, "\\dont-search-path", &loaded_image, false, true)) {
```
kayhayen, line 668: This need to convert to ANSI path, is worrisome. I guess using `GetShortPathNameW` before would make sure it works.

nuitka/build/Onefile.scons
```diff
@@ -688,6 +688,9 @@ target = env.Program(result_exe, source_files + source_targets)
 # Avoid dependency on MinGW libraries.
 if win_target and gcc_mode and not clang_mode:
     env.Append(LINKFLAGS=["-static-libgcc"])
+# on Windows link against imagehlp to support signed applications detection
+if win_target:
```
kayhayen, line 692: Do you think it would be wrong for Nuitka to call the signing tool itself, or to be informed it will. Because the code you are adding to onefile bootstrap is not conditional, so Nuitka would have to know ahead of time and not include overhead for those not using it.

kayhayen:
I was thinking, that if INCBIN was working with MSVC, we ought to have used that. Putting all the data through the C compiler seems bad, but doable.

Windows Resources might be an alternative, although they won't have ReadFile working, and are ugly to attach. And generally we could try and search for a unique pattern from end of file, that indicates the start of our data, which would be most portable.

For now, I am leaning towards including your code, if it was reliable. I think without short paths, conversion to ANSI can fail for general unicode. Nuitka does this e.g. in `getWindowsShortPathName` which miracously makes things work that otherwise won't.

But changing it to use resources is something that ought to be done, but I sure lack the time right now.

nuitka/build/Onefile.scons
bearsh: I don't think Nuitka should call the signing tool as signing could take place on another machine (e.g. if using a hw token). I'm fine with adding a option to Nuitka to enable/disable the PE header parsing but I think it should be enabled by default as in morst cases, singing an exe just works.

bearsh:
code is updated to use GetShortPathNameW()

kayhayen:
I (actually the C compiler) found a few problems with it, one being 

```
    l = GetShortPathNameW(pszPath, shortPath, l);
    if (l = 0) {
        goto err_shortPath;
    }
```

So it should be ineffective as is, I am fixing that myself right now.

kayhayen:
The github develop branch should be good now.

kayhayen:
Dang, there is another issue, the buffer is allocated without storage for the null byte.

From the docs: 

> If the function succeeds, the return value is the length, in TCHARs, of the string that is copied to lpszShortPath, not including the terminating null character.

I am comparing with other uses in Nuitka, minimizing the code diff right now.

bearsh:
> > If the function succeeds, the return value is the length, in TCHARs, of the string that is copied to lpszShortPath, not including the terminating null character.
> 
> I am comparing with other uses in Nuitka, minimizing the code diff right now.

but it than says:

> If the lpszShortPath buffer is too small to contain the path, the return value is the size of the buffer, in TCHARs, that is required to hold the path and the terminating null character.

kayhayen:
At least on MinGW64 it was crashing reproducibly and adding one extra character, fixed it. The crash was always there when run via test runner, and not there when run manually.

With malloc it's defaulting to `\0` for the buffer. I have had similar issues with `GetTempPathW` and MinGW64 in the past as well, and there I saw it's not putting termination in the buffer. This might be a difference in the libraries provided by MinGW and MSVC. I wonder if something unhealtly is going on there, but in this way it seems to work with both. It might be because, this code is not really coming from DLLs, but provided by the C compilers that they deviate.

Of course, using MinGW64 is not recommended anyway, but for some people, e.g. on Windows XP, which there are actually still people on, it's the only working choice, so I keep it alive.

But even the example code shows that you code should be perfectly working. Just me getting it to work with more things. :)

kayhayen:
And even with MSVC, there was this it seems: https://github.com/Nuitka/Nuitka/runs/3260140162?check_suite_focus=true

```
--- C:\hostedtoolcache\windows\Python\3.9.6\x64\python.exe (stdout)
+++ nuitka (stdout)
Error, results differed (stdout,exit_code).

@@ -1,2 +1 @@

-Hello Onefile World!
 
Exit codes 0 (CPython) != 3221226356 (Nuitka)
```

The exit code on Windows is garbage in case of a segfault. So I don't know what to make out of it. I know I didn't get away with that code myself.

bearsh:
therefor I try to avoid using windows :smile: 

anyway thanks for merging and fixing!

PR #1172, opened 2021-07-30
Add support for pkg_resources.require calls

CharlesB2:
# What does this PR do?

This PR adds support for `pkg_resources.require` calls in the `pkg-resources` plugin.

Since these calls are meant to raise an exception at runtime, my take is to raise the exception at Nuitka compilation if requirement isn't met, and suppress the statement if it is met.

Currently `pkg_resources.require` raises an exception when compiled by Nuitka, even with the `pkg-resources` plugin.

# PR Checklist

      cover enough.

### DIFF
--- a/nuitka/plugins/standard/PkgResourcesPlugin.py
+++ b/nuitka/plugins/standard/PkgResourcesPlugin.py
@@ -77,6 +77,20 @@ class NuitkaPluginResources(NuitkaPluginBase):
 
                 source_code = source_code.replace(match[0], value)
 
+            for match in re.findall(
+                r"""\b(pkg_resources\.require\(\s*['"](.*?)['"]\s*\))""",
+                source_code,
+            ):
+                # Explicitly call the require function at Nuitka compile, and
+                # if it fails remove it so that it doesn't fail at execution
+                try:
+                    self.pkg_resources.require(match[1])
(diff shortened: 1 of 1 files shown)


kayhayen:
I agree, or maybe even ignore it entirely, but definitely better than doing anything at runtime that is not working or wasteful.

nuitka/plugins/standard/PkgResourcesPlugin.py
```diff
@@ -77,6 +77,20 @@ def onModuleSourceCode(self, module_name, source_code):
...
+            for match in re.findall(
+                r"""\b(pkg_resources\.require\(\s*['"](.*?)['"]\s*\))""",
+                source_code,
+            ):
+                # Explicitly call the require function at Nuitka compile, and
+                # if it fails remove it so that it doesn't fail at execution
+                try:
+                    self.pkg_resources.require(match[1])
+                except self.pkg_resources.ResolutionError:
+                    raise self.pkg_resources.ResolutionError(
```
kayhayen, line 89: I wouild do a `self.sysexit()` style program termination here, but that's definitely OK.

kayhayen: I will add this is a cleanup commit myself.

PR #1196, opened 2021-09-01
Added option to use PGO on Nuitka-generated code.

Maxwell175:

# What does this PR do?

Performs PGO on the Nuitka-generated code. 

Greatly improved performance. 

# PR Checklist

      cover enough.

### DIFF
--- a/nuitka/MainControl.py
+++ b/nuitka/MainControl.py
@@ -25,6 +25,8 @@ a distribution folder.
 """
 
 import os
+import shlex
+import subprocess
 import sys
 
 from nuitka.build.DataComposerInterface import runDataComposer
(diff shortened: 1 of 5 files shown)


kayhayen:
Things that are not yet solved. I think this is only for non-Windows, and it won't work well for standalone mode at all, as I think data files won't be there yet, or post processing of the file done, and I don't really like that it's not possible to re-use existing profiling information. I am going to add stuff to disallow using it for modes that won't work and/or hide it from the options help temporarily.

kayhayen:
I am also making `--lto` an option that defaults to `auto`, and is implied with `--pgo` when given as `auto`, this way I am more sure it's used properly. Also allows to disable the LTO for MSVC, which sort of was missing. That was a TODO in my mind for a long time now. It will also enable Nuitka to hint at using --lto=yes or --lto=no dependent on your setup needs.

Still need to figure out, how to detect if LTO is going to work in advance, with different gcc versions bytecode versions, not sure how to detect them from files. But much like with `--static-libpython` it might be good enough to inform the user is missing out on something.

Issue #1011, opened 2021-03-03
[WIN10+MSYS2mingw64+PyGObject] ImportError: cannot import name '_gi' from 'gi'


maurocsound:
## Description of the issue
I'm on Windows 10 64bit. 
I would like to compile a python program (which uses GTK, so PyGObject package) into an exe. 
I correctly installed MSYS2 and updated the "MSYS2 MINGW 64-bit" environment as per instructions.

I installed nuitka with the following command:
```shell
$ pacman -S mingw-w64-x86_64-python-nuitka
```

Then I ran, in the dir where I have my python script named 'p.py'):
```shell
$ python -m nuitka --standalone --mingw64 p.py
```

It seems to go all right... There are no errors from the nuitka script. Now I have 2 dir: p.build and p.dist. I go into the p.dist subdir and when I try to run the generated exe I have the following error and the script doesn't run:

```shell
$ ./p.exe
Traceback (most recent call last):
  File "C:\msys64\home\Mauro\prova\P7B09~1.DIS\p.py", line 1, in <module>
  File "C:\msys64\home\Mauro\prova\P7B09~1.DIS\gi/__init__.py", line 40, in <module gi>
ImportError: cannot import name '_gi' from 'gi' (C:\msys64\home\Mauro\prova\P7B09~1.DIS\gi/__init__.py)

```

This is the content of the p.py script I tried to compile:
```
# p.py:
import gi

print("Hello!")
```

## Context information (for bug reports)
```
$ python -m nuitka --version
0.6.12.3
Python: 3.8.8 (default, Feb 20 2021, 07:16:03)  [GCC 10.2.0 64 bit (AMD64)]
Executable: C:/msys64/mingw64/bin/python.exe
OS: Windows
Arch: x86_64
```

maurocsound:
P.S.: I installed the pygobject package as per instructions from the official [docs](https://pygobject.readthedocs.io/en/latest/getting_started.html#windows-getting-started):

`
pacman -S mingw-w64-x86_64-gtk3 mingw-w64-x86_64-python3 mingw-w64-x86_64-python3-gobject
`

kayhayen:
Wow, really, this can only be pip installed on Windows, and then only with msys2, can you find out, what happens to this code:

```
        elif full_name == "gi._gi":
            gtk_dll_path = locateDLL("gtk-3")

            if gtk_dll_path is not None:
[...]

PR #1227, opened 2021-09-30
Add GI plugin to copy typelib files and set search path

zegervdv:
# What does this PR do?

Include typelib files for gi libraries and set up the correct search path.

Following from a discussion on #1011 

I marked it as draft because I still want to improve some things, but the proof of concept seems to work.
I verified with strace to see that the typelib files are loaded from the dist folder.

# TODO

The typelib files seem to be some kind of database format, but they are not actually libraries, so I opted for the `considerDataFiles` method. 
- [ ] Should I still use the DLL option?
- [ ] I hard-coded the path to the `girepository-1.0` folder, but I need to figure out how to detect that from the installation.
- [ ] Currently I am copying the whole folder, but I should add a check to only copy the matching typelibs for loaded modules.

# PR Checklist

      cover enough.

### DIFF
--- a/nuitka/plugins/standard/GiPlugin.py
+++ b/nuitka/plugins/standard/GiPlugin.py
@@ -1,0 +1,89 @@
+""" Support for gi typelib files
+"""
+import os
+
+from nuitka import Options
+from nuitka.plugins.PluginBase import NuitkaPluginBase
+from nuitka.freezer.IncludedDataFiles import makeIncludedDataDirectory
+
+
+def _isGiModule(module):
+    full_name = module.getFullName()
+    return full_name == "gi"
+
+
+class GiPlugin(NuitkaPluginBase):
+    plugin_name = "gi"
+    plugin_desc = "Support for GI dependencies"
+
+    @classmethod
+    def isRelevant(cls):
+        """This method is called one time only to check, whether the plugin might make sense at all.
+
+        Returns:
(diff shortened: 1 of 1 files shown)


zegervdv:
strace shows typelib files are correctly loaded from the included dist path:
```
openat(AT_FDCWD, "/home/zegervdv/Projects/gtk-test/__main__.dist/girepository-1.0/Gtk-3.0.typelib", O_RDONLY) = 3
```

kayhayen:
K  tbe path hardcoding, do we believe it's different anywhere yet, and it seems you could use like the last element of the run time detected path. The information query is cached, so you do not have to deal and can just use it everywhere. Looks very good to me already.

kayhayen:
Also, I guess, you would also get away with any path name you choose, it's only driven by the environment variable, so lets just not care, and remove version number alltogether?

zegervdv:
Yes, I figured out later you can get the path from the runtime query, so I believe it should work for all platforms.

I used the same name for the folder as in the installation, but I can remove the version from it.
I'll push an update when I get back to my PC.

zegervdv:
Renamed it to `girepository`.

The only improvement that still remains is to only load the .typelib files that are required, but that would probably require a runtime inspection of the application to be compiled.

kayhayen:
Very nice job, thanks a lot @zegervdv

PR #1243, opened 2021-10-12
Added ipcqueue dependency on _cffi_backend

mikehaben69:
This PR adds a single entry to the ImplicitImports plugin - the "ipcqueue" Python module (Posix and SysV message-queues) has a dependency upon _cffi_backend.


kayhayen:
If this works for CI, which I bet it will, I will directly put it to the hotfix.

kayhayen:
Part of the hotfix release.

PR #1264, opened 2021-10-31
Normalize paths before comparing

zegervdv:
# What does this PR do?

Normalize the paths of the downloaded gcc on Windows before comparing them.

When compiling on msys2 (mingw64) and after downloading the proposed gcc.exe I was getting an error that the path does not match.
It seems msys2 is very tolerant about using forward (/) or backward (\) slashes in paths, and in my case they were not matching up, causing a miscompare.
Using `os.path.normpath` normalizes these paths so they can be safely compared.

# PR Checklist

      cover enough.


kayhayen:
You should have used "areSamePaths" which will also abspath. I will do that in a cleanup myself.

Issue #1288, opened 2021-11-27
distutils deprecation


nschloe:
Nuitka still uses `distutils` in its own code. This should probably be removed at some point.
```
python3.10 -c "import distutils"
```
```
<string>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
````

**Comment by kayhayen** [comment]
Well, deprecation warnings are disabled in Nuitka, e.g. we use `optparse` which is deprecated since 3.2, and we support Python 2.6 still, so we are not too hot with it.

That said, `distutils` used in only one spot, and that is:

```
def copyTree(source_path, dest_path):
    """Copy whole directory tree, preserving attributes.

    Args:
        source_path: where to copy from
        dest_path: where to copy to, may already exist

    Notes:
        This must be used over `shutil.copytree` which has troubles
        with existing directories.
    """

    from distutils.dir_util import copy_tree

    return copy_tree(source_path, dest_path)
```

I am open to replacement code, esp. as it's not using the same tricks as `copyFileWithPermissions` for filesystems that will give errors in attempts to set ACL, which we do not need. So there would even be an immediate benefit. Long term, we are probably going to need a function that scans a directory, and then lets plugins mess with file contents, and the place files individually, maybe in a filesystem, maybe somewhere else, so this will become unused anyway. 

Tests use it for no good reason (the comment doesn't apply) to copy python files, and otherwise old plugins, which will be changed to data files, and data files, because it's bad. So maybe we will stop having it anyway.

**Comment by nschloe** [comment]
> That said, distutils used in only one spot,

Seems more than one:
```
git grep distutils nuitka/
```
```
nuitka/PythonFlavors.py:            from distutils.dir_util import _multiarch
nuitka/__main__.py:        # only lead to trouble, having e.g. a "distutils" in sys.path that comes
nuitka/__main__.py:        # from "nuitka.distutils".
nuitka/build/inline_copy/lib/scons-2.3.2/SCons/Defaults.py:from distutils.msvccompiler.
nuitka/build/inline_copy/lib/scons-3.1.2/SCons/Defaults.py:from distutils.msvccompiler.
nuitka/build/inline_copy/lib/scons-4.3.0/SCons/Defaults.py:from distutils.msvccompiler.
nuitka/distutils/DistutilCommands.py:""" Nuitka distutils integration.
[...]

PR #1290, opened 2021-11-27
add alternative for distutils.dir_util.copy_tree

nschloe:
Cf. #1288.

### DIFF
--- a/nuitka/utils/FileOperations.py
+++ b/nuitka/utils/FileOperations.py
@@ -567,11 +567,15 @@ def copyTree(source_path, dest_path):
         This must be used over `shutil.copytree` which has troubles
         with existing directories.
     """
-
-    # False alarm on travis, pylint: disable=I0021,import-error,no-name-in-module
-    from distutils.dir_util import copy_tree
-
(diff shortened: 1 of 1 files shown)


nuitka/utils/FileOperations.py
```diff
@@ -567,11 +567,15 @@ def copyTree(source_path, dest_path):
         This must be used over `shutil.copytree` which has troubles
         with existing directories.
     """
-
-    # False alarm on travis, pylint: disable=I0021,import-error,no-name-in-module
-    from distutils.dir_util import copy_tree
-
-    return copy_tree(source_path, dest_path)
+    try:
```
kayhayen, line 570: Can you please make this a if `python_version >= 0x380` which will make it more explicit, what version it's about, catching exceptions is not very reliable and causes ugly and confusing tracebacks about causes in case the except handler fails.

Otherwise this seems perfectly fine solution.

kayhayen:
The pylint error, I will take care of, I think I will just move the pylint part to 3.9 now

kayhayen:
Thanks, good job.

PR #1305, opened 2021-12-08
Optimization: avoid loop in generated code for args_count == 1

gdementen:
I noticed the code at https://github.com/Nuitka/Nuitka/blob/f4b08e086c2898013a08b74bb45b0ee01f04be8a/nuitka/build/static_src/HelpersCalling2.c#L358 has a useless loop and tried to generate simpler code for that case.

Honestly, I have no idea if C compilers are smart enough these days to detect this, so I don't know whether or not this change will have any net effect.

PS: This is against factory, but I can redo it against develop if you think it's worth it.

### DIFF
--- a/nuitka/codegen/templates/CodeTemplateCallsPositional.j2
+++ b/nuitka/codegen/templates/CodeTemplateCallsPositional.j2
@@ -44,7 +44,9 @@ PyObject *CALL_FUNCTION_WITH_ARGS{{args_count}}(PyObject *called, PyObject *cons
         PyObject *result;
 
         if (function->m_args_simple && {{args_count}} == function->m_args_positional_count){
-{% if args_count != 0 %}
+{% if args_count == 1 %}
+            Py_INCREF(args[0]);
+{% elif args_count > 1 %}
             for (Py_ssize_t i = 0; i < {{args_count}}; i++) {
                 Py_INCREF(args[i]);
             }
@@ -122,7 +124,10 @@ PyObject *CALL_FUNCTION_WITH_ARGS{{args_count}}(PyObject *called, PyObject *cons
                 python_pars[0] = method->m_object;
                 Py_INCREF(method->m_object);
 
-{% if args_count != 0 %}
+{% if args_count == 1 %}
+                python_pars[1] = args[0];
+                Py_INCREF(args[0]);
+{% elif args_count > 1 %}
                 for (Py_ssize_t i = 0; i < {{args_count}}; i++) {
                     python_pars[i + 1] = args[i];
                     Py_INCREF(args[i]);


kayhayen:
Good job, I like cleaner code, even if generated. I am pretty sure that the compilers optimize it away.

kayhayen:
Bonus points if you manage to make a Jinja macro out of it though. I am searching a website where generated code can be checked.

kayhayen:
Wow, I got evidence they do not!

kayhayen:
Check this out: https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:15,endLineNumber:6,positionColumn:15,positionLineNumber:6,selectionStartColumn:15,selectionStartLineNumber:6,startColumn:15,startLineNumber:6),source:'//+Type+your+code+here,+or+load+an+example.%0Aint+square(int+num)+%7B%0A++++for+(int+i+%3D+0%3B+i+%3C+0%3B+i%2B%2B)+%7B%0A++++++++num+%3D+num*2%3B%0A++++%7D%0A++++return+num%3B%0A%7D'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cg112,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'',selection:(endColumn:12,endLineNumber:15,positionColumn:12,positionLineNumber:15,selectionStartColumn:12,selectionStartLineNumber:15,startColumn:12,startLineNumber:15),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+11.2+(C,+Editor+%231,+Compiler+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

kayhayen:
Naj, it's OK, with -O2 the gcc does do it.

kayhayen:
I played a bit around and the oldest gcc available does it, as does MSVC, but only if optimization options are on.

gdementen:
> Bonus points if you manage to make a Jinja macro out of it though. 

Sorry, I don't understand how that change could be made into a macro. And this is probably beyond my Jinja skill anyway 😉

kayhayen:
Checkout the change I pushed with regards to `CHECK_OBJECTS`, macros are fairly easy, but the call code didn't use any so far, and that's mostly because of inconsistencies, that it would have been ugly to add. Now that code generation more often uses Jinja2 templates, obviously there is a need to not repeat standard things too often.

Issue #561, opened 2019-10-27
Issue with Trio and Nuitka


ItWasAllIntended:
Nuitka appears to be incompatible with the latest version of Trio. As far as I can tell, this is unrelated to #410. This happens with all Trio programs I've tested, the simplest being:
```py
import trio

async def main():
    print("Hello")

trio.run(main)
```
This runs without issue regularly but produces the traceback
```Traceback (most recent call last):
  File "C:\Users\sethw\PycharmProjects\PyCom\venv\lib\site-packages\trio\_core\_run.py", line 1769, in run
    run_impl(runner, async_fn, args)
  File "C:\Users\sethw\PycharmProjects\PyCom\venv\lib\site-packages\trio\_core\_run.py", line 1803, in run_impl
    runner.init_task = runner.spawn_impl(
  File "C:\Users\sethw\PycharmProjects\PyCom\venv\lib\site-packages\trio\_core\_run.py", line 1351, in spawn_impl
    coro.cr_frame.f_locals.setdefault(
AttributeError: 'NoneType' object has no attribute 'f_locals'
```
I've tested this on Python 3.7 & 3.8 and the latest Nuitka from PyPi & an install from the `factory` branch, all of which produce the same traceback

JorjMcKie:
@kayhayen - I am not looking into this more closely.

The reason why this does not yet work is here:

```
def f(x):
    print(x)


async def make(f):
    return await f


af = make(f)

print("af.frame =", af.cr_frame)
```
Running this interpreted yields
```
af.frame = <frame at 0x00000241463E2A28, file 'test.py', line 5, code make>
sys:1: RuntimeWarning: coroutine 'make' was never awaited
```

Whereas in standalone we get:
```
af.frame = None
```
This makes trio fail.

kayhayen:
Nuitka delays frame creation until usage. There may not ever be a usage after all. Therefore it is `None`. We could in theory become good enough to create it once its looked at though. That is of dubious help though.

Next the usage of `.setdefaults` on `f_locals` which might be a write (not seeing what value though) is not going to impress it a lot either. We could have a look at the source code, and what it tries to do there, and possibly engage with upstream there. If it expects to have an impact on program behavior that will be bad.

Integrating with a locals dictionary is not going to fly well. But I am also confused, because I think that Python3 should not be using that locals dictionary in the normal case when written to, but maybe I misremember, so maybe it's nothing but unnecessary to do any write.

[...]

PR #1352, opened 2022-01-02
Add Trio plugin

JamesOldfield:
Adds a plug-in to work around incompatibility between Nuitka and Trio's use of `coro.cr_frame.f_locals` for interrupt handling (#561 and python-trio/trio#1752).

The plug-in works by patching Trio's source code to simply remove that handling. Ctrl+C will still work (at least, there's no reason it shouldn't), but if Trio is running in the main thread then it will be handled less gracefully than without the patch. But at least it will run at all, which it didn't in Nuitka without this patch.

The fact that this is a source code patch means that, unfortunately, it is likely to be fairly fragile to changes in Trio's source code in future versions.

This is based on the "develop" branch, and follows the coding conventions (I ran `autoformat-nuitka-source` and checked with `check-nuitka-with-pylint`). I won't take any offense at any changes, feel free to totally rewrite it! This is just to show what works (for me).

### DIFF
--- a/nuitka/plugins/standard/trio_plugin.py
+++ b/nuitka/plugins/standard/trio_plugin.py
@@ -1,0 +1,66 @@
+"""Trio plugin module."""
+
+from nuitka.plugins.PluginBase import NuitkaPluginBase
+
+_trio_patches = {
+    "trio._core._ki": (
+        """\
+enable_ki_protection = _ki_protection_decorator(True)  # type: Callable[[F], F]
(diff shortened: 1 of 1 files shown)


kayhayen:
Looks good to me, except for using "not x in y", prefer "x not in y", it's a nicer read. I am surprised pylint doesn't complain about that. For plugins there is a `self.sysexit` with nicer looking error message.

And yes, I take it as a great input. I will merge it for the next release, but have a TODO to integrate it into the anti-bloat plugin, one that one becomes default.

PR #1424, opened 2022-02-15
WIP: Allow switching compression algorithms for generated AppImages

squidink7:
# What does this PR do?
Adds a flag to set different supported compression algorithms for generating AppImages on Linux (currently only gzip and xz are supported)

xz is currently the default, however it takes a long time to extract and offers little size benefits over gzip in my testing. However rather than replacing it entirely this patch allows the user to manually choose which one they want (although it defaults to gzip as per appimagetool's default)

# PR Checklist

- [ ✓ ] Correct base branch selected? Should be `develop` branch.
- [ ✓ ] All tests still pass. Check the Developer Manual about `Running the Tests`.
      cover enough.
- [ X ] Ideally new features or fixed regressions ought to be covered via new tests.
- [ X ] Ideally new or changed features have documentation updates.

Note: Don't merge yet! I haven't tested it yet as it's 11:00 on a school night and I really should be getting some sleep. I'll update this once I have done some testing and confirmed it works.

### DIFF
--- a/nuitka/OptionParsing.py
+++ b/nuitka/OptionParsing.py
@@ -1267,6 +1267,15 @@ linux_group.add_option(
     help="Add executable icon for onefile binary to use. Can be given only one time. Defaults to Python icon if available.",
 )
 
+linux_group.add_option(
+    "--linux-onefile-compression",
+    action="store",
+    dest="appimage_compression",
+    metavar="COMPRESSION",
+    default="gzip",
+    help="Compression method to use for Linux onefile builds. gzip (default): larger file but faster startup, xz: smaller file but slower startup."
+)
+
 parser.add_option_group(linux_group)
 
 plugin_group = OptionGroup(parser, "Plugin control")
--- a/nuitka/OptionParsing.py
+++ b/nuitka/OptionParsing.py
@@ -1271,9 +1271,10 @@ linux_group.add_option(
     "--linux-onefile-compression",
     action="store",
     dest="appimage_compression",
+    choices=("gzip", "xz"),
     metavar="COMPRESSION",
     default="gzip",
-    help="Compression method to use for Linux onefile builds. gzip (default): larger file but faster startup, xz: smaller file but slower startup."
+    help="Compression method to use for Linux onefile builds. Defaults to gzip for faster decompression"
 )
 
 parser.add_option_group(linux_group)
--- a/nuitka/freezer/Onefile.py
+++ b/nuitka/freezer/Onefile.py
@@ -170,26 +170,22 @@ Categories=Utility;"""
     stdout_file = openTextFile(stdout_filename, "wb")
     stderr_file = openTextFile(stderr_filename, "wb")
 
-    if options.appimage_compression == "xz" or options.appimage_compression == "gzip":
-        command = (
-            _getAppImageToolPath(
-                for_operation=True, assume_yes_for_downloads=assumeYesForDownloads()
-            ),
-            dist_dir,
-            "--comp",
-            options.appimage_compression,
-            "-n",
-            onefile_output_filename,
-        )
+    if options.appimage_compression == "gzip":
+        compress_flags = "--comp gzip"
+    elif options.appimage_compression == "xz":
+        compress_flags = "--comp xz"
     else:
-        command = (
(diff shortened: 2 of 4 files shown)


kayhayen:
You will need to do `git rebase upstream/develop` and `git rebase --skip` for all conflicting commits.

squidink7:
Ok, I've confirmed this to work with my 1 test case, although I would highly recommend you look over it to make sure I haven't made any massive mistakes (small PR, only 2 files changed but still, I'm far from experienced at this). Also I haven't written any documentation (other than the option's help text) or tests so let me know if I should do any of that.

nuitka/freezer/Onefile.py
```diff
@@ -170,13 +170,19 @@ def packDistFolderToOnefileLinux(onefile_output_filename, dist_dir, binary_filen
     stdout_file = openTextFile(stdout_filename, "wb")
     stderr_file = openTextFile(stderr_filename, "wb")
 
+    if options.appimage_compression == "gzip":
```
kayhayen, line 173: For options, we add an accessor function, like `getAppImageCompression()` for use here.

nuitka/OptionParsing.py
```diff
@@ -1267,6 +1267,15 @@ def _process_long_opt(self, rargs, values):
     help="Add executable icon for onefile binary to use. Can be given only one time. Defaults to Python icon if available.",
 )
 
+linux_group.add_option(
+    "--linux-onefile-compression",
+    action="store",
+    dest="appimage_compression",
```
kayhayen, line 1273: There is something for choices, that will allow the option parsing to check, `choices=("original", "runtime", "frozen")` is used elsewhere in the file.

nuitka/freezer/Onefile.py
kayhayen, line 173: One of the benefits of having a choice, is that you get to not check here anymore, you can just use it here, not even with a variable at all.

I think `subprocess` and friends would hate having one string with a space or an empty value (which you don't need anymore), they would be passed in `argv` as a value, and not e.g. be ignoring for empty string, or become two entries, so actually AppImage ought to error out with this code, but if it doesn't, it surely is not generally possible.

So  `"--comp,getAppImageCompression(),` would be good there I suppose.

kayhayen (changes requested):
Good job, don't worry about there being findings. PRs are always a collaboration and there is no way of knowing everything about how a project functions.

nuitka/freezer/Onefile.py
squidink7: I added one in Options.py, I take it that was the correct location?

nuitka/OptionParsing.py
squidink7: Thanks for the tip!

nuitka/freezer/Onefile.py
squidink7: I've implemented this change, could you please confirm I haven't missed anything? thanks a bunch!
And you're right, appimagetool was giving me a weird error with that code (strange because it worked the first time then never again)

kayhayen (approved):

kayhayen:
Looks perforce to me. I recently added sorcery, which is not yet fully configured to make sense.

squidink7:
Look like the python 3.9 check failed. I ran bin/autoformat-nuitka-source.py but it complained about me having too new versions of a couple of modules installed, looking for 3.9 versions (Arch updated its python to 3.10 a while back). Should I try to install the old version and run this file again? if so I'll do it when I next have some free time.

kayhayen:
I am usually updating the autoformat requirements on a regular basis, but it may be a few weeks now. Usually installing `requirements-devel.txt` with `-U` option will make it work, but I can do it for you and merge this, no problem.

kayhayen:
Ah, it's on 3.9 job, because I am not making the autoformat check for every Python version the Github uses the same things, but the Python version play no role. It would also fail with 3.7, and you are relatively free to develop with 3.6 or higher, at least I believe that is what every tool used there still supports, but it's increasing over time naturally. But upwards often works. One exception is that some compiled packages often do not work with something like 3.10 after release, or are buggy with them, so I am not too eager to bump it, but might be time for that now.

PR #1429, opened 2022-02-17
adding param allowing to overwrite existing directories

hirschrobert:
setup: qemu vm win10 with python 3.8.10 and nuitka 0.6.19.7

When building a binary with `--onefile` option python error

> FileExistsError: [WinError 183]

came up. with every retry, other files were tried to be overwritten. Changing the second parameter of `os.makedirs()` fixed it (see [python docs](https://docs.python.org/3/library/os.html#os.makedirs)). But I don't know if the files that are overwritten now are needed elsewhere. `os.path.isdir()` from line before seems to "fail" in this case. maybe race conditions, when in line 166, condition of 165 is not true anymore?.

### DIFF
--- a/nuitka/utils/FileOperations.py
+++ b/nuitka/utils/FileOperations.py
@@ -163,7 +163,7 @@ def makePath(path):
 
     with withFileLock("creating directory %s" % path):
         if not os.path.isdir(path):
-            os.makedirs(path)
+            os.makedirs(path, exist_ok=True)
 
 
 def isPathExecutable(path):


kayhayen:
The change is Python3 only code, but what's more, it's also bug hiding. Something must be afoul there. The lock is supposed to make sure, that Nuitka cannot have races with file operations. But maybe it's a file and not a directory for some reason. Did you raise an issue with the original phenomenon?

I would much rather try to understand why Nuitka is creating a directory named like an existing file, and have it stop doing that.

kayhayen:
This should be an issue report, rather than a PR that I cannot merge. The file locking might be missing somewhere, but I cannot tell from this "report". Please open an issue.

PR #1504, opened 2022-04-01
Small stylistic changes

ralienpp:
While going through the doc in search of a solution for my problem, I stumbled upon a few typos.

### DIFF
--- a/Developer_Manual.rst
+++ b/Developer_Manual.rst
@@ -1216,7 +1216,7 @@ The solution is this:
 Frame Stack
 ===========
 
-In Python, every function, class, and module has a frame. It creates
+In Python, every function, class, and module has a frame. It is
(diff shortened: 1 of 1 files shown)


Developer_Manual.rst
```diff
@@ -4590,7 +4590,7 @@ dropping non interesting, or rewriting exceptions.
 Caching of Python level compilation
 ===================================
 
-While the C compilation result is already cached with ccache and friends
+While the C compilation result is already cached with `ccache` and friends
```
kayhayen, line 4593: The quoting in rest should be double back-ticks, it's easy to make that mistake as it's different from markdown, otherwise it's good.

kayhayen:
I will just go ahead and merge it despite my comments, and amend accordingly.

PR #1567, opened 2022-05-19
merge plugin yaml and create schema + dropbox-sdk-python fix

Fire-Cube:
- merge anti-bloat.yml, data-files.yml, dll-files.yml, implicit-imports.yml to config.yml
- create schema for config.yml
- add fix for dropbox-sdk-python

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -6,9 +6,9 @@ In this document, we track the per version changes and comments. This
 becomes a document on the website, as well as individual posts on the
 Nuitka blog.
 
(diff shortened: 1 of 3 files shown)


nuitka/plugins/standard/config.yml
```diff
@@ -0,0 +1,1088 @@
+- "module-name": "boto3"
```
kayhayen, line 1: Can we please not quote keywords, quotes sjhould only be applied to string values.

PR #1571, opened 2022-05-21
Fix typo in warning message

kamgha:
The module in question is `orderedset`, see: https://github.com/Nuitka/Nuitka/blob/1adf285ad2416876aabe7c474ce9f1ace2344ed9/requirements.txt#L27


kayhayen:
I will put it on the next hotfix too, kind of important, we want people to paste the names.

PR #1616, opened 2022-06-20
pyreadstat_implicit_dep

gdementen:
This is a follow-up to #1072

The strange thing is that this was already required when I did the #1072 PR (it was introduced in pyreadstat 1.0.3), so I suppose something else added the multiprocessing module implicitly?


kayhayen:
It used to be that Nuitka included all of stdlib, i.e. multiprocessing as well, but it started to include parts of it only, when they are actually used. That can explain it.

kayhayen:
Nuitka is taking a risk by not forcing programs to include these, hope is that C modules very rarely use multiprocessing, but that was wrong in this case at least.

PR #1627, opened 2022-06-26
wip: Added the methods for list.

Cjw9000-py:

# What does this PR do?

- Fixed a typo that i saw.
- Added the methods for list in the specialize tool, and added some unfinished implementation for it, in nodes.

- The file ListNodes.py is unfinished.

# PR Checklist

- [X] Correct base branch selected? Should be `develop` branch.
- [X] Enabled commit hook or executed `./bin/autoformat-nuitka-source`.
- [X] All tests still pass. Check the Developer Manual about `Running the Tests`.
      cover enough.
- [ ] Ideally new features or fixed regressions ought to be covered via new tests.  # code unfinished so no tests
- [x] Ideally new or changed features have documentation updates.  # ~

### DIFF
--- a/nuitka/nodes/ListNodes.py
+++ b/nuitka/nodes/ListNodes.py
@@ -61,7 +61,11 @@ class ExpressionListOperationAppend(
         list_arg = self.subnode_list_arg
         item = self.subnode_item
 
+        # the state of the object changes
         trace_collection.removeKnowledge(list_arg)
+
+        # exposing the variable content
+        item.onContentEscapes(trace_collection)
 
         return self, None, None
 
@@ -87,13 +91,14 @@ class ExpressionListOperationClear(
     def computeExpression(self, trace_collection):
         list_arg = self.subnode_list_arg
 
+        # only the state of object changes
         trace_collection.removeKnowledge(list_arg)
 
         return self, None, None
 
 
 class ExpressionListOperationCopy(
-    ExpressionListShapeExactMixin, ExpressionChildHavingBase
+    SideEffectsFromChildrenMixin, ExpressionChildHavingBase
 ):
     """This operation represents l.copy()."""
 
@@ -111,6 +116,9 @@ class ExpressionListOperationCopy(
         )
 
     def computeExpression(self, trace_collection):
+        list_arg = self.subnode_list_arg
+
+        list_arg.onContentEscapes(trace_collection)
         return self, None, None
 
 
@@ -120,7 +128,6 @@ class ExpressionListOperationExtend(
     """This operation represents l.extend(iterable)."""
 
     kind = "EXPRESSION_LIST_OPERATION_EXTEND"
-
     name_children = ('list_arg', 'value')
 
     def __init__(self, list_arg, value, source_ref):
@@ -137,10 +144,14 @@ class ExpressionListOperationExtend(
         list_arg = self.subnode_list_arg
         value = self.subnode_value
 
+        # the content of the list changes
         trace_collection.removeKnowledge(list_arg)
 
-        trace_collection.onExceptionRaiseExit(BaseException)
+        # expose the input value
+        value.onContentEscapes(trace_collection)
 
+        # if the argument is no iterable
+        trace_collection.onExceptionRaiseExit(BaseException)
         return self, None, None
 
 
@@ -150,7 +161,6 @@ class ExpressionListOperationIndex(
     """This operation represents l.index(value, start, stop)."""
 
     kind = "EXPRESSION_LIST_OPERATION_INDEX"
-
     name_children = ('list_arg', 'value', 'start', 'stop')
 
     def __init__(self, list_arg, value, start, stop, source_ref):
@@ -168,10 +178,10 @@ class ExpressionListOperationIndex(
         list_arg = self.subnode_list_arg
         value = self.subnode_value
 
-        trace_collection.removeKnowledge(list_arg)
+        value.onContentEscapes(trace_collection)
 
+        # raises ValueError when the value is not found
         trace_collection.onExceptionRaiseExit(BaseException)
-
         return self, None, None
 
 
@@ -181,7 +191,6 @@ class ExpressionListOperationInsert(
     """This operation represents l.insert(index, item)."""
 
     kind = "EXPRESSION_LIST_OPERATION_INSERT"
-
     name_children = ('list_arg', 'index', 'item')
 
     def __init__(self, list_arg, index, item, source_ref):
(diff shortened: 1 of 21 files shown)


nuitka/tools/specialize/templates_python/AttributeNodeFixed.py.j2
```diff
@@ -412,3 +425,97 @@ class ExpressionAttributeLookupBytes{{attribute_name_class}}(SideEffectsFromChil
...
+
+    def computeExpression(self, trace_collection):
+        return self, None, None
+
+{% if list_operation_version_presence %}
+    @staticmethod
+    def _computeExpressionCall(call_node, list_arg, trace_collection):
+        def wrapExpressionListOperation{{attribute_name_class}}({{formatArgs(list_operation_args, starting=True, finishing=False)}} source_ref):
+{% if python3_operation_name %}
+            if str is list:
```
gdementen, line 465: I do not think this test makes sense. It seems like a bad copy-paste/bad replace to me. AFAIK, the original test (str is bytes) is just a way to test for Python3 vs Python2 but "str is list" will always be False, right?

Cjw9000-py: Yes, gonna change that.
I replaced bytes with list, and didn't notice that.

nuitka/nodes/ListNodes.py
```diff
@@ -0,0 +1,317 @@
...
+
+        ExpressionChildHavingBase.__init__(
+            self,
+            value={'list_arg': list_arg, 'item': item},
+            source_ref=source_ref,
+        )
+
+    def computeExpression(self, trace_collection):
+        list_arg = self.subnode_list_arg
+        item = self.subnode_item
```
kayhayen, line 64: You should also do this:

```
            # They take over the variable content, exposing it to currently untraced usages.
            item.onContentEscapes(trace_collection)
```

Through being appended to the list, we loose track of values, and should take into account, that other places get access to it.

nuitka/nodes/ListNodes.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def computeExpression(self, trace_collection):
+        list_arg = self.subnode_list_arg
+
+        trace_collection.removeKnowledge(list_arg)
+
+        return self, None, None
+
+
+class ExpressionListOperationCopy(
```
kayhayen, line 97: You ought to inherit `SideEffectsFromChildrenMixin` for this one. It will implement a few methods for cases, where something has no effect other than producing the children values. The coping of the list has no effect by itself if not used.

But also call the onContentEscapes on the list value in `computeExpression` I guess.

I think like the dictionary copy which uses that mixin btw, we ought to also annotate that it cannot fail, and provide the equivalent of 

```
    def mayRaiseException(self, exception_type):
        return self.subnode_dict_arg.mayRaiseException(exception_type)
```

Notice, I didn't have these so far, but 0.9.2 hotfix corrects that. So far it didn't consider if the dictionary value itself raised an exception, very wrong. You probably also need that everywhere. Later we can add `mayRaiseExceptionOperation()` to optimize the actual operation to not falsely assume exceptions, but that is not too urgent now.

nuitka/nodes/ListNodes.py
```diff
@@ -0,0 +1,317 @@
...
+            source_ref=source_ref,
+        )
+
+    def computeExpression(self, trace_collection):
+        list_arg = self.subnode_list_arg
+        value = self.subnode_value
+
+        trace_collection.removeKnowledge(list_arg)
+
+        trace_collection.onExceptionRaiseExit(BaseException)
```
kayhayen, line 144: With the `mayRaiseException` as described above, we could remove it, but `extend` is actually one of the cases, where `MemoryError` while be worth considering. Unlike `append` which never allocates much memory really, this could easily overflow.

nuitka/nodes/ListNodes.py
```diff
@@ -0,0 +1,317 @@
...
+
+        return self, None, None
+
+
+class ExpressionListOperationIndex(
```
kayhayen, line 149: This has `int` result shape. It's what an operation returns. Most operations on containers return `None` because they are pseudo-expressions, but not all.

nuitka/nodes/ListNodes.py
```diff
@@ -0,0 +1,317 @@
...
+    def computeExpression(self, trace_collection):
+        list_arg = self.subnode_list_arg
+
+        trace_collection.removeKnowledge(list_arg)
+        trace_collection.onExceptionRaiseExit(BaseException)
+
+        return self, None, None
+
+
+class ExpressionListOperationReverse(
```
kayhayen, line 266: I am not annotating all that should be None, but this one too returns None

nuitka/specs/BuiltinListOperationSpecs.py
```diff
@@ -0,0 +1,104 @@
...
+            pos_only_args=(),
+            kw_only_args=(),
+        )
+
+
+# Python3 only
+list_clear_spec = ListMethodSpecNoKeywords("clear", arg_names=())
+
+list_append_spec = ListMethodSpecNoKeywords("append", arg_names=("object",))
+list_copy_spec = ListMethodSpecNoKeywords("copy", arg_names=())
```
kayhayen, line 75: No keywords seems strange when where are no arguments. I would use `ListMethodSpec` in that case.

nuitka/nodes/ListNodes.py
```diff
@@ -0,0 +1,317 @@
...
+    ExpressionStrOrUnicodeExactMixin,
+    ExpressionStrShapeExactMixin,
+    ExpressionTupleShapeExactMixin,
+)
+from .NodeBases import SideEffectsFromChildrenMixin
+from .NodeMetaClasses import NodeCheckMetaClass
+
+
+class ExpressionListOperationAppend(
+    ExpressionListShapeExactMixin, ExpressionChildHavingBase
```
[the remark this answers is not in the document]
Cjw9000-py: So i should remove `ExpressionListShapeExtractMixin`?

kayhayen:
If you look for tests, checkout the `tests/generated/DictMethodsTest.py.j2` file, it does call all methods that have specs, and there are a bunch of conditions, that allows you to specific the arguments to test with. You do not have to cover wrong argument counts, that is generic code, but wrong types used (even possible with list, maybe `extend` comes to mind with a non-iterable), and it produces a `.py` file when you execute `tests/generated/run_all.py only DictMethodsTest` that you can look at.

kayhayen:
Any chance for progress on this one? :-)

Cjw9000-py:
Sorry i was pretty busy lately. I dont know if i will find much time in the near future. I hope the code can help at least a little.

kayhayen:
@Cjw9000-py no worries, thanks for letting me know, I will try and salvage what we can from here, this is a big part, and @Fire-Cube is currently working on something that help us generate nodes or their bases at least automatically, then this will be useful

kayhayen:
This didn't make it for 1.0 nor 1.1, but I am hoping to merge it for 1.2 soon, even if maybe commented out until we got that generator for node classes.

kayhayen:
I have just merged this and am currently adapting it to the new base class generation from specs, that will eliminate the boilerplate code somewhat. Thanks a lot for taking it on, very helpful.

PR #1632, opened 2022-06-27
structure changes for yaml

Fire-Cube:


misc/nuitka-package-config-schema.json
```diff
@@ -9,42 +9,38 @@
...
+                    },
+                    "patterns": {
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        }
+                    },
+                    "control_tags": {
+                        "type": "string",
+                        "pattern": "(([a-zA-Z]|[0-9]|_)|and|is|not| )"
```
kayhayen, line 32: we should also have `or` here.

misc/nuitka-package-config-schema.json
```diff
@@ -105,10 +100,9 @@
                             }
                         },
                         "control_tags": {
-                            "type": "array",
-                            "items": {
-                                "type": "string"
-                            }
+                            "type": "string",
+                            "pattern": "(([a-zA-Z]|[0-9]|_)|and|is|not| )"
```
kayhayen, line 104: Does the schema format allow to define this and use it later somehow, this is repeating itself a lot otherwise.

misc/nuitka-package-config-schema.json
```diff
@@ -132,37 +127,66 @@
...
                                 },
-                                {
-                                    "const": "recommend"
+                                "control_tags": {
+                                    "type": "string",
+                                    "pattern": "(([a-zA-Z]|[0-9]|_)|and|is|not| )"
                                 },
-                                {
-                                    "const": "recommend_on_macos"
+                                "description": {
```
kayhayen, line 186: maybe we have description always first?

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -245,8 +245,7 @@
   data-files:
     patterns:
       - "Lorem ipsum.txt"
-    control_tags:
-      - anti-bloat
+    control_tags: anti-bloat
```
kayhayen, line 248: This one is not quoted, while others are. I suggest quotes for all.

kayhayen (changes requested):

misc/nuitka-package-config-schema.json
Fire-Cube: I think I can do that.

misc/nuitka-package-config-schema.json
gdementen: also, shouldn't the "node" be called "control_tag" since there can only be one now?

Fire-Cube: I have deleted this regex. It becomes to complex.

kayhayen:
I sort of agree with `control_tags` being a bad name now, maybe we change it to `when`, which expressed more clearly that it has become a condition. But to not further complate it for you, I would delay this until a next iteration, or when I merge. It will be a simple search and replace, but it will make the diff even harder to read for resuting file. Right now lets now make more changes than absolutely necessary now. Getting the options evacuated out of implicit imports is a main goal now.

nuitka/tools/quality/auto_format/YamlFormatter.py
```diff
@@ -355,15 +366,28 @@ def formatYaml(path):
...
         if "implicit-imports" in sorted_entry:
-            sorted_entry["implicit-imports"] = dict(
-                OrderedDict(
-                    [
-                        (key, entry["implicit-imports"][key])
-                        for key in IMPLICIT_IMPORTS_KEYS
-                        if key in entry["implicit-imports"].keys()
-                    ]
+            for sub_entry in sorted_entry["implicit-imports"]:
+                sub_entry = dict(
```
kayhayen, line 370: The `dict(OrderedDict` calls look strange, is this because Yaml will not take them back unless as dict

kayhayen:
Habe ich auf factory gemerged. Sieht gut aus, ich habe die checks in eigene Funktionen ausgelagert, was man wohl immer schon hätte machen sollen. Muss ich natürlich jetzt testen, ob das alles passt.

PR #1653, opened 2022-07-04
check for unknown tags

Fire-Cube:
and add recommendation for jinja2 extension

### DIFF
--- a/nuitka/plugins/PluginBase.py
+++ b/nuitka/plugins/PluginBase.py
@@ -32,6 +32,8 @@ import inspect
 import sys
 from collections import namedtuple
 
+from ordered_set import OrderedSet
+
 from nuitka import Tracing
 from nuitka.__past__ import getMetaClassBase
 from nuitka.freezer.IncludedDataFiles import (
@@ -99,6 +101,7 @@ class NuitkaPluginBase(getMetaClassBase("Plugin")):
     # For working with options, user plugins must set this variable to
     # the script's path (use __file__, __module__ or __name__).
     plugin_name = None
+    used_tags = OrderedSet()
 
     @staticmethod
     def isAlwaysEnabled():
@@ -859,8 +862,7 @@ except ImportError:
         # Virtual method, pylint: disable=no-self-use,unused-argument
         return ()
 
-    @staticmethod
-    def evaluateControlTags(full_name, control_tags):
+    def evaluateControlTags(self, full_name, control_tags):
         # Note: Caching makes no sense yet, this should all be very fast and
         # cache themselves. TODO: Allow plugins to contribute their own control
         # tag values during creation and during certain actions.
@@ -885,7 +887,9 @@ except ImportError:
             context["before_python" + big + major] = is_lower_version
 
         # We trust the yaml files, pylint: disable=eval-used
-        return eval(control_tags, context)
+        result = eval(control_tags, context)
+        NuitkaPluginBase.used_tags.add(control_tags, result, self.plugin_name)
+        return result
 
     @classmethod
     def warning(cls, message):
--- a/nuitka/plugins/PluginBase.py
+++ b/nuitka/plugins/PluginBase.py
@@ -919,23 +919,18 @@ def standalone_only(func):
 class TagContext(dict):
     def __init__(self, full_name, *args, **kwargs):
         self.full_name = full_name
-        super().__init__(*args, **kwargs)
+        dict.__init__(*args, **kwargs)
 
     def __getitem__(self, key):
         try:
-            dict.__getitem__(self, key)
+            return dict.__getitem__(self, key)
 
         except KeyError:
             if key.startswith("use_"):
-                Tracing.plugins_logger.info(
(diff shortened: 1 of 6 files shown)


kayhayen:
The extension I would have to try, name suggests HTML, but something that at least marks up Jinja braces would be very welcome indeed.

nuitka/plugins/PluginBase.py
```diff
@@ -910,3 +914,28 @@ def wrapped(*args, **kwargs):
...
+
+
+class TagContext(dict):
+    def __init__(self, full_name, *args, **kwargs):
+        self.full_name = full_name
+        super().__init__(*args, **kwargs)
+
+    def __getitem__(self, key):
+        try:
+            dict.__getitem__(self, key)
```
kayhayen, line 926: Missing return here.

nuitka/plugins/PluginBase.py
```diff
@@ -910,3 +914,28 @@ def wrapped(*args, **kwargs):
...
+
+        except KeyError:
+            if key.startswith("use_"):
+                Tracing.plugins_logger.info(
```
kayhayen, line 930: One advantage of being in every plugin is that we can just `self.info` which will then give the plugin name.

nuitka/plugins/PluginBase.py
```diff
@@ -910,3 +914,28 @@ def wrapped(*args, **kwargs):
                 return None
 
     return wrapped
+
+
+class TagContext(dict):
+    def __init__(self, full_name, *args, **kwargs):
+        self.full_name = full_name
+        super().__init__(*args, **kwargs)
```
kayhayen, line 922: Since `super()` works badly for older Python, I avoid it always, just write `dict.__init__` and be done with it. Yes not pretty, but for portable Python you need to pass arguments, that petty much render it useless to use.

nuitka/plugins/PluginBase.py
```diff
@@ -910,3 +914,28 @@ def wrapped(*args, **kwargs):
...
+
+        except KeyError:
+            if key.startswith("use_"):
+                Tracing.plugins_logger.info(
+                    "Use default 'False' for not existing tag '" + key + "'."
+                )
+                return False
+
+            Tracing.plugins_logger.sysexit(
+                "Tag '"
```
kayhayen, line 936: Please use `%` format, it is so much more readable. `"Tag '%s' in module ..." % (tag, ...)` is what all of Nuitka uses. I played with `.format` too, and would love f-strings, but it's a good compromise in my book for not too large messages.

kayhayen (changes requested):
Pretty much there, good job.

nuitka/plugins/PluginBase.py
Fire-Cube: This isn't in a plugin.

kayhayen (approved):
Looks perfect to me, thanks for the good work, this will enable tags to become super useful.

nuitka/plugins/PluginBase.py
```diff
@@ -32,6 +32,9 @@
 import sys
 from collections import namedtuple
 
+from ordered_set import OrderedSet
```
kayhayen, line 35: Cannot do it directly, 3.6 and 3.10 cannot use the same `OrderedSet`, but there is `nuitka.containers.oset` that you can use.

nuitka/plugins/PluginBase.py
```diff
@@ -910,3 +918,23 @@ def wrapped(*args, **kwargs):
...
+    def __getitem__(self, key):
+        try:
+            return dict.__getitem__(self, key)
+
+        except KeyError:
+            if key.startswith("use_"):
+                self.info("Use default 'False' for not existing tag '%s'." % key)
+                return False
+
+            Tracing.plugins_logger.sysexit(
```
kayhayen, line 937: This should be `self.sysexit` so the using plugin becomes part of the trace automatically.

nuitka/plugins/standard/ImplicitImports.py
```diff
@@ -97,7 +97,9 @@ def _getImportsByFullname(self, module, full_name):
         if config:
             for entry in config:
                 if entry.get("control_tags"):
-                    if not self.evaluateControlTags(entry.get("control_tags")):
+                    if not self.evaluateControlTags(
+                        full_name, entry.get("control_tags")
```
kayhayen, line 101: Once it's more than one argument, I prefer to pass as keyword arguments for clarity, but it's not strict requirement.

As aside, once we might want to put the module name also into context, because I imagine at some point, we will define rules, that work on pattern matches rather than exact module names, in which case, the condition could still be applied, so passing this is not only good for the error message and information keeping, but also for the actual conditions mid term. But now of course, doesn't make yet sense to add it.

kayhayen (approved):

kayhayen:
I merged this, but I will move it to factory branch and work a bit on it, I believe it breaks XML reporting, otherwise very good. But I want to do something nice for that reporting anyway.

PR #1657, opened 2022-07-05
fix ModuleNotFoundError for 'sklearn.utils.' and 'rapidfuzz.'

nicolaichuk:
fix ModuleNotFoundError for 'sklearn.utils.' and 'rapidfuzz.'

please review and merge to upstream


kayhayen (approved):
In the future, such changes ought to only happen in the yaml package configuration file, but I can see how you can existing stuff here.

PR #1691, opened 2022-07-22
Move the implicit imports from plugin to yaml

Fire-Cube:
- move the implicit imports from ImplicitImport.py to the yaml files
- fix schema
- add schema-check for stdlib2 and stdlib3

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -275,6 +275,6 @@
     "esbonio.sphinx.confDir": "",
     "cSpell.spellCheckOnlyWorkspaceFiles": true,
     "yaml.schemas": {
-        "misc/nuitka-package-config-schema.json": "standard.nuitka-package.config.yml"
+        "misc/nuitka-package-config-schema.json": ["standard.nuitka-package.config.yml", "stdlib2.standard.nuitka-package.config.yml", "stdlib3.standard.nuitka-package.config.yml"]
     }
 }
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -3,6 +3,26 @@
 # yamllint disable rule:comments-indentation
 # too many spelling things, spell-checker: disable
 ---
+- module-name: "_dbus_glib_bindings"
+  implicit-imports:
+    - depends:
+        - "_dbus_bindings"
+
+- module-name: "_mysql"
+  implicit-imports:
+    - depends:
+        - "_mysql_exceptions"
+
+- module-name: "_ruamel_yaml"
+  implicit-imports:
+    - depends:
+        - "ruamel.yaml.error"
+
+- module-name: "_yaml"
+  implicit-imports:
+    - depends:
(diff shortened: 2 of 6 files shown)


.vscode/settings.json
```diff
@@ -275,6 +275,6 @@
     "esbonio.sphinx.confDir": "",
     "cSpell.spellCheckOnlyWorkspaceFiles": true,
     "yaml.schemas": {
-        "misc/nuitka-package-config-schema.json": "standard.nuitka-package.config.yml"
+        "misc/nuitka-package-config-schema.json": ["standard.nuitka-package.config.yml", "stdlib2.standard.nuitka-package.config.yml", "stdlib3.standard.nuitka-package.config.yml"]
```
kayhayen, line 278: merge problem?

Fire-Cube: Why do you think?

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -110,6 +174,10 @@
     empty_dirs:
       - "."
 
+- module-name: "curses"
+  implicit-imports:
+    - depends:
+        - "_curses"
```
kayhayen, line 180: That's actually a stdlib module that we overlooked before.

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -130,6 +198,19 @@
   implicit-imports:
     - depends:
         - "cv2.cv2"
+        - "numpy"
+        - "numpy.core"
+
```
kayhayen, line 203: I will try and tweak auto-format for Yaml to enfore a separation of lines, probably on the ASCII level...

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -420,6 +867,32 @@
...
+        - "numpy.random.bit_generator"
+        - "numpy.random.bounded_integers"
+        - "numpy.random.common"
+        - "numpy.random.generator"
+        - "numpy.random.mt19937"
+        - "numpy.random.pcg64"
+        - "numpy.random.philox"
+        - "numpy.random.sfc64"
+
+      # TODO: Clarify if entropy is needed for 1.18 or at all.
```
kayhayen, line 893: Thanks for even maintaining the TODOs properly, looks very good to my eyes.

kayhayen (approved):
This is extremely well done. For my constant scare of breaking things, I will write a small tool, that will iterate over all available modules, and produce an output, that I can then diff. In this instance, it will have to be identical. There is already code to iterate over all modules, and we basically only need to output plugin decisions. I will merge, once I have the tool written, hopefully early this week.

nuitka/plugins/standard/standard.nuitka-package.config.yml
kayhayen: I think on Windows, curses doesn't work until something from pip is installed, that kind of messes things up.

PR #1722, opened 2022-08-07
Resolve symbolic links to the executable binary

chrisy:
# What does this PR do?

For all platforms (tested on Linux, Windows and macOS), adds a mechanism to resolve any symbolic links in the path of the executable when determining the directory of the binary. 

For all POSIX like systems, this makes use of `realpath()` to fully resolve the path. On Windows similar is achieved with `GetFinalPathNameByHandle()` (with a W suffix for `wchar` support.)

Without this, loading of bundled dynamic libraries and Python package resources will fail if a symbolic link that points to the executable file is used since Nuitka currently sets the Python directory to that of the link itself, not the file the link targets.

This change allows, for example, a dist directory built in `--standalone` mode to be deployed to a system in an arbitrary location and a symbolic link created to the executable from a directory already on the system PATH to that dist directory. Another use case might be to deploy a collection of utilities, each in their own directory, but with a common `bin` dir that contains symlinks to the individual tools, this requiring only one directory to be added to the `PATH`.

# PR Checklist

- [X] Correct base branch selected? Should be `develop` branch.
- [X] Enabled commit hook or executed `./bin/autoformat-nuitka-source`.
      cover enough.
  _Tests pass locally; won't know if the actions pass until they're enabled for me_
  _I did not find an existing place to effectively test this sort of change, other than assuming success since existing standalone and onefile tests still succeed._
  _Does this need any docs? TBH I had assumed this would just work and was surprised when it did not_

### DIFF
--- a/nuitka/build/static_src/CompiledCodeHelpers.c
+++ b/nuitka/build/static_src/CompiledCodeHelpers.c
@@ -1567,6 +1567,49 @@ NUITKA_MAY_BE_UNUSED static void stripFilenameW(wchar_t *path) {
 
 #if defined(_NUITKA_EXE)
 
+#ifdef _WIN32
+void resolveFileSymbolicLinksW(wchar_t *filename, wchar_t *resolved_filename, size_t resolved_filename_size) {
+    // Resolve any symbolic links in the filename.
+    // Copies the resolved path over the top of the parameter.
+    HANDLE file_handle;
+    DWORD len;
+
+    // Open the file in the most non-exclusive way possible
+    file_handle = CreateFileW(filename, 0, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, NULL, OPEN_EXISTING,
+                              FILE_ATTRIBUTE_NORMAL, NULL);
+
+    if (unlikely(file_handle == INVALID_HANDLE_VALUE)) {
+        abort();
+    }
+
+    // Resolve the path, get the result with a drive letter
+    len = GetFinalPathNameByHandleW(file_handle, resolved_filename, resolved_filename_size,
+                                    FILE_NAME_NORMALIZED | VOLUME_NAME_DOS);
+
+    CloseHandle(file_handle);
+
+    if (unlikely(len >= resolved_filename_size)) {
+        abort();
+    }
+}
+
+#else  // ifndef _WIN32
+
+void resolveFileSymbolicLinks(char *filename, char *resolved_filename, size_t resolved_filename_size) {
+    char *result;
+
+    if (resolved_filename_size < PATH_MAX) {
+        abort();
+    }
+
+    result = realpath(filename, resolved_filename);
+
+    if (result == NULL) {
+        abort();
+    }
+}
+#endif // _WIN32
+
 #ifndef _WIN32
 char const *getBinaryDirectoryHostEncoded(void) {
     static char binary_directory[MAXPATHLEN + 1];
@@ -1576,13 +1619,18 @@ char const *getBinaryDirectoryHostEncoded(void) {
         return binary_directory;
     }
 
+    char binary_filename[MAXPATHLEN + 1];
+
 #if defined(__APPLE__)
-    uint32_t bufsize = sizeof(binary_directory);
-    int res = _NSGetExecutablePath(binary_directory, &bufsize);
+    uint32_t bufsize = sizeof(binary_filename);
+    int res = _NSGetExecutablePath(binary_filename, &bufsize);
 
     if (unlikely(res != 0)) {
         abort();
     }
+
+    // Resolve any symlinks we were invoked via
+    resolveFileSymbolicLinks(binary_filename, binary_directory, sizeof(binary_directory));
 
     // On macOS, the "dirname" call creates a separate internal string, we can
     // safely copy back.
@@ -1597,12 +1645,15 @@ char const *getBinaryDirectoryHostEncoded(void) {
     mib[1] = KERN_PROC;
     mib[2] = KERN_PROC_PATHNAME;
     mib[3] = -1;
-    size_t cb = sizeof(binary_directory);
-    int res = sysctl(mib, 4, binary_directory, &cb, NULL, 0);
+    size_t cb = sizeof(binary_filename);
+    int res = sysctl(mib, 4, binary_filename, &cb, NULL, 0);
 
     if (unlikely(res != 0)) {
         abort();
     }
+
+    // Resolve any symlinks we were invoked via
+    resolveFileSymbolicLinks(binary_filename, binary_directory, sizeof(binary_directory));
 
     /* We want the directory name, the above gives the full executable name. */
     copyStringSafe(binary_directory, dirname(binary_directory), sizeof(binary_directory));
@@ -1611,12 +1662,15 @@ char const *getBinaryDirectoryHostEncoded(void) {
 
     /* The "readlink" call does not terminate result, so fill zeros there, then
      * it is a proper C string right away. */
-    memset(binary_directory, 0, sizeof(binary_directory));
-    ssize_t res = readlink("/proc/self/exe", binary_directory, sizeof(binary_directory) - 1);
+    memset(binary_filename, 0, sizeof(binary_filename));
+    ssize_t res = readlink("/proc/self/exe", binary_filename, sizeof(binary_filename) - 1);
 
     if (unlikely(res == -1)) {
         abort();
     }
+
+    // Resolve any symlinks we were invoked via
+    resolveFileSymbolicLinks(binary_filename, binary_directory, sizeof(binary_directory));
 
     copyStringSafe(binary_directory, dirname(binary_directory), sizeof(binary_directory));
 #endif
@@ -1633,8 +1687,11 @@ wchar_t const *getBinaryDirectoryWideChars(void) {
         binary_directory[0] = 0;
 
 #if defined(_WIN32)
-        DWORD res = GetModuleFileNameW(NULL, binary_directory, sizeof(binary_directory));
+        wchar_t binary_filename[MAXPATHLEN + 1];
+        DWORD res = GetModuleFileNameW(NULL, binary_filename, sizeof(binary_filename));
         assert(res != 0);
+
+        resolveFileSymbolicLinksW(binary_filename, binary_directory, sizeof(binary_directory));
 
         stripFilenameW(binary_directory);
 


kayhayen:
Sorry, had to make a rebase of the stable release even, due to a severe regression, but I can work off your commit.

kayhayen:
So, what this does, is to change the launched filename to the realpath.

But standard Python doesn't do this, or does it?

```
 A  hayen@anna: ~/repos/Py2C$ ln -s m.py l.py
 A  hayen@anna: ~/repos/Py2C$ python m.py 
m.py
/home/hayen/repos/Py2C/m.py
 A  hayen@anna: ~/repos/Py2C$ python l.py 
l.py
/home/hayen/repos/Py2C/l.py
```

where `m.py` is this:
```
import sys
print(sys.argv[0])
print(__file__)
```

Now, why should Nuitka then do a `os.path.realpath` rather than the original code doing it? Also hiding the original launching directory might have other adverse effects. For finding files near you, in standalone and onefile, this code is recommened:

```
   # This will find a file near your onefile.exe
   open(os.path.join(os.path.dirname(sys.argv[0]), "user-provided-file.txt"))
   # This will find a file inside your onefile.exe
   open(os.path.join(os.path.dirname(__file__), "user-provided-file.txt"))
```

This might be a missing an `os.path.realpath` if you want to automatically go to near the binary, for standalone the two ought to be equivalent, but as you suggest, being called through a symlink. I am adding a section for standalone, and amending the onefile for symlinks:

```
Standalone: Finding files
-------------------------

The standard code that normally works, also works, you should refer to
``os.path.dirname(__file__)`` or use all the packages like ``pkgutil``,
``pkg_resources``, ``importlib.resources`` to locate data files near the
standalone binary.

For bonus points, use ``os.path.dirname(os.path.realpath(__file__))`` which will
also remove symlinks.

.. important::

   What you should **not** do, is use the current directory ``os.getcwd``,
   assuming that this is the script directory, that is not generally true, and
   was never good code. Links, to a program, etc. will all fail in bad ways.

```

kayhayen:
If you agree, I would close this issue. Basically you need to implement "executing through symlink" support in your Python program. I am not sure about 3rd-party code, that will still be fooled by the symlink, so this may still have merrit. But it's also loosing information unconditionally, never a good thing, namely from where a program was called.

Maybe we would have to patch 3rd party libraries with anti-bloat to do the realpath stuff if we cared about them. I will try and checkout with Jinja2 PackageLoader, which I think will fall into that category. Maybe we can change one (`__file__`), but not the other (`sys.argv[0]`). It also would be mostly only `__main__` that really has that other location, other modules will load where they live, so that should be the more compatible thing.

I will need to check if your code change allows for keeping both things.

chrisy:
Unfortunately, from inside Python I do not seem able to coerce Nuitka+Python to load its own shared libraries.

```python
#!/usr/bin/env python

import os
import sys

print(f"sys.path:{sys.path!r}")

import subprocess
```

The binary is in `.../nuitka-test/test.dist/test`. The symlink is in a different directory, called `.../nuitka-test/mytest`.

Output:

```console
$ ./mytest
sys.path:['.../nuitka-test/test.dist', '.../nuitka-test']
  File ".../nuitka-test/test.py", line 8, in <module>
    import subprocess
  File ".../nuitka-test/subprocess.py", line 78, in <module>
ImportError: dlopen(.../nuitka-test/_posixsubprocess.so, 0x0002): tried: '.../nuitka-test/_posixsubprocess.so' (no such file)
```

This is what I tried to solve for. The correct directory is on `sys.path` already, but it's not what Nuitka tells Python to use for its `_*.so` files. That is set by `Py_SetPythonHome`, and for standalone mode Nuitka sets that to the result of `getBinaryDirectory*()`, which is in turn derived from the path to the executable.

kayhayen:
I see, that is another case, where `realpath` must be used. Makes me wonder, if it should be used outside of `sys.argv[0]` and `__main__` module `__file__` at all. Ok, this looks good then. Will you be able to rebase the PR, there is a section in the developer manual on how to do it. Basically `git rebase upstream/develop` and skipping all conflicting commits with `git rebase --skip` will usually work perfectly.

chrisy:
Rebased. It occurs to me, if you're worried about regressions, we could move this symlink resolution to _just_ the call to `Pt_SetPythonHome` in  `prepareStandaloneEnvironment`. I _think_ that would be sufficient. I placed it inside `getBinaryDirectory*` because the name suggested it should return the directory the binary is in. Could also make it optional or disableable.

kayhayen:
Making things optional is typically a cop out. The default is always wrong, and people who need the non-default are never going to understand what it is about in the first place. Yes, we can move the code, but I am getting the sense that `MAKE_RELATIVE_PATH` needs to use that, and it is using `getBinaryDirectoryObject` which is just a wrapper around that.

So I think it should be used most often, and like I said, only `sys.argv[0]` and the main module `__file__` value, must point to where the symlink is not resolved.

kayhayen:
I am going to merge this now, but don't be surprised, if it goes away soon after, I will move it to my factory branch and apply a few changes and tests on my side first.

kayhayen:
@chrisy I am deeply confused, I cannot see the problem. I think we are talking about lauching a standalone distribution via a symlink, right?

I created a standalone distribution that imports a module, and outputs the `file` stuff on Linux, and it's already correct:

```
sys.argv[0] m.dist/m
__main__ module __file__ /home/hayen/repos/Py2C/m.dist/m.py
import path ['/home/hayen/repos/Py2C/m.dist', '/home/hayen/repos/Py2C/m.dist']
/home/hayen/repos/Py2C/m.dist/x.py
```

or through symlink `y`from elsewhere:

```
sys.argv[0] ./y
__main__ module __file__ /home/hayen/repos/Py2C/m.dist/m.py
import path ['/home/hayen/repos/Py2C/m.dist', '/home/hayen/repos/Py2C/m.dist']
/home/hayen/repos/Py2C/m.dist/x.py
```

So I am now confused, if the change is even necessary, but I found out why:

```
    /* The "readlink" call does not terminate result, so fill zeros there, then
     * it is a proper C string right away. */
    memset(binary_filename, 0, sizeof(binary_filename));
    ssize_t res = readlink("/proc/self/exe", binary_filename, sizeof(binary_filename) - 1);
```

So Linux doesn't need this at all, or better yet, always did it like this. I actually wouldn't know how to get the unresolved path, it is only in `sys.argv[0]` and makes the main `__file__` incompatible, but that would be close to impossible to fix, and of dubious value, since I doubt many people need that to work, whereas creating a symlink to a binary after it has been compiled, will be much more common. Not changing behavior is more important there.

kayhayen:
This also means, my concerns are probably all addressed, checking on Windows now.

kayhayen:
On Windows, the CPython DLL fails to load when executing a standalone distribution through a symlink, the DLL loading, will not add the symlink target to the path, and therefore it cannot work.  It being a hard link doesn't improve things naturally, since then there is nothing to resolve. So we don't even start to execute any of our own code in this case. I really wish I had a solution for that Python DLL being allowed to load from elsewhere.

kayhayen:
I see, you put the link target to `PATH` environment, that would help yes.

kayhayen:
Ok, I can confirm, on Windows, with the CPython DLL being made loadable, it actually makes a difference indeed. I changed the code to be only used when `--experimental=symlinks` is given, just because I am trying to get this onto develop without disrupting other things.

chrisy:
Heh, I see you've been busy as I was driving to work. Let me know if there's anything I can do to assist, but it sounds like you've ended up with similar conclusions to mine.

andyfase:
Just figured I would add that this experimental feature proved a life saver for me. I'm attempting to distribute a Python based CLI application through homebrew on OSX. 

Homebrew relies on symlink'ing to installed application. On OSX @executable_path only resolves to the location of the symlink not to the executable itself. Hence I would end up with errors such as 

```
~> path/to/symlink/cli_tool
  File "/path/to/symlink/cli_tool.py", line 5, in <module>
  File "/path/to/symlink/shelve.py", line 59, in <module>
  File "/path/to/symlink/pickle.py", line 33, in <module>
  File "/path/to/symlink/struct.py", line 13, in <module>
ImportError: dlopen(/path/to/symlink/_struct.so, 0x0002): tried: '/path/to/symlink/_struct.so' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/path/to/symlink/_struct.so' (no such file), '/path/to/symlink/_struct.so' (no such file)
```
When running directly to the fullpath of the actual CLI executable - it runs without issue.

With `--experimental=symlinks` this issue goes away and the symlinked application can be run without issue. Recommend reconsidering taking this out of "experimental" status

kayhayen:
You are kind of right, for the paths, I guess it's essential for loading. Looking at that now. I feel there was a drawback.

kayhayen:
So, that `dlopen` should use a different path, I am not convinced that the symlinks do not work with `@executable_path`. We probably just need both, the original path without symlink resolved for some purposes, like `__file__` and the one with, for other purposes, namely finding the extension modules. I wonder if the same wouldn't have to happen to Linux too, so I will try it now.

kayhayen:
For whatever reason, Linux is not affected. I have made both branches used, and selective where. This will also make some things more robust, e.g. `PYTHONHOME`, DLL directory of Windows, etc. will now be resolved paths, and others are not. This might go wrong.

I have a feeling that Linux resolves `/proc/self/exe` whereas on macOS we do not have that, so there might be issues caused by it, but lets face it, it cannot be worse. Right now no extension module can be loaded in the situation. Therefore Linux maybe was almost correct already, but on Windows I see that it should improve things and work out the box now.

kayhayen:
Lets trace this here, will be in 1.6, and now on factory.

Thanks for your report, this is fixed on the factory branch, which is a development version under rapid development. You can try it out by going here: https://nuitka.net/doc/factory.html 

Feedback if this is working is very welcome, just please do not share plans of doing it, but rather confirmations or denials of it working.

andyfase:
Hi thanks for looking at this. I checked out `1.6rc8` rebuilt without the experimental flag and can confirm that the resulting binary works through symlinks now 

Thanks again!

kayhayen:
This is part of the stable release 1.6 that I just made. 

It however most likely breaks Windows Vista and Windows XP, which don't have this. I made it dependent on the target API there, but MinGW64 targets XP, while apparently MSVC does not. We need some sort of target platform in the future, or make this code dynamically use that function accepting failure to load. I am therefore reopening this one to clean it up for 1.6 early on.

kayhayen:
I wonder where that milestone came from, did I really slack that long on this, tsk.

PR #1751, opened 2022-08-23
do not check known kernel shared libraries for relative paths

gvegidy:
Nuitka contains a list of known "kernel" shared libraries that are not included.
Since these are not included, we do not need to ensure that they don't contain
relative paths.

Without this patch I got this error (`nuitka3 --standalone diag.py`):
```
FATAL: Error: Found a dependency with a relative path. Was a dependency copied to dist early? linux-vdso.so.1
```

This occured under these conditions:

    python -m nuitka --version

1.0.5
Commercial: None
Python: 3.10.6 (main, Aug  3 2022, 17:39:45) [GCC 12.1.1 20220730]
Flavor: Unknown
Executable: /usr/bin/python
OS: Linux
Arch: x86_64
Distribution: Sysrescue (based on arch) 9.04

    How did you install Nuitka and Python
    Arch Linux packages python-3.10.6-1 and nuitka-1.0.5-1 installed through pacman.
    No virtualenv used.

### DIFF
--- a/nuitka/freezer/DllDependenciesPosix.py
+++ b/nuitka/freezer/DllDependenciesPosix.py
@@ -114,13 +114,6 @@ def detectBinaryPathDLLsPosix(dll_filename, package_name, original_dir):
         # been seen with Qt at least.
         filename = os.path.normpath(filename)
 
-        # If we encounter a valid relative path, resolve it to an absolute one.
-        if not os.path.isabs(filename):
-            inclusion_logger.sysexit(
-                "Error: Found a dependency with a relative path. Was a dependency copied to dist early? "
-                + filename
-            )
-
         # Do not include kernel DLLs on the ignore list.
         filename_base = os.path.basename(filename)
         if any(
@@ -128,6 +121,13 @@ def detectBinaryPathDLLsPosix(dll_filename, package_name, original_dir):
             for entry in _linux_dll_ignore_list
         ):
             continue
+
+        # Do not allow relative paths for shared libraries
+        if not os.path.isabs(filename):
+            inclusion_logger.sysexit(
+                "Error: Found a dependency with a relative path. Was a dependency copied to dist early? "
+                + filename
+            )
 
         result.add(filename)
 


kayhayen:
Thanks, I had to implement this blindly, i.e. without testing, this is a very good solution for that too.

kayhayen:
I am not sure when 1.1 will be released, but develop is usually very much stable too.

gvegidy:
Wouldn't it make sense to cherry-pick this change into the next stable release (1.0.7) because it is a fix for a regression?

gvegidy:
When I was using nuitka 0.8.4, using the `--standalone` option worked without errors. I kept all other packages from Arch Linux at exactly the same version (date-snapshot from Arch Linux Archive) and just updated nuitka to 1.0.5 to test the new --onefile function without AppImage. Then the problem with the relative path error message came up. So in my eyes this makes it a regression in nuitka.

kayhayen:
Ok, I see, too bad. I will consider it then, but I believe there may not be another hotfix before 1.1 at all.

kayhayen:
Ok, I see, too bad. I will consider it then, but I believe there may not be another hotfix before 1.1 at all. 

When the issue came up in Discord, I was pretty sure, it affects only newer glibc, which due to a change is giving a filename for that kernel library. I think, what happened is that I hardened the check at some point, to make it complain about non-found DLLs, rather than ignoring those. So, that might have been 0.9 material, so in a sense it's pretty old. We could look it up, but it doesn't matter. Also what date are you using, a fairly recent one, or older one?

gvegidy:
The Arch Linux snapshot I've tested this with is from `2022-08-06`, so yes, quite recent.

kayhayen:
In my mind, that is when the regression happened. Updating to newest Linux and using stable Nuitka is not a core use case. Actually for standalone builds, I do recommend to not use OS packages, which probably require to new glibc, so your binaries built on latest greatest glibc will not be all that portable. I recommend using CentOS7 and old Debian/Ubuntu for building. In Nuitka commercial (https://nuitka.net/doc/commercial.html), there is a podman based build, which does it transparently on latest OS in a container. Yet another of its undescribed features, taking note to add a description of that there too.

PR #1790, opened 2022-09-14
Add support for Nuitka-Python to set link flags.

Maxwell175:

# What does this PR do?

Add support for Nuitka-Python to set link (and compile) flags.
Also fix an issue that caused Nuitka to not find dlls that python is linked to.

# PR Checklist

      cover enough.

### DIFF
--- a/nuitka/freezer/DllDependenciesMacOS.py
+++ b/nuitka/freezer/DllDependenciesMacOS.py
@@ -17,13 +17,14 @@
 #
 """DLL dependency scan methods for macOS. """
 
+import json
 import os
 import sys
 
 from nuitka.containers.OrderedDicts import OrderedDict
 from nuitka.containers.OrderedSets import OrderedSet
 from nuitka.Errors import NuitkaForbiddenDLLEncounter
-from nuitka.PythonFlavors import isAnacondaPython
+from nuitka.PythonFlavors import isAnacondaPython, isNuitkaPython
 from nuitka.Tracing import inclusion_logger
 from nuitka.utils.FileOperations import areSamePaths, isPathBelow
 from nuitka.utils.SharedLibraries import (
@@ -146,7 +147,6 @@ def _resolveBinaryPathDLLsMacOS(
 
     rpaths = _detectBinaryRPathsMacOS(original_dir, binary_filename)
     rpaths.update(package_specific_dirs)
-    print(rpaths)
 
     for path in paths:
         if path.startswith("@rpath/"):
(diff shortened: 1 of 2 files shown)


nuitka/freezer/DllDependenciesMacOS.py
```diff
@@ -145,6 +146,7 @@ def _resolveBinaryPathDLLsMacOS(
 
     rpaths = _detectBinaryRPathsMacOS(original_dir, binary_filename)
     rpaths.update(package_specific_dirs)
+    print(rpaths)
```
kayhayen, line 149: That stray print is going to kill tests

nuitka/freezer/DllDependenciesMacOS.py
```diff
@@ -161,6 +163,14 @@ def _resolveBinaryPathDLLsMacOS(
         elif os.path.basename(path) == os.path.basename(binary_filename):
             # We ignore the references to itself coming from the library id.
             continue
+        elif not os.path.isabs(path) and not os.path.exists(path):
```
kayhayen, line 166: So, relative paths with any '@' are a thing, where did you see these, how does macOS handle these?!

Maxwell175: When setting up the linking of one of the necessary dll dependencies for one of the packages that i need for Nuitka-Python, that is the default way that it links it.

Here is the otool Load command:
```
Load command 27
          cmd LC_LOAD_DYLIB
      cmdsize 40
         name libCg.dylib (offset 24)
   time stamp 2 Wed Dec 31 19:00:02 1969
      current version 0.0.0
compatibility version 0.0.0
```

Here is where it searches:
```
dyld[39260]: Library not loaded: 'libCg.dylib'
  Referenced from: '/Users/maxwell/test/python3'
  Reason: tried: 'libCg.dylib' (no such file), '/usr/local/lib/libCg.dylib' (no such file), '/usr/lib/libCg.dylib' (no such file), '/Users/maxwell/test/libCg.dylib' (no such file), '/usr/local/lib/libCg.dylib' (no such file), '/usr/lib/libCg.dylib' (no such file)
```

nuitka/freezer/DllDependenciesMacOS.py
Maxwell175: Fixed.

nuitka/freezer/DllDependenciesMacOS.py
Maxwell175: It seems that this happens when a dll is linked in using the `-l` option and the library directory is specified with `-L`. This is really only an issue with Nuitka Python so I updated the code to be in a `isNuitkaPython` check and also to be more precise and robust.

PR #1797, opened 2022-09-18
YAMLFormatter: change standard quote to single

Fire-Cube:
What the title says ;)


kayhayen:
This doesn't change the Yaml at all, but it looks like a nice cleanup still, but you might have to find a bug here.

PR #1798, opened 2022-09-18
add tool to only format the package-config YAML files

Fire-Cube:
what the title says ;)


kayhayen:
I would much rather than more binaries add an option to the existing auto-format such that it limits the file scan to a subset, say `--yaml`, `-c`, and `--python` which make everything else excluded.

Therefore I reject this approach, sorry.

PR #1827, opened 2022-10-04
wip: generate StrNodes from jinja2 template

Fire-Cube:
# What does this PR do?

- add template to generate StrNodes
- add GenerateNodes.py to manage the generation
It is designed so that you can easily extend it.


nuitka/tools/specialize/GenerateNodes.py
```diff
@@ -0,0 +1,62 @@
...
+            ["Lstrip2", "chars", "Str", False, True, "Strip2", True, "str.lstrip"],
+            ["Rstrip2", "chars", "Str", False, True, "Strip2", True, "str.rstrip"],
+        ],
+    }
+]
+
+
+def main():
+    for spec in specifications:
+        with open(
```
kayhayen, line 45: For reading templates there are helpers in Nuitka that are easy to use, or so I am thinking, check `nuitka.utils.Jinja2`

nuitka/tools/specialize/templates_python/StrNodes.py.j2
```diff
@@ -0,0 +1,137 @@
...
+                description="Built-in 'str.{{ operation_name.lower() }}' with constant values.",
+                user_provided=str_arg.user_provided,
+            )
+
+        trace_collection.onExceptionRaiseExit(BaseException)
+
+        return self, None, None
+        {% endif %}
+    {% endif %}
+    {% if operation_name in ["Partition", "Rpartition"]%}
```
kayhayen, line 132: Using real Python level operation name rather than class name would be better, there is probably a rule, using `operation_name.title()` to get the class name from it, that templates should use, not the other way around, i.e. ` operation_name.lower()` is kind of wrong way around.

nuitka/tools/specialize/templates_python/StrNodes.py.j2
```diff
@@ -0,0 +1,137 @@
...
+    return (
+        cls
+        for kind, cls in NodeCheckMetaClass.kinds.items()
+        if kind.startswith("EXPRESSION_STR_OPERATION_")
+    )
+
+{% for operation_name, arg_name, shape, is_base, inheritance_base, base_name, simulator, simulator_return_value in specifications %}
+class ExpressionStrOperation{{ operation_name }}(
+    {% if inheritance_base %}
+    ExpressionStrOperation{{ base_name }}Base
```
kayhayen, line 54: Actually I would like to get rid of the base classes when we do the generation, i.e. their content is generated into the class.

Fire-Cube: Cool, that makes a lot of things easier.

nuitka/tools/specialize/templates_python/StrNodes.py.j2
```diff
@@ -0,0 +1,137 @@
...
+                user_provided=str_arg.user_provided,
+            )
+
+        trace_collection.onExceptionRaiseExit(BaseException)
+
+        return self, None, None
+        {% endif %}
+    {% endif %}
+    {% if operation_name in ["Partition", "Rpartition"]%}
+    @staticmethod
```
kayhayen, line 133: Some sort of description of the operation should exist somewhere and be available in the template and say e.g. iterable result or not, and if static length, what it is, and maybe even the type shape. I saw the `shape` template arg, that's fine, but this can get quickly too many things, puting it is a dict by the operation name makes it easy for dumping related information.

kayhayen:
This has been implemented based on these ideas.

PR #1894, opened 2022-11-18
Removing from exlude list `libstdc++` on Alpine. Fixes #1893

### LINKED ISSUE #1893: Remove from `_linux_dll_ignore_list` `libstdc++` for `musl` Linux(Alpine).
[    # The C++ standard library can also be ABI specific, and can cause system](https://github.com/Nuitka/Nuitka/blob/bc8ede75c6bde306546177a15f186e8f1a8e535a/nuitka/freezer/DllDependenciesPosix.py#L199)

For linuxes without `libstdc++` like Alpine it would be nice to include `libstdc++` in build.
Example:
1. Install `scipy` in Alpine - it requires `libstdc`
2. Perform standalone build and move binary to new Alpine
3. `import scipy.fftpack`
4. `ImportError: Error loading shared library libstdc++.so.6: No such file or directory`

Tested the same with removed `libstdc` from ignored list and it is ok.

bigcat88:
# What does this PR do?

- Fixes `standalone` build for Alpine when `libstdc++` is required.

https://github.com/Nuitka/Nuitka/issues/1893

### DIFF
--- a/nuitka/freezer/DllDependenciesPosix.py
+++ b/nuitka/freezer/DllDependenciesPosix.py
@@ -27,7 +27,7 @@ from nuitka.PythonFlavors import isAnacondaPython
 from nuitka.Tracing import inclusion_logger
 from nuitka.utils.Execution import executeProcess, withEnvironmentPathAdded
 from nuitka.utils.SharedLibraries import getSharedLibraryRPATH
-from nuitka.utils.Utils import isPosixWindows, isAlpineLinux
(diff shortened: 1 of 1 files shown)


kayhayen:
Thanks, that seems very reasonable, esp. as I hard to find out the hard way that Anaconda has the very same issue.

kayhayen:
btw, you should apply the commit hook, not do "isort", as Nuitka has its own ways, but I will survive if that causes a build failure

Issue #1591, opened 2022-06-01
Windows: Random "access denied" on clcache objects in Scons


deajan:
Hello,

Been a long time. I begun using Nuitka again recently for a project I am working on.
I've built a small API with `fastapi` and `uvicorn`, but I hit some strange Nuitka problems as my code is growing in size.

I compile my project with `c:\Python39-64\python.exe -m nuitka --include-module=ztl --include-plugin-directory=www --standalone web_server.py`
Randomly, I get cons access denied errors. The more dependencies my project has, the more it's likely I get those errors:

```
Nuitka-Options:INFO: Used command line options: --include-module=ztl --standalone web_server.py
Nuitka:INFO: Starting Python compilation with Nuitka '0.8.3' on Python '3.9' commercial None.
Nuitka-Plugins:INFO: implicit-imports: Implicit dependencies of module 'fastapi' added 'fastapi.routing'.
Nuitka-Plugins:INFO: implicit-imports: Implicit dependencies of module 'anyio._core._eventloop' added 'anyio._backends._asyncio'.
Nuitka-Plugins:INFO: implicit-imports: Implicit dependencies of module 'pydantic' added 'pydantic.typing,pydantic.fields,pydantic.utils,pydantic.schema,pydantic.env_settings,pydantic.main,pydantic.error_wrappers,pydantic.validators,pydantic.mypy,pydantic.version,pydantic.types,pydantic.color,pydantic.parse,pydantic.json,pydantic.datetime_parse,pydantic.dataclasses,pydantic.class_validators,pydantic.networks,pydantic.errors'.
Nuitka-Plugins:INFO: implicit-imports: Implicit dependencies of module 'pydantic.typing' added 'typing_extensions'.
Nuitka-Plugins:INFO: anti-bloat: Handling module 'psutil' with 11 change(s) for: resolve platform specific imports at compile time.
Nuitka-Plugins:INFO: implicit-imports: Implicit dependencies of module 'uvicorn' added 'uvicorn.loops,uvicorn.lifespan,uvicorn.protocols'.
Nuitka-Plugins:INFO: implicit-imports: Implicit dependencies of module 'uvicorn.protocols' added 'uvicorn.protocols.http,uvicorn.protocols.websockets'.
[...]

PR #1914, opened 2022-12-02
Windows: Retry replace clcache entry after a delay

ittegrat:
# What does this PR do?
Apply `decoratorRetries` in clcache to overcome random "access denied" errors tied to anti-virus.

See #1591


kayhayen:
Pitch perfect, the only thing that we could do different is to use scons logger, but it wouldn't really be much related to scons anyway. Thanks for your contribution!

kayhayen:
I will put this one into a hotfix later this weekend btw.

PR #1922, opened 2022-12-07
Don't use PyFloat_AS_DOUBLE() as an l-value

vstinner:
* Use copysign() with PyFloat_FromDouble() in HelpersConstantsBlob.c.
* Add PyFloat_SET_DOUBLE() function and use it in in-place operations.

Python 3.12 PyFloat_AS_DOUBLE() can no longer be used as an l-value: https://github.com/python/cpython/commit/02f72b8b938e301bbaaf0142547014e074bd564c

# What does this PR do?

# PR Checklist

      cover enough.

### DIFF
--- a/nuitka/build/static_src/HelpersConstantsBlob.c
+++ b/nuitka/build/static_src/HelpersConstantsBlob.c
@@ -976,11 +976,7 @@ static unsigned char const *_unpackBlobConstants(PyObject **output, unsigned cha
                 static PyObject *_const_float_minus_0_0 = NULL;
 
                 if (_const_float_minus_0_0 == NULL) {
-                    _const_float_minus_0_0 = PyFloat_FromDouble(0.0);
-
-                    // Older Python3 has variable signs from C, so be explicit about it.
-                    PyFloat_AS_DOUBLE(_const_float_minus_0_0) =
-                        copysign(PyFloat_AS_DOUBLE(_const_float_minus_0_0), -1.0);
+                    _const_float_minus_0_0 = PyFloat_FromDouble(copysign(0.0, -1.0));
                 }
                 z = _const_float_minus_0_0;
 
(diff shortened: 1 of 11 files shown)


kayhayen:
Thank you so much for pointing this out and sharing the information.

I am having 3.11 compile some things at this time, with --experimental=python3.12 you could probably compile it using current develop branch. As Nuitka is crazy enough to support all kinds of older Python versions, it will probably require some sort of abstraction to work with older ones and backport of the macro in case it is not defined. I didn't find a 3.12 branch yet. Can you point me to the macro definition on your branch?

kayhayen:
Sorry for my bad reading skills, you of course linked the commit:

```
static inline double PyFloat_AS_DOUBLE(PyObject *op) {
    return _PyFloat_CAST(op)->ob_fval;
}
#define PyFloat_AS_DOUBLE(op) PyFloat_AS_DOUBLE(_PyObject_CAST(op))
```

I didn't find `PyFloat_SET_DOUBLE` but I imagine it follows what happened to `Py_SIZE` and `Py_SET_SIZE` in 3.11 before, and I agree, it makes sense. I will change it as suggested. Your Jinja2 template change removed a branch though, that I would like to keep, so I am going to make the change now in a similar fashion as you suggested.

vstinner:
> I didn't find PyFloat_SET_DOUBLE

I added it to Nuitka. It doesn't exist in Python 3.12. Maybe it's a bad name. My implementation modifies a reference to a pointer: ``PyObject**``. Maybe it's a bad name. I don't know Nuitka naming convention.

>  Your Jinja2 template change removed a branch though

Yeah, I moved the code into a function, to avoid code duplication. The function can be declared as a static inline function, if you prefer.

> I will change it as suggested

Great! Thanks!

PR #1930, opened 2022-12-12
User Manual: Escape backslashes in Windows paths

joouha:
# What does this PR do?

This PR makes a minor correction to the User Manual - the back-slashes in the Windows file paths in the table on use-case 4 were not escaped, so were not visible when the page was rendered.

It was initiated because I was reading the User Manual and noticed missing back-slashes in the Windows file paths in the table on use-case 4.

# PR Checklist

      cover enough.


kayhayen:
Thanks for fixing this, very good.

PR #1945, opened 2022-12-19
report missing modules

Fire-Cube:
# What does this PR do?
reports the missing modules

### DIFF
--- a/nuitka/BytecodeCaching.py
+++ b/nuitka/BytecodeCaching.py
@@ -112,7 +112,7 @@ def writeImportedModulesNamesToCache(module_name, source_code, used_modules):
     cache_name = makeCacheName(module_name, source_code)
     cache_filename = _getCacheFilename(cache_name, "json")
 
-    used_modules = [module._asdict() for module in used_modules]
+    used_modules = [module.asDict() for module in used_modules]
     for module in used_modules:
         module["source_ref_line"] = module["source_ref"].getLineNumber()
         del module["source_ref"]
--- a/nuitka/BytecodeCaching.py
+++ b/nuitka/BytecodeCaching.py
@@ -23,8 +23,7 @@ such that it allows to restore it directly.
 
 import os
 
-from nuitka.importing.Importing import getPackageSearchPath, isPackageDir
-from nuitka.nodes.ImportNodes import ModuleUsageAttempt
+from nuitka.importing.Importing import getPackageSearchPath, isPackageDir, ModuleUsageAttempt
 from nuitka.plugins.Plugins import Plugins
 from nuitka.utils.AppDirs import getCacheDir
 from nuitka.utils.FileOperations import listDir, makePath
--- a/nuitka/BytecodeCaching.py
+++ b/nuitka/BytecodeCaching.py
@@ -84,7 +84,7 @@ def getCachedImportedModulesNames(module_name, source_code):
     if data["module_name"] != module_name:
         return None
 
-    used_modules = list(data['modules_used'])
+    used_modules = list(data["modules_used"])
     for module in used_modules:
         module["module_name"] = ModuleName(module["module_name"])
 
--- a/nuitka/BytecodeCaching.py
+++ b/nuitka/BytecodeCaching.py
@@ -24,6 +24,7 @@ such that it allows to restore it directly.
 import os
 
 from nuitka.importing.Importing import getPackageSearchPath, isPackageDir
+from nuitka.nodes.ImportNodes import ModuleUsageAttempt
 from nuitka.plugins.Plugins import Plugins
 from nuitka.utils.AppDirs import getCacheDir
 from nuitka.utils.FileOperations import listDir, makePath
@@ -62,7 +63,7 @@ def hasCachedImportedModulesNames(module_name, source_code):
 
 
 # Bump this is format is changed or enhanced implementation might different ones.
-_cache_format_version = 2
+_cache_format_version = 3
 
 
 def getCachedImportedModulesNames(module_name, source_code):
@@ -83,24 +84,26 @@ def getCachedImportedModulesNames(module_name, source_code):
     if data["module_name"] != module_name:
         return None
 
-    return [
-        (ModuleName(used_module_name), line_number)
-        for (used_module_name, line_number) in data["modules_used"]
-    ]
+    used_modules = [module for module in data["modules_used"]]
+    for module in used_modules:
+        module["module_name"] = ModuleName(module["module_name"])
+
+    return [ModuleUsageAttempt(**module) for module in used_modules]
 
 
 def writeImportedModulesNamesToCache(module_name, source_code, used_modules):
     cache_name = makeCacheName(module_name, source_code)
     cache_filename = _getCacheFilename(cache_name, "json")
 
+    used_modules = [module._asdict() for module in used_modules]
+    for module in used_modules:
+        module["source_ref"] = module["source_ref"].getLineNumber()
+
     data = {
         "file_format_version": _cache_format_version,
         "module_name": module_name.asString(),
         # We use a tuple, so preserve the order.
-        "modules_used": tuple(
-            (used_module_name.asString(), source_ref.getLineNumber())
-            for used_module_name, _filename, _finding, _level, source_ref in used_modules
-        ),
+        "modules_used": used_modules,
     }
 
     makePath(os.path.dirname(cache_filename))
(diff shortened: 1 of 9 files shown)


nuitka/nodes/ImportNodes.py
```diff
@@ -148,6 +148,9 @@
 
 hard_modules_limited = ("importlib.metadata", "ctypes.wintypes", "importlib_metadata")
 
+global missing_modules
```
kayhayen, line 151: Global has no effect here

nuitka/nodes/ImportNodes.py
```diff
@@ -330,6 +333,10 @@ def _checkHardModules():
 _checkHardModules()
 
 
+def getMissingModules():
```
kayhayen, line 336: This must be attached to a module object.

nuitka/nodes/ImportNodes.py
```diff
@@ -1015,9 +1022,11 @@ def _attemptFollow(self, module_name):
             module_name = resolveModuleName(module_name)
 
             while True:
+                backup_module_name = str(module_name)
```
kayhayen, line 1025: Why would you remember strings ever and not the module name objects. Convert them to strings when time comes.

kayhayen:
I was expecting this to pick up failure to locate from multiple node types. I am not so sure the factory function is the only or the real place to check things.

kayhayen (changes requested):

nuitka/nodes/ImportNodes.py
```diff
@@ -39,6 +39,7 @@
 from nuitka.importing.ImportResolving import resolveModuleName
 from nuitka.importing.Recursion import decideRecursion
 from nuitka.importing.StandardLibrary import isStandardLibraryPath
+from nuitka.ModuleRegistry import addMissingModule
```
kayhayen, line 42: That approach will never work well with caching, missing modules must be attached to a module. We will want to know what module, and where (source_ref), missed a module. In the case of caching we need to restore that information.

kayhayen:
This needs a rebase on current develop, I don't think I did a rebase, so this is strange, maybe mishandling git on your side?

nuitka/BytecodeCaching.py
```diff
@@ -83,24 +84,26 @@ def getCachedImportedModulesNames(module_name, source_code):
...
+    return [ModuleUsageAttempt(**module) for module in used_modules]
 
 
 def writeImportedModulesNamesToCache(module_name, source_code, used_modules):
     cache_name = makeCacheName(module_name, source_code)
     cache_filename = _getCacheFilename(cache_name, "json")
 
+    used_modules = [module._asdict() for module in used_modules]
+    for module in used_modules:
+        module["source_ref"] = module["source_ref"].getLineNumber()
```
kayhayen, line 100: The `source_ref` is not restored it seems. Probably needs similar treatment as the `ModuleName()`.

nuitka/Reports.py
```diff
@@ -85,6 +86,20 @@ def writeCompilationReport(report_filename):
 
         root.append(module_xml_node)
 
+    modules_node = TreeXML.Element("missing-modules")
+
+    for module_name in OrderedSet(
```
kayhayen, line 91: That iteration over an `OrderedSet`  does not really contribute, or does it, could just as well iterate over the list generator, which probably could also be a generator.

nuitka/BytecodeCaching.py
```diff
@@ -24,6 +24,7 @@
 import os
 
 from nuitka.importing.Importing import getPackageSearchPath, isPackageDir
+from nuitka.nodes.ImportNodes import ModuleUsageAttempt
```
kayhayen, line 27: The `ModuleUsageAttempt` should be moved to `nuitka.importing.Importing`, it normally is never necessary to import nodes to do something with data received.

nuitka/nodes/ImportNodes.py
```diff
@@ -1002,11 +1012,12 @@ def _attemptFollow(self, module_name):
 
                     if name_import_module_filename is not None:
                         self.used_modules.append(
-                            (
+                            ModuleUsageAttempt(
```
kayhayen, line 1015: Using keyword arguments would be nicer.

nuitka/nodes/ImportNodes.py
```diff
@@ -1002,11 +1012,12 @@ def _attemptFollow(self, module_name):
...
                                 name_import_module_name,
                                 name_import_module_filename,
                                 name_import_finding,
                                 1,
+                                None,
```
kayhayen, line 1020: We would have a `source_ref` here, do we not?

nuitka/nodes/ImportNodes.py
```diff
@@ -1015,9 +1026,19 @@ def _attemptFollow(self, module_name):
...
                 module_name = module_name.getPackageName()
 
                 if module_name is None:
+                    self.used_modules = [
+                        ModuleUsageAttempt(
+                            backup_module_name,
+                            module_filename,
+                            self.finding,
+                            level,
+                            None,
```
kayhayen, line 1039: See above, setting source_ref would be very good here too. Also this is overwriting `self.used_modules` rather than adding to it. It should be a list of everything attempted during the pass, not just one.

nuitka/tree/Building.py
```diff
@@ -1004,24 +1005,14 @@ def _loadUncompiledModuleFromCache(module_name, is_package, source_code, source_
...
-                used_module_name,
-                module_filename,
-                finding,
+            ModuleUsageAttempt(
+                module.module_name,
+                module.filename,
+                module.finding,
                 0,
-                source_ref.atLineNumber(line_number),
+                source_ref.atLineNumber(module.source_ref),
```
kayhayen, line 1015: That is probably just wrong, atLineNumber takes an integer. Are we not already getting ModuleUsageAttempt objects. If you solve the persistance better, there is nothing to convert here. However, you removed the new locate. We cannot use the cache result, rather we need to check if it is still there, or still not there.

nuitka/tree/Operations.py
```diff
@@ -73,18 +74,13 @@ def onEnterNode(self, node):
...
         if node.isExpressionBuiltinImport():
 
-            for (
-                used_module_name,
-                used_module_filename,
-                finding,
-                level,
-            ) in node.getUsedModules():
+            for module in node.getUsedModules():
                 self.used_modules.add(
```
kayhayen, line 78: Same here, you are getting attempt objects to create new ones, which should be unneccessary.

nuitka/nodes/ImportNodes.py
Fire-Cube: yes

nuitka/BytecodeCaching.py
Fire-Cube: it becomes restored in` _loadUncompiledModuleFromCache` `source_ref.atLineNumber(module.source_ref)`

nuitka/BytecodeCaching.py
```diff
@@ -83,24 +84,29 @@ def getCachedImportedModulesNames(module_name, source_code):
...
+
     data = {
         "file_format_version": _cache_format_version,
         "module_name": module_name.asString(),
         # We use a tuple, so preserve the order.
-        "modules_used": tuple(
-            (used_module_name.asString(), source_ref.getLineNumber())
-            for used_module_name, _filename, _finding, _level, source_ref in used_modules
-        ),
+        "modules_used": used_modules,
```
kayhayen, line 109: Glaring inconsistency we have had there.

kayhayen (approved):
Very good job

kayhayen:
I am going to merge this one, right after I stabilized my current changes. One thing that I feel is missing is that we should save the line numbers to the data file, they are currently more or less lost when going though cache, which is not a big issue, but still not as nice as it should be. Maybe we also should not put a source ref, but the module_name and line number of the code making the attempt, which then of course would have to resolved to a filename later, should we want to output it, but for persistence, filenames are kind of bad.

nuitka/BytecodeCaching.py
```diff
@@ -83,24 +84,29 @@ def getCachedImportedModulesNames(module_name, source_code):
     if data["module_name"] != module_name:
         return None
 
-    return [
-        (ModuleName(used_module_name), line_number)
-        for (used_module_name, line_number) in data["modules_used"]
-    ]
+    used_modules = list(data["modules_used"])
```
kayhayen, line 87: Why would we convert to a list here, elements of a tuple can be modified still, if it's their properties only

nuitka/Reports.py
kayhayen: Converting a list to an Orderset that is then iterated over has no effect, really.

PR #1978, opened 2023-01-05
compile time optimizations

Fire-Cube:
# What does this PR do?

- reduce call of `isdir` in `getModuleNameAndKindFromFilename`

This PR is unfinished

### DIFF
--- a/nuitka/importing/Importing.py
+++ b/nuitka/importing/Importing.py
@@ -141,11 +141,12 @@ def getModuleNameAndKindFromFilename(module_filename):
     """
 
     # TODO: This does not handle ".pyw" files it seems.
-    if os.path.isdir(module_filename):
-        module_name = ModuleName(os.path.basename(module_filename))
-        module_kind = "py"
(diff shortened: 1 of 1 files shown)


nuitka/importing/Importing.py
```diff
@@ -141,12 +141,13 @@ def getModuleNameAndKindFromFilename(module_filename):
...
-        module_name = ModuleName(os.path.basename(module_filename))
-        module_kind = "py"
-    elif module_filename.endswith(".py"):
+    if module_filename.endswith(".py"):
         module_name = ModuleName(os.path.basename(module_filename)[:-3])
         module_kind = "py"
+
+    elif os.path.isdir(module_filename):
+        module_name = ModuleName(os.path.basename(module_filename))
+        module_kind = "py"
```
kayhayen, line 150: Maybe if the branches became "return driven", there wouldn't have to be elif sequences, and the final `else` would not impose a limit on where you can easily move this branch. I will just do that myself I guess.

PR #1993, opened 2023-01-11
Added sqlalchemy implicit dependecies

interseba5:
# What does this PR do?
This pull request add two implicit dependencies of the sqlalchemy package

A compiled exe that use sqlalchemy with pyodbc or use the default_comparator module raise an import error without this


kayhayen:
Just a small tip, you should use the autoformat as indicated in the template. It also formats strings, etc. in the yaml and does some sorting. Not a big deal for me to do it afterwards though, but with a commit hook it's automatic.

Issue #1990, opened 2023-01-09
Removing numba from shap


interseba5:
* Nuitka version, full Python version, flavor, OS, etc. as output by *this*
  command.

  > python -m nuitka --version
```
1.3.6
Commercial: None
Python: 3.9.5 (tags/v3.9.5:0a7dcbd, May  3 2021, 17:27:52) [MSC v.1928 64 bit (AMD64)]
Flavor: CPython Official
Executable: C:\Python39\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 10
Version C compiler: C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.29.30133\bin\HostX64\x64\cl.exe (cl 14.2).
```

* How did you install Nuitka and Python

Installed using pip

* The specific PyPI names and versions

pip 22.0.4 from c:\python39\lib\site-packages\pip (python 3.9)

  > python -m pip freeze

```
shap==0.41.0
numba==0.56.2
```

* Also supply a [Short, Self Contained, Correct, Example](https://sscce.org/)

```Python
import shap

print("Hello world")
```
* Provide in your issue the Nuitka options used
```
python -m nuitka  --onefile --noinclude-numba-mode=nofollow test.py
```

### Explanation
When i try to compile the script without the `--noinclude-numba-mode=nofollow` option i obtain this warning:
```
Nuitka-Plugins:WARNING: options-nanny: Using module (numba (0, 56, 2)) with no support: Numba is not yet working with
Nuitka-Plugins:WARNING: Nuitka standalone, recommended to use '--noinclude-numba-mode'
```
So i have tried to add the `--noinclude-numba-mode=nofollow` options, because numba is not supported, but when i execute the compiled exe i obtain an import error in the shap\utils\\_clustering.py file.
This error is given by this import in the file:
```Python
from numba import jit
...
@jit
def _pt_shuffle_rec(i, indexes, index_mask, partition_tree, M, pos):
```
Further investigating i have found out that numba is imported also in other shap file that are shap/explainers/_exact.py, shap/links.py, shap/maskers/_image.py, shap/maskers/_tabular.py and shap/utils/_masked_model.py.
I have seen that in some recent commit you have removed the numba dependecies from some packages, so is it possible to remove numba also from shap in the compiled exe?

kayhayen:
Yes, do you want to give it a try: https://nuitka.net/doc/nuitka-package-config.html#anti-bloat

[...]

PR #1994, opened 2023-01-11
Removed numba references from shap

interseba5:
# What does this PR do?
This PR remove all numba references from shap packages

Numba is not currently supported in standalone mode. So if you try to run with --noinclude-numba-mode=nofollow it will raise an import error. More info in issue #1990.


kayhayen:
I had a severe case of the flu, which kept me away from Nuitka for a couple of days. Good job. I will only adapt the descriptions to match the general scheme, but this looks awesome as it is. I am going to make a 1.3.8 with your two PRs these days.

Issue #1996, opened 2023-01-12
Check for zstandard before compilation when using --onefile


Duckle29:
Hi there, and thanks for an amazing bit of software :)

I'm packaging a small app that uses pandas, and as such compilation takes a while.  
Because of this it'd be nice if Nuitka checked for optional dependencies as the first thing, like it does with `oredered-set`

A quick warning with `"Won't be able to compress onefile executable as zstandard dependency isn't installed"` or similar would be nice :)

```
python -m nuitka --version
1.3.7
Commercial: None
Python: 3.10.9 (tags/v3.10.9:1dd9be6, Dec  6 2022, 20:01:21) [MSC v.1934 64 bit (AMD64)]
Flavor: CPython Official
Executable: C:\Users\mjep\AppData\Local\pypoetry\Cache\virtualenvs\mtf-criticals-lMkUpbCG-py3.10\Scripts\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 10
Version C compiler: C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.34.31933\bin\HostX64\x64\cl.exe (cl 14.3).
```

### How did you install Nuitka and Python
Installed into app venv via py poetry

### Options

using --standalone and --onefile

kayhayen:
You are more than welcome to create the PR that implements it.

Duckle29:
Seems solved and merged here https://github.com/Nuitka/Nuitka/pull/2004 Thanks :)

Duckle29:
Woops, no closing. forgot

kayhayen:
Indeed, and 1.4 was released with it, but this was never tracked through develop.

PR #2002, opened 2023-01-16
Added check for zstandard before compilation when onefile

interseba5:
Added check for zstandard before compilation when using --onefile as requested in #1996. The code used to check is taken from ./freezer/Onefile.py

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,8 +1,8 @@
-nuitka (1.4~rc9+ds-1) unstable; urgency=medium
+nuitka (1.4~rc10+ds-1) unstable; urgency=medium
 
   * New upstream pre-release.
 
- -- Kay Hayen <kay.hayen@gmail.com>  Wed, 11 Jan 2023 08:02:46 +0100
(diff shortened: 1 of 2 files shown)


nuitka/Options.py
```diff
@@ -871,6 +872,27 @@ def commentArgs():
 Error, to compile a package, specify its directory but, not the '__init__.py'."""
             )
 
+    if isOnefileMode():
+        zstandard_supported_pythons = ("3.5", "3.6", "3.7", "3.8", "3.9", "3.10")
```
kayhayen, line 876: This reaks of duplication, maybe use this via a local import (likely a cyclic dependency would otherwise trigger) from the `nuitka.freezer.onefile`, or make that call in `nuitka.MainControl`. I would like it for the warnings to come from the onefile logger, so a call of `getCompressorPython` will be all good. The found python is cached anyway.

kayhayen (changes requested):
Good job, but code duplication is not nice and can be avoided.

PR #2004, opened 2023-01-18
Added check for zstandard before compilation when onefile

interseba5:
I have closed PR #2002 by error. In this pull request i have made the changes that @kayhayen suggested in that PR.


kayhayen:
Thanks, I think the autoformat wants to do stuff still, but I can take it from here. Honestly, I am surprised how easy it is in the end. Going to add a comment to why it's called and maybe something that will avoid the duplicate

kayhayen:
hah, you pushed while I was typing... :)

Issue #2006, opened 2023-01-19
AntiBloat patch when no_docstring


interseba5:
* Nuitka version, full Python version, flavor, OS, etc. as output by *this*
  command.

  > python -m nuitka --version
```
1.3.6
Commercial: None
Python: 3.9.5 (tags/v3.9.5:0a7dcbd, May  3 2021, 17:27:52) [MSC v.1928 64 bit (AMD64)]
Flavor: CPython Official
Executable: C:\Python39\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 10
Version C compiler: C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.29.30133\bin\HostX64\x64\cl.exe (cl 14.2).
```

* How did you install Nuitka and Python

Installed using pip

### Explanation

Hi, 
i have a similar problem as #1855 for the xgboost package.
So i was trying to follow @kayhayen advice by patching the code with the anti-bloat plugin.
I have write this in the yaml config:
```yaml
- module-name: 'xgboost.sklearn'
  anti-bloat:
    - description: 'remove docstring dependency' 
      replacements-plain:
        'assert XGBModel.fit.__doc__ is not None' : 'XGBModel.fit.__doc__ = "Fit gradient boosting model"'
      when: 'no_docstrings and no_asserts'
```
But when i try to compile i obtain the following error:
```
FATAL: anti-bloat: Identifier 'no_docstrings' in 'when' configuration of module 'xgboost.sklearn' is unknown
```
Am i missing something?

kayhayen:
Very interesting. Are you saying you are using these options? Currently, the Python flags as given with --python-flag=no_asserts are not exposed, but indeed, there is no reason to not add them. I will do so right away.

kayhayen:
Thanks for your report, this is fixed on the factory branch, which is a development version under rapid development. You can try it out by going here: https://nuitka.net/doc/factory.html 

This has the tags you proposed added, and also `no_annotations` which is however very flaky etc. but one day, these can be per module decisions, then it will useful too.

kayhayen:
Makes me really happy.

kayhayen:
This is part of the stable release 1.5 that I just made.

PR #2007, opened 2023-01-19
Anti-bloat: Removed xgboost docstring dependencies

interseba5:
# What does this PR do?
This PR removes docstrings dependecies from xgboost.sklearn package only if python flags no_docstrings and no_asserts are given.

It was initiated because xgboost don't work with -OO python flag. Also described in #2006.


kayhayen:
Before merging, I need to put the necessary change to develop first. But looks perfect to me.

kayhayen:
On closer inspection, this was probably already on develop what was needed, but hey, memory is not what it used to be in my age group.

Issue #2016, opened 2023-01-27
PyQt5*.py standalone tests failing: AttributeError: 'tuple' object has no attribute 'discard'


jayvdb:
Running any of the standalone tests PyQt5*.py fail locally with the following.  The value of `removed` is `/lib64/libKF5Archive.so.5`.

However it looks like [`removeDllDependencies`](https://github.com/Nuitka/Nuitka/blame/08b799fcf7db345c303aa811d532d59205f25494/nuitka/plugins/Plugins.py#L585)  is in need of some maintenance.  The docstring says it "Returns" something, but it doesnt use `return`.  It used to `return` but stopped in https://github.com/Nuitka/Nuitka/commit/ea4aef0fb0fc1787a4b3cef8e9d50f0f3c09dee3.  Now the arg `dll_filenames` is overwritten inside the function, so it cant be used to modify the list that the caller has access to.  In effect, this function appears to be a no-op.  patch coming.

```
...
Comparing output of 'PyQt5Plugins.py' using '/usr/bin/python3' with flags silent, expect_success, --standalone, remove_output, cpython_cache, timing, --nowarn-mnemonic=debian-dist-packages, ignore_warnings ...
--- /usr/bin/python3 (stdout)
+++ nuitka (stdout)

@@ -1 +0,0 @@

-[PyQt5.QtCore.QByteArray(b'bmp'), PyQt5.QtCore.QByteArray(b'cur'), PyQt5.QtCore.QByteArray(b'gif'), PyQt5.QtCore.QByteArray(b'ico'), PyQt5.QtCore.QByteArray(b'jpeg'), PyQt5.QtCore.QByteArray(b'jpg'), PyQt5.QtCore.QByteArray(b'pbm'), PyQt5.QtCore.QByteArray(b'pgm'), PyQt5.QtCore.QByteArray(b'png'), PyQt5.QtCore.QByteArray(b'ppm'), PyQt5.QtCore.QByteArray(b'svg'), PyQt5.QtCore.QByteArray(b'svgz'), PyQt5.QtCore.QByteArray(b'xbm'), PyQt5.QtCore.QByteArray(b'xpm')]
--- /usr/bin/python3 (stderr)
+++ nuitka (stderr)

@@ -0,0 +1,14 @@

+Traceback (most recent call last):
+  File "/home/jayvdb/python/Nuitka/nuitka/__main__.py", line 151, in <module>
+    main()
[...]

PR #2018, opened 2023-01-27
Plugins: Fix removeDllDependencies

jayvdb:
Fixes https://github.com/Nuitka/Nuitka/issues/2016


kayhayen:
I see, this was broken in the rewrite of the dependency parser, the call however would not use the return value, so this would not have the intended effect it seems, even if it doesn't crash.

jayvdb:
I'll update `Standalone.py` and see how it goes.

jayvdb:
@kayhayen is this what you had in mind?

kayhayen:
Using set allows the order to change. We should replace it with OrderedSet therefore. Otherwise DLL scanning code becomes indemistic. I once had that before and it can be very annoying for debugging.

kayhayen:
Thanks, this too is going to be in 1.4.1 thanks for your help.

Issue #2013, opened 2023-01-27
PandasUsing failing locally on openSUSE due to local additional packages


jayvdb:
On `main` (will retry on `develop` soon)...

 I initially had two separate fatal errors which I could resolve by removing the packages concerned.

```
FATAL: Error, forbidden import of 'setuptools_scm' in module 'pyarrow' at '/usr/lib64/python3.10/site-packages/pyarrow/__init__.py:44' encountered.
```

```
+FATAL: Error, forbidden import of 'numba' in module 'pandas.core._numba.kernels.var_' at '/usr/lib64/python3.10/site-packages/pandas/core/_numba/kernels/var_.py:11' encountered.
```

With those out of the way, I get this very strange error

```
Consider output of standalone mode compiled program: PandasUsing.py
Comparing output of 'PandasUsing.py' using '/usr/bin/python3' with flags silent, expect_success, --standalone, remove_output, cpython_cache, timing, --nowarn-mnemonic=debian-dist-packages, plugin_enable:no-qt, plugin_disable:pylint-warnings ...
--- /usr/bin/python3 (stdout)
+++ nuitka (stdout)

@@ -1 +1,3 @@

+its use. As a result an "ImportError" might be given at run time. Uninstall
+it for full compatible behavior with the uncompiled code to debug it.
 1.5.2
Updating CPython cache by force due to non-matching comparison results.
--- /usr/bin/python3 (stdout)
+++ nuitka (stdout)

@@ -1 +1,3 @@

+its use. As a result an "ImportError" might be given at run time. Uninstall
+it for full compatible behavior with the uncompiled code to debug it.
 1.5.2
Error, results differed (stdout).
...
```

Somehow the first line of this error message is being removed, rendering the error not very helpful to understand the problem.
https://github.com/Nuitka/Nuitka/blob/a7d12c7650ed5b26b6b214023d493ad54304f9f1/nuitka/plugins/standard/PySidePyQtPlugin.py#L1042
Fixing that error reporting is IMO the most important aspect of this, so then I can find out what else needs to be removed in order to be able to run the tests.

kayhayen:
[...]

PR #2022, opened 2023-01-28
Tests: Ignore warnings from PandasUsing

jayvdb:
Closes https://github.com/Nuitka/Nuitka/issues/2013


kayhayen:
See my comment, this is the "wrong" solution. Instead, the "info" given should be properly recognized, which it will be with my recent changes.

jayvdb:
ok

PR #2023, opened 2023-01-29
Update Python auto-formatting

jayvdb:
# What does this PR do?

- Bump isort version
- Improve isort invocation, using self-explanatory args, including fixing possible bug of "-ot" and"-tc" being used instead of "--ot" and "--tc" which are the correct syntax according to `--help` and the docs.
- Reduce number of files not included in auto-formatting

- The isort version is quite old now.
- There were a lot of files excluded from auto-formatting that don't need to be excluded, most can be solved by using  inline `# isort:start` to avoid a long exception list.

# PR Checklist

      cover enough.


tests/benchmarks/pystone3.py
```diff
@@ -57,56 +57,60 @@
                 making the benchmark an integer benchmark again.
 
 """
+from time import time
```
jayvdb, line 60: I wasnt sure whether to explicitly exclude this file.  It is authored by this project, but these changes will make it harder to correlate with the pystone.py which it is based on.

tests/packages/sub_package/kitty/speak/hello.py
```diff
@@ -17,9 +17,13 @@
...
 #     limitations under the License.
 #
-def speak():
-    print( "hello kitty" )
 
 # Test Issue#115, in recursing modules, this was misbehaving.
 import types
+
+
+def speak():
```
jayvdb, line 25: I am a bit unsure about whether this change makes the test no longer test what it is supposed to.

tests/benchmarks/pystone3.py
kayhayen: Such changes are unlikely at this point, plus what one can always do, is to format the upstream. But in principle you are right.

tests/packages/sub_package/kitty/speak/hello.py
kayhayen: Nah it's ok, what matters is that the type check with `==` is working fine. I tacked this on here apparently, to get this checked in module module, to not need another test, but the import does not matter where it is.

tests/programs/cyclic_imports/cyclic_importing_package/Child2.py
```diff
@@ -17,6 +17,8 @@
 #
 print("Child2:", __name__)
 
+# isort:start
```
kayhayen, line 20: Excellent that you caught this and figured out how to solve that. It's not too well documented I have to admit.

tests/benchmarks/constructs/CallCompiledClassCreationPosArgsConstant6.py
```diff
@@ -19,31 +19,37 @@
...
     C("some", "random", "values", "to", "check", "call")
     C("some", "other", "values", "to", "check", "call")
     C("some", "new", "values", "to", "check", "call")
 
-# construct_alternative
+    # construct_alternative
     pass
+
+
 # construct_end
```
kayhayen, line 50: Here the auto-format misses the indentation. This stupid pre-processor of mine is something that ought to be replaced with Jinja2 one day soon. The support for Python in the Jinja2 mode is very excellent, and there not being a good way to edit them otherwise was part of the motivation to do it. Luckily this will map `if else endif` of Jinja2 pretty easily.

nuitka/tools/quality/auto_format/AutoFormat.py
```diff
@@ -646,7 +665,8 @@ def autoFormatFile(
                     else:
                         copyFile(tmp_filename, filename)
 
-            changed = True
+            if filename != "nuitka/build/static_src/HelpersDictionariesGenerated.c":
```
jayvdb, line 668: something strange is going on with this file. It says it is changed, when there are no changes.

tests/benchmarks/constructs/CallCompiledClassCreationPosArgsConstant6.py
jayvdb: I believe the change in indentation here is ok - the code looks for this anywhere on the line.

nuitka/tools/quality/auto_format/AutoFormat.py
```diff
@@ -59,6 +59,30 @@
 
 from .YamlFormatter import formatYaml
 
+# black no longer supports Python 2 syntax, and sometimes removes import
```
kayhayen, line 62: Actually, and I checked with `ClassesTest` now, the tests either should be portable already, or named with `_2`. In that e.g. the `__metaclass__` syntax is used, some of these are just so old, they still contain print statements, because I was using Python2 for the comparison benchmark `GeneratorFunctionVsGeneratorExpression` at the time. This one has no value anymore I suspect.

I guess, I will take this list as a kind of bug report, e.g. `tests\benchmarks\constructs\RichComparisonConditionStrings.py` is not valid Python for no good reason, adding a `pass` would make it work.

With changing to Jinja2, however, I think we loose ability to auto-format anything, is that correct, or is there black ways to do it still. Maybe with tricks like changing Jinja2 annotations to comments and back.

kayhayen:
I am amazed to see this, also the level of quality performed. To me it almost seems perfect to merge. I will try and take note of a few actions. Like reducing the black skip list further, migrating to Jinja2 for the constructs. It seems your isort changes actually clean up stuff. Thanks for the dedication to quality here.

tests/benchmarks/pystone3.py
kayhayen: Actually, now that I saw that `pystone.py` is still excluded, I remembered that I actually did that Python3 port at some point myself, I doubt anybody else ever cared to port it, so you can lift that branch too, doesn't really matter anymore.

nuitka/tools/quality/auto_format/AutoFormat.py
kayhayen: There are whitespace only changes. It plagues me very much myself. I think on GitHub it auto-formats wrong with the clang-format used. I suspect that a new runner must be used or something like that. My Debian Buster also does it wrong. Newest Ubuntu and Windows do it correctly. So bugs in clang-format are very annoying.

Since the files are generated, they very much need auto-format, and cannot just be ignored. I would hate for the state of the file to toggle back and forth depending on who last generated it on what platform. I might have to figure out which clang-format it is, and outright reject updating the file unless it's the correct one, or something like that.

Very bad situation and unresolved. Saying it never changed, I am not sure what the impact is, maybe it then does not get replaced when generating it anew.

nuitka/tools/quality/auto_format/AutoFormat.py
jayvdb: It might be appropriate to rename any py2 specific files to `*_py2.py` (if py26 is still supported), or `*_27.py` so it is very clear - then the black invocation code can even check that the matching files are not valid python3 code.  Then the list of explicit oddities are tests which black would reformat that it shouldnt, such as the unicode string prefix test, and I guess that a "black" inline marker could tell black to ignore those changes - iirc there is a well maintained fork of black which allows inline ignore markers.

tests/benchmarks/pystone3.py
jayvdb: > ... can lift that branch too,

Do you mean remove it from the exclusions, and reformat it, and push an update to this PR?
If so, any other changes I should make?

nuitka/tools/quality/auto_format/AutoFormat.py
kayhayen: Yes, that suffix surely means, it's Python2 only, and most often due to syntax. In some cases, the test framework migrates these on the fly with `2to3` and then it's good like that too. But that is in my mind, actually not a very good idea anyway.

nuitka/tools/quality/auto_format/AutoFormat.py
kayhayen: I just checked, and on my Debian clang-format-8 is used, and does it wrong. On my Ubuntu, a fallback I added at some point is used, namely to use VSCode extension:

```
    # Search Visual Code C++ extension for LLVM path.
    if not clang_format_path:
        for extension_path, extension_filename in listDir(
            os.path.expanduser("~/.vscode-server/extensions")
        ):
            if extension_filename.startswith("ms-vscode.cpptools-"):
                with withEnvironmentPathAdded(
                    "PATH",
                    os.path.join(extension_path, "LLVM/bin"),
                ):
                    clang_format_path = getExecutablePath("clang-format")

                break
```

I guess, that is fine, but we need something like a version check in the end, which makes this code less liberal:

```
    clang_format_path = (
        getExecutablePath("clang-format-12")
        or getExecutablePath("clang-format-12")
        or getExecutablePath("clang-format-11")
        or getExecutablePath("clang-format-10")
        or getExecutablePath("clang-format-9")
        or getExecutablePath("clang-format-8")
        or getExecutablePath("clang-format-7")
        or getExecutablePath("clang-format")
    )
```

kayhayen: For developers, I would say that the VS Code should not be the fallback, but the first thing tried. But on GitHub Actions, the extension may not be easy to have, so using installed clang-format of proper version should still be good. Problem remains, current runner, `ubuntu-20.04` doesn't have them. But a newer LTS will probably have it. I will make these changes after merging your stuff here though, as you can see, it's already broken as is.

jayvdb: On the latest Linux workers, it is possible to install a specific and recent version of clang-format. c.f. https://packages.ubuntu.com/search?suite=jammy&searchon=names&keywords=clang-format

On the older `focal` workers used in this github actions, the choice of clang-format isnt as good, with clang-format-12 as the highest. c.f. https://packages.ubuntu.com/search?suite=focal&searchon=names&keywords=clang-format

There are packages hosted elsewhere that can be used to install more recent clang-format for `focal`, as I used to do this on C++ projects, but I'd need to dig deeper to find them.

if `clang-format-12` is good enough, it is easy enough to install it almost everywhere, and because it is usually installed as `clang-format-12` binary name, the auto formatter can demand it exists if the PR would contain .c , etc file changes.

kayhayen: For the user, Visual Code will be good enough, and actually in case of not finding, I will recommend the user to install and use it, but to at least have it. As you can see, the code is not following clang-format releases very well anyway. I did that because debian packages do not provide `clang-format` which in the light of this thing, is probably understandable.

Mind you, I never noticed I didn't have clang-format installed on my other machine. I tried out v11, which is what I can have on my Debian, and it didn't work any better than the v8. As long as the GitHub action runner has something compatible to what Visual Code extension has (which I suspect is always latest greated), we can do well, and there is always backports I guess.

kayhayen:
@jayvdb I am very tempted to just click merge on this, anything left from your side, looks all good to me now?

nuitka/tools/quality/auto_format/AutoFormat.py
jayvdb: I checked the various debian distros, and indeed they do not have `clang-format-12` :/  They have several others though. https://packages.debian.org/search?suite=sid&searchon=names&keywords=clang-format

And brew has 15, 11 and 8 https://formulae.brew.sh/formula/clang-format

And chocolatey only has LLVM 15 https://community.chocolatey.org/packages?q=clang-format&moderatorQueue=&moderationStatus=all-statuses&prerelease=false&sortOrder=relevance

Seems this is complicated enough that it is worth a separate issue to discuss possible solutions, fallbacks, etc

jayvdb:
:+1: There are a few TODOs coming out of this, but I think they can be deferred to new issues or PRs.

kayhayen:
I already managed to re-arrange things, such that now Visual Code extension wins everywhere, on Windows the MSVC 2022 fallback will be there for those not having it, and for the GitHub runner, I will now stupidly try and update the Ubuntu version used simply, so that might all be good then on that side, and if that fails, can still try and get a backport installed.

PR #2049, opened 2023-02-11
WIP:  specialize ctypes

Fire-Cube:
# What does this PR do?
Add nodes to generate better ctypes code


ctypes_test.py
```diff
@@ -0,0 +1,4 @@
+import ctypes
```
Fire-Cube, line 1: In the end i will delete it. Only for demonstration

kayhayen: You can place it in the tests folder, probably `CtypesUsing.py` would love it, but of course then you should use a DLL that exists everywhere or guard with `os.name` tests, etc.

nuitka/code_generation/CodeGeneration.py
```diff
@@ -896,6 +896,7 @@ def generateHelpersCode():
         "EXPRESSION_CTYPES_CDLL_REF": generateImportModuleNameHardCode,
         "EXPRESSION_CTYPES_CDLL_SINCE38_CALL": generateCtypesCdllCallCode,
         "EXPRESSION_CTYPES_CDLL_BEFORE38_CALL": generateCtypesCdllCallCode,
+        "EXPRESSION_CTYPES_C_INT_CALL": generateCtypesCIntCallCode,
```
kayhayen, line 899: Usually there is also a REF node. But I think but hard imports these are added automatically. I never remember these things myself.

nuitka/nodes/BuiltinOperationNodeBasesGenerated.py
```diff
@@ -1,20 +1,3 @@
-#     Copyright 2022, Kay Hayen, mailto:kay.hayen@gmail.com
```
kayhayen, line 1: That is actually a weakness of the generation code that bitten me. I only run it where it gets later added, not good for collaboration. It should instead add them for you.

nuitka/nodes/CtypesNodes.py
```diff
@@ -48,3 +49,15 @@ def replaceWithCompileTimeValue(self, trace_collection):
...
 
         return self, None, None
+
+
+class ExpressionCtypesCIntCall(ExpressionCtypesCIntCallBase):
+    """Function reference ctypes.c_int"""
+
+    kind = "EXPRESSION_CTYPES_C_INT_CALL"
+
+    def replaceWithCompileTimeValue(self, trace_collection):
```
kayhayen, line 59: For `ctypes.c_int(5)` values you would need one of these pseudo or actual compile time constant values and use compute them.

I am quoting an example here:

```
        return trace_collection.getCompileTimeComputationResult(
            node=self,
            computation=lambda: self.builtin_spec.simulateCall(given_values),
            description="Built-in call to '%s' computed."
            % (self.builtin_spec.getName()),
        )
```

For the call simulation to work, and produce a constant (or correct exception), quite a bit of work will be necessary. Compile time constants need quite a bit of handling, in code generation, in various constant query functions (mutable, is one, etc), data composer writing them, and in constant blob unstreaming.

There is plenty of examples to look at. I think `sys.version_info` is probably closest, although for searching others will be better. Another function to change is `makeConstantRefNode` needing a case.

Since these are going to be probably a lot of similar ones, maybe we need to make as much as possible of this more generic, but I do not know how really. Maybe go through it once, and then tell us how `c_long` could be made easier.

nuitka/specs/HardImportSpecs.py
```diff
@@ -144,3 +144,5 @@
     ),
     default_count=4,
 )
+
+ctypes_c_int_spec = BuiltinParameterSpec("ctypes.c_int", ("value",), default_count=1)
```
kayhayen, line 148: Checking it, I saw this `class c_int(_ctypes._SimpleCData)` and the bases classes it has, maybe for the data value, we do not have to know very precisely in the class what kind of c types data it is, so `c_long` or not, can be inspected at compile time, if needed, but rather we might be able to treat all values the same, I mean for constant checks (probably `isinstance` needed there), and so on.

So I guess, you try and finish `c_int` and then we generalize it to cover all the types with the same data class from `replaceCompileTimeValue` generated, then things would be super easy.

As it stands, right now this only avoids the use of `ctypes.c_int` attribute lookups, not a whole lot.

Issue #2053, opened 2023-02-13
Feature request: install with extras


bersbersbers:
After `pip install nuitka`, during runtime, Nuitka asks to also install `ordered-set` (for performance, I guess) and `zstandard` (this is even required, I think).

I wonder if it might make sense to be able to install these soft-dependencies with pip, e.g., using extras such that `pip install nuitka[orderedsets,onefile]` installs both dependencies. Otherwise, projects depending on `nuitka` will have to separately document and maintain these dependencies.

kayhayen:
Can you make a PR with  this change, I am not familiar with it, but totally open to it. The problem with these is that they would have to be versioned, as e.g. `orderedset` fails to build without MSVC installed, and `zstandard` is not available everywhere. Please help me here.

bersbersbers:
Sure! See #2058, considering that this is more of a technical demo at this point. Feel free to adapt especially the naming.

kayhayen:
Yes, I commented there already. Super useful indeed. Thanks for the proposal and demo code.

bersbersbers:
Works great on `develop` already, thanks!

bersbersbers:
Or so I though. Even though `zstandard` is installed, it is not found:

> Nuitka-Onefile:WARNING: Onefile mode cannot compress without 'zstandard' module installed.

Not sure if this is an issue with installation or rather Python 3.11 or Windows, both of which I have been trying for the first time with Nuitka:

```
(project_dev) C:\code\project>python -m pip show nuitka && python -m pip install zstandard && echo. > empty.py && python -m nuitka empty.py --onefile
Name: Nuitka
Version: 1.5rc9
Summary: Python compiler with full language support and CPython compatibility
Home-page: https://nuitka.net
Author: Kay Hayen
Author-email: Kay.Hayen@gmail.com
License: Apache License, Version 2.0
Location: C:\Users\bers\.pyenv-win-venv\envs\project_dev\Lib\site-packages
Requires: ordered-set, zstandard
Required-by:
Requirement already satisfied: zstandard in c:\users\bers\.pyenv-win-venv\envs\project_dev\lib\site-packages (0.19.0)
Nuitka-Options:INFO: Used command line options: empty.py --onefile
Nuitka:WARNING: The Python version '3.11' is not officially supported by Nuitka '1.5rc9', but an upcoming release will change that. In the mean time use Python version '3.10' instead or newer
Nuitka:WARNING: Nuitka.
[...]

PR #2058, opened 2023-02-14
Optional requirements in `setup.py`

bersbersbers:

# What does this PR do?

Enable `pip install -e .[all]`, for instance, to install optional dependencies. Names of dependencies TBC, this is more a technical demo.

#2053

Requirements were pulled from `requirements.txt` and diffed against existing ones in `setup.py` - missing ones added as optional.

# PR Checklist

      cover enough.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -430,4 +430,14 @@ Python compiler with full language support and CPython compatibility""",
     # As we do version specific hacks for installed inline copies, make the
     # wheel version and platform specific.
     distclass=BinaryDistribution,
+    extras_require={
+        "all": [
+            "ordered-set; python_version >= '3.10'",
+            "subprocess32; python_version == '2.7'",
+            "zstandard; python_version >= '3.5'",
+        ],
+        "msvc": [
+            "orderedset; python_version >= '3.6' or os.name != 'nt'",
+        ],
+    },
 )


setup.py
```diff
@@ -430,4 +430,14 @@ def has_ext_modules():
...
     # wheel version and platform specific.
     distclass=BinaryDistribution,
+    extras_require={
+        "all": [
+            "ordered-set; python_version >= '3.10'",
+            "subprocess32; python_version == '2.7'",
+            "zstandard; python_version >= '3.5'",
+        ],
+        "msvc": [
+            "orderedset; python_version >= '3.6' or os.name != 'nt'",
```
kayhayen, line 440: For orderedset, it's way more of a mess. Should be more like this:

```
orderedset == 2.0.3 ; os.name != "nt" and sys.platform != "darwin" and python_version < "3.7"
orderedset == 2.0.3 ; os.name == "nt" and python_version >= '3.4' and python_version < "3.7"
orderedset == 2.0.3 ; sys.platform == "darwin" and python_version < "3.7"
ordered-set == 4.1.0; python_version >= "3.7"
```

and it's not for MSVC, but I think I get the idea.

kayhayen:
So, we could offer these names for installation, and `all` would install the recommended dependencies, maybe later e.g. also what it takes to check the schema of user yaml files. To avoid conflicts, I would want to have the versions be more of the `>=` kind rather than `==` or else it may create conflicts.

I will trry it with 1.5 release, but keep it unofficial until I tried it out. I don't think that keeping `onefile` separate is useful, so lets start with `all`.

I think I can take it from here, not sure how useful it is to do it in a PR now.

setup.py
bersbersbers: Sure! (I just copy-and-pasted blindy from `requirements.txt`.)

bersbersbers:
> To avoid conflicts, I would want to have the versions be more of the `>=` kind

Yes!

> I will trry it with 1.5 release

Cool!

> I think I can take it from here, not sure how useful it is to do it in a PR now.

Agreed - continue as it suits you, and feel free to close this PR.

kayhayen:
Ultimately I think I realized, I just should have `install_requires` properly and if people not want them, they should tell pip about that, a `all` only expansion would suffer from most people forever being unaware of it, and having dependencies is not like unhear of.

bersbersbers:
Fully agree. Just note it can hard to tell pip to ignore a dependency (https://github.com/pypa/pip/issues/8076), so people would have to uninstall after the fact (assuming they were able to install in the first place).

PR #2065, opened 2023-02-16
Prevent raising file error when used DLLs and included data files conflict at same folder.

Alihakan001:
# What does this PR do?

When copied DLLs and included data files conflict at same folder, shutil module raises SameFileError and this exception stops Nuitka. Simply, when same file exists at same folder, don't raise an error and continue to generating executable.

# PR Checklist

- [*] Correct base branch selected? Should be `develop` branch.
- [ ] All tests still pass.

### DIFF
--- a/nuitka/utils/FileOperations.py
+++ b/nuitka/utils/FileOperations.py
@@ -824,6 +824,9 @@ def copyFile(source_path, dest_path):
     while 1:
         try:
             shutil.copyfile(source_path, dest_path)
+        except shutil.SameFileError:
+            pass
+       
         except PermissionError as e:
             if e.errno != errno.EACCES:
                 raise
--- a/nuitka/utils/FileOperations.py
+++ b/nuitka/utils/FileOperations.py
@@ -824,7 +824,7 @@ def copyFile(source_path, dest_path):
     while 1:
         try:
             shutil.copyfile(source_path, dest_path)
-        except shutil.SameFileError:
+        except shutil.Error:
             pass
        
         except PermissionError as e:


nuitka/utils/FileOperations.py
```diff
@@ -824,6 +824,9 @@ def copyFile(source_path, dest_path):
     while 1:
         try:
             shutil.copyfile(source_path, dest_path)
+        except shutil.Error:
```
kayhayen, line 827: Bug hiding?

kayhayen:
I am not convinced this is a solution. I would much rather want to understand how it comes to pass. Nuitka tries hard to not copy code as data files. If there is a conflict, we should detect and report it.

Alihakan001:
Actually, I tried to include some DLL files as data files. When I tried to compile the program with --standalone option, nuitka tried to copy again the dll files which I included before at the same folder and shutil raised an error. I don't know how clean this solution is, but it worked for me.

kayhayen:
I think including DLLs as data files is utterly wrong, and Nuitka ought to not even attempt it. I do not have test coverage for that, but it is how it is supposed to be. Be prepared for things to break once I happen to get this fixed, which I do occasionaly when I see it happen by accident in my uses. But happy it works for you. However please consider making these DLL dependencies shared, like e.g. done in https://nuitka.net/posts/nuitka-package-config-kickoff.html

kayhayen:
I think this is the wrong approach, please file a bug report with what is happening, then we can devise a real solution that doesn't hide the bug in other places.

PR #2070, opened 2023-02-18
add option to profile optimization

Fire-Cube:


nuitka/MainControl.py
```diff
@@ -24,9 +24,12 @@
 
 """
 
+import cProfile
 import os
 import sys
 
+import pyprof2calltree
```
kayhayen, line 31: These imports should be hidden away where they are used, maybe even its own module is warranted, this slows down Nuitka startup for all users, and makes these modules mandatory.

nuitka/MainControl.py
```diff
@@ -937,12 +940,23 @@ def main():
     addIncludedDataFilesFromFileOptions()
     addIncludedDataFilesFromPackageOptions()
 
+    if Options.options.profile_optimization:
```
kayhayen, line 943: This feels like it wants to be a context manager, where this is done, potentially also delaying the imports to there inside a decision, if that is going to do anything at all.

requirements.txt
```diff
@@ -39,3 +39,5 @@ zstandard; python_version >= '3.5'
 pyyaml; python_version == '2.7' or python_version >= '3.6'
 
 Jinja2 >= 2.10.2
+
```
kayhayen, line 42: I put here only stuff absolutely necessary for compiling, so I would remove that here.

nuitka/OptionParsing.py
```diff
@@ -649,6 +649,16 @@
 
 debug_group = parser.add_option_group("Debug features")
 
+debug_group.add_option(
+    "--profile-optimization-time",
+    action="store_true",
+    dest="profile_optimization",
+    default=False,
+    help="""\
+Profile the time of the optimization and stores it in a file named 'callgrind'. Can be opened with qcachegrind. Defaults to off.
```
kayhayen, line 658: Check out `run-valgrind.py`. I personally would be OK with the file being named more sensibly, like `main_filename.callgrind` derived from what we compile, and for us to launch Valgrind directly. Probably also something to move to `nuitka.utils.profiling` I guess, together with the context manager.

kayhayen: It would be cool if the CI would show the callgrind results, imagine being able to tell from a downloaded callgrind file, what kind of improvements we have. Not sure if this can be GitHub though, since compiling SciPy per commit is not really an option. But maybe we could have a separate workflow to make these artifacts, I would look into adding that.

Issue #2115, opened 2023-03-15
Nuitka doesn't include `autograd` and some other dependencies for `pymoo`


M4a1x:
Hi there,

thanks for the awesome program and the elaborate template! First all the requested information:

> How did you install Nuitka and Python? (And `pymoo`. All following commands inside the activated venv)
```bash
> sudo dnf install python3.10
> python3.10 -m ensurepip
> python3.10 -m venv .venv
> source .venv/bin/activate
> python -m pip install nuitka pymoo
```

> Nuitka version, full Python version, flavor, OS, etc. as output by *this* command.
```bash
> python -m nuitka --version
1.5.1
Commercial: None
Python: 3.10.10 (main, Feb  8 2023, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]
Flavor: Fedora Python
Executable: /home/max/Projects/nuitka-test/.venv/bin/python
OS: Linux
Arch: x86_64
Distribution: Fedora 37
Version C compiler: /usr/lib64/ccache/gcc (gcc).
```

> The specific PyPI names and versions (shortened to what I think is relevant)
```bash
> python -m pip freeze
autograd==1.5
matplotlib==3.7.1
Nuitka==1.5.1
numpy==1.24.2
pymoo==0.6.0.1
scipy==1.10.1
```

> The [MRE](https://en.wikipedia.org/wiki/Minimal_reproducible_example) is taken from the [pymoo introduction](https://pymoo.org/getting_started/source_code.html#Source-Code) and placed inside `test.py`
```python
import numpy as np

from pymoo.algorithms.moo.nsga2 import NSGA2
from pymoo.core.problem import ElementwiseProblem
from pymoo.optimize import minimize
from pymoo.visualization.scatter import Scatter

class MyProblem(ElementwiseProblem):

    def __init__(self):
        super().__init__(n_var=2,
                         n_obj=2,
                         n_ieq_constr=2,
                         xl=np.array([-2, -2]),
                         xu=np.array([2, 2]))

    def _evaluate(self, x, out, *args, **kwargs):
        f1 = 100 * (x[0] ** 2 + x[1] ** 2)
        f2 = (x[0] - 1) ** 2 + x[1] ** 2

        g1 = 2 * (x[0] - 0.1) * (x[0] - 0.9) / 0.18
        g2 = - 20 * (x[0] - 0.4) * (x[0] - 0.6) / 4.8

        out["F"] = [f1, f2]
[...]

PR #2147, opened 2023-04-02
Add pymoo implicit dependencies to plugins

M4a1x:
As mentioned in my issue #2115 `pymoo` has some implicit dependencies, namely `autograd` and some of its own cython extensions. This PR adds these implicit dependencies.

I tried to follow the [official guide on package configuration](https://nuitka.net/doc/nuitka-package-config.html) and took some inspiration from other modules in the same file. I hope this is the recommended way and I didn't miss anything - otherwise, I'm happy to make corrections.

Due to the `./bin/autoformat-nuitka-source` a few other comments got indented as well. I hope that's alright.

Should some documentation be updated? Is there a list of supported 3rd party modules?

Cheers,
Max

### DIFF
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -1270,7 +1270,7 @@
 - module-name: 'numpy.random'
   implicit-imports:
     - depends:
-      # These are post-1.18 names. TODO: Once we detect versions of packages, be proper selective here.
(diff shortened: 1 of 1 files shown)


kayhayen:
Looks good to me. I will merge it for 1.5.5 then, it's strange that the comments were moved by your commit, I believe auto-format should have done this before. Certainly looks more correct on the level they are now.

A minor thing is that relative stuff could be done with `".*"` such that it's more obvious it's child packages of the current package, but that is not generally adhered to anyway, and might be something the autoformat should do one day in a consistent mannner.

Thanks for your contribution.

Issue #2215, opened 2023-05-03
FATAL: Error, failed to find path @rpath/QtQuickEffects.framework/Versions/A/QtQuickEffects


dryarullin:
**Error description:**

FATAL: Error, failed to find path @rpath/QtQuickEffects.framework/Versions/A/QtQuickEffects (resolved DLL to /Users/dryarullin/miniconda3/envs/pyqt_new/lib/python3.11/site-packages/PySide6/Qt/qml/QtQuick/Effects/QtQuickEffects.framework/Versions/A/QtQuickEffects) for /Users/dryarullin/miniconda3/envs/pyqt_new/lib/python3.11/site-packages/PySide6/Qt/qml/QtQuick/Effects/libeffectsplugin.dylib, please report the bug.

**nuitka3 --version**

1.5.7
Commercial: None
Python: 3.11.0 (main, Mar 1 2023, 12:49:28) [Clang 14.0.6 ]
Flavor: Anaconda Python
Executable: /Users/dryarullin/miniconda3/envs/pyqt_new/bin/python
OS: Darwin
Arch: x86_64
Version C compiler: /usr/bin/clang (clang).

pip freeze
PySide6==6.5.0

**Answer:**
libeffectsplugin.dylib refers to a non-existent QtQuickEffects. PySide6 is installed from pip without QtQuickEffects.framework. I don't know why it's not included in the package. Perhaps this is a developer error. How can I tell Nuitka not to use it? I tried to use `--noinclude-dlls=libpdfquickplugin.dylib` or `--noinclude-dlls=Pyside6/libpdfquickplugin.dylib` or `--noinclude-dlls=*QtQuickEffects*`, but no one helped me. I also don't use this plugin in my project, but it is still detected by Nuitka.

dryarullin:
**main.py:**

import sys

```
from PySide6.QtGui import QGuiApplication
from PySide6.QtQml import QQmlApplicationEngine


app = QGuiApplication(sys.argv)

engine = QQmlApplicationEngine()
engine.quit.connect(app.quit)
engine.load('main.qml')

sys.exit(app.exec())
```

**main.qml:**
```
import QtQuick
import QtQuick.Controls

ApplicationWindow {
    visible: true
    width: 600
    height: 500
    title: "HelloApp"

    Text {
        anchors.centerIn: parent
        text: "Hello World"
        font.pixelSize: 24
    }
}
```

**Command:**

[...]

PR #2217, opened 2023-05-04
Standalone, macOS PySide6.5.0, skip DLLs

dryarullin:
# What does this PR do?
PR increases the probability of a successful build using PySide6.5.0.

#2215
In addition to one dll added others. The entire list of problematic dlls is returned by the `getListSkippedDLLs` method. I tried to put the logic into the PySide6 plugin.

# PR Checklist

      cover enough.

### DIFF
--- a/nuitka/freezer/DllDependenciesMacOS.py
+++ b/nuitka/freezer/DllDependenciesMacOS.py
@@ -23,6 +23,7 @@ import sys
 from nuitka.containers.OrderedDicts import OrderedDict
 from nuitka.containers.OrderedSets import OrderedSet
 from nuitka.Errors import NuitkaForbiddenDLLEncounter
+from nuitka.plugins.Plugins import getActivePlugins, hasActivePlugin
 from nuitka.PythonFlavors import isAnacondaPython, isNuitkaPython
 from nuitka.PythonVersions import python_version
 from nuitka.Tracing import inclusion_logger
@@ -90,6 +91,7 @@ def detectBinaryPathDLLsMacOS(
         binary_filename=binary_filename,
         paths=paths,
         package_specific_dirs=package_specific_dirs,
+        package_name=package_name,
     )
 
     if recursive:
@@ -141,7 +143,7 @@ def _parseOtoolListingOutput(output):
 
 
 def _resolveBinaryPathDLLsMacOS(
-    original_dir, binary_filename, paths, package_specific_dirs
+    original_dir, binary_filename, paths, package_specific_dirs, package_name
 ):
     # Quite a few variations to consider
     # pylint: disable=too-many-branches,too-many-locals
@@ -201,10 +203,23 @@ def _resolveBinaryPathDLLsMacOS(
             if candidate is not None and os.path.exists(candidate):
                 resolved_path = candidate
 
+        def getRightActivePlugin(required_plugin_name):
+            for active_plugin in getActivePlugins():
+                if required_plugin_name == active_plugin.plugin_name and hasattr(
+                    active_plugin, "getListSkippedDLLs"
+                ):
+                    return active_plugin
+
         if not os.path.exists(resolved_path):
-            # TODO: Make this a plugin decision, to move this from here to PySide6 plugin:
-            if os.path.basename(binary_filename) == "libqpdf.dylib":
-                raise NuitkaForbiddenDLLEncounter(binary_filename, "pyside6")
+            # The name of the plugin must match the name of the package.
+            plugin_name = package_name.lower()
+            if hasActivePlugin(plugin_name):
+                plugin = getRightActivePlugin(plugin_name)
+                if (
+                    plugin
+                    and os.path.basename(binary_filename) in plugin.getListSkippedDLLs()
+                ):
+                    raise NuitkaForbiddenDLLEncounter(binary_filename, plugin_name)
 
             inclusion_logger.sysexit(
                 "Error, failed to find path %s (resolved DLL to %s) for %s, please report the bug."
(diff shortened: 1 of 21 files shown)


nuitka/plugins/standard/PySidePyQtPlugin.py
```diff
@@ -1387,6 +1387,17 @@ def __init__(self, qt_plugins, no_qt_translations):
 Support for PyQt6 is not perfect, e.g. Qt threading does not work, so prefer PySide6 if you can."""
         )
 
+    @staticmethod
```
kayhayen, line 1390: Now these are quite a few. I think I dislike having to provide the full name.

nuitka/freezer/DllDependenciesMacOS.py
```diff
@@ -201,10 +203,23 @@ def _resolveBinaryPathDLLsMacOS(
             if candidate is not None and os.path.exists(candidate):
                 resolved_path = candidate
 
+        def getRightActivePlugin(required_plugin_name):
+            for active_plugin in getActivePlugins():
+                if required_plugin_name == active_plugin.plugin_name and hasattr(
+                    active_plugin, "getListSkippedDLLs"
```
kayhayen, line 209: So what I would do is to have an implementation that returns an empty tuple in the base class of all plugins, then no `hasattr` check is needed. And I would pass `package_name` to the plugin, so it can decide for itself, if it's responsible or not.

nuitka/freezer/DllDependenciesMacOS.py
```diff
@@ -252,38 +267,40 @@ def fixupBinaryDLLPathsMacOS(
...
-                break
-        else:
-            dist_path = None
-
-        if dist_path is None:
-            inclusion_logger.sysexit(
-                """\
-Error, problem with dependency scan of '%s' with '%s' please report the bug."""
-                % (original_location, rpath_filename)
+        for resolved_filename, rpath_filename in rpath_map.items():
```
kayhayen, line 283: I think I am not getting what is going on here, why do we catch and swallow here, when we could just not raise. The DLL is also not forbidden in my mind, which is what that exception was for, but missing. 

I think what we should become able to, is tell outside code, this one wasn't found, but something told us, that's OK, so we remove it maybe in the `install_name_tool` call of things that use it.

kayhayen:
So, to me this looks pretty good already. I think I want there to be a better handling of the situation, like ideally that the uses of this DLL become removed by outside code. That might prove hard. I will be able to try your code this week and get back to you, probably just doing a few adjustments for the control flow.

nuitka/freezer/DllDependenciesMacOS.py
dryarullin: Added a base `getListSkippedDLLs`method, removed`hasattr`check.

nuitka/freezer/DllDependenciesMacOS.py
dryarullin: I got an error if I didn't use this.

```
  File "/Users/dryarullin/reps/Nuitka/nuitka/__main__.py", line 151, in <module>
    main()
  File "/Users/dryarullin/reps/Nuitka/nuitka/__main__.py", line 137, in main
    MainControl.main()
  File "/Users/dryarullin/reps/Nuitka/nuitka/MainControl.py", line 1090, in main
    _main()
  File "/Users/dryarullin/reps/Nuitka/nuitka/MainControl.py", line 1000, in _main
    copyDllsUsed(
  File "/Users/dryarullin/reps/Nuitka/nuitka/freezer/Standalone.py", line 659, in copyDllsUsed
    fixupBinaryDLLPathsMacOS(
  File "/Users/dryarullin/reps/Nuitka/nuitka/freezer/DllDependenciesMacOS.py", line 266, in fixupBinaryDLLPathsMacOS
    had_self, rpath_map = detectBinaryPathDLLsMacOS(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dryarullin/reps/Nuitka/nuitka/freezer/DllDependenciesMacOS.py", line 89, in detectBinaryPathDLLsMacOS
    had_self, resolved_result = _resolveBinaryPathDLLsMacOS(
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dryarullin/reps/Nuitka/nuitka/freezer/DllDependenciesMacOS.py", line 217, in _resolveBinaryPathDLLsMacOS
    raise NuitkaForbiddenDLLEncounter(binary_filename, plugin_name)
nuitka.Errors.NuitkaForbiddenDLLEncounter: ('/Users/dryarullin/miniconda3/envs/pyqt_new/lib/python3.11/site-packages/PySide6/Qt/qml/QtQuick/Effects/libeffectsplugin.dylib', 'pyside6')
```

nuitka/plugins/standard/PySidePyQtPlugin.py
dryarullin: Do you mean the full name of  DLLs or method?

nuitka/freezer/DllDependenciesMacOS.py
kayhayen: I think `fixupBinaryDLLPathsMacOS` ought to be reuse previous results, and not scan again. That is a weakness of the current code. For macOS there is this extra step to do the  "fixup". I need to look into making it so that it will not call `_resolveBinaryPathDLLsMacOS` twice. I believe that would be the proper place to cache.

Thanks for your hard work on this. This is hard stuff.

nuitka/freezer/DllDependenciesMacOS.py
kayhayen: Yes, that's always a good idea. DLLs must not be globally banned or anything, just for the module(s) that a plugin is responsible for. Eventually I want to make it so that modules register for namespaces, and not act globally. Not having that argument will then be an issue.

nuitka/plugins/standard/PySidePyQtPlugin.py
kayhayen: Nah, the OS specific suffix, like `.dylib`. I image that makes plugins enter the business of knowing platform specific DLL suffixes. These should instead be compared in a way that the DLL suffix is stripped and then a match is done. I see how for now, this is macOS only code, but as a plugin interface it will otherwise hinder abstraction.

kayhayen:
I am going to take it from where it's at now, it seems good enough to me now, and my latest pre-release made a rebase necessary, and I might polish it a bit still, wanting to have it in 1.6

kayhayen:
I manually cherry-picked this to factory and changed a few things. First of all, the plugin interface is generic, so no searching is needed anymore and no matching of package names. Second I cleaned up the naming of the API a bit, allowing to accept or deny a DLL, with check if the plugins agree.

One day, these should not be handled as forbidden, but something similar, maybe incomplete. I need to test this myself now if it still works. Thanks for helping with this.

kayhayen:
This is part of the stable release 1.6 that I just made.

Issue #2207, opened 2023-04-29
Nuitka clears files in build/ which cause issues with `versioneer.py`


pyrito:
## Version
```
1.5.6
Commercial: None
Python: 3.8.16 (default, Mar  1 2023, 21:19:10) 
Flavor: Anaconda Python
Executable: /Users/kvelayutham/opt/anaconda3/envs/py38/bin/python
OS: Darwin
Arch: x86_64
Version C compiler: /usr/bin/clang (clang).
```

## Issue Description

I'm currently using Nuitka as part of a build for a project as such:
```python
setup(
    name="test",
    command_options={
        "nuitka": {
            "--module": None,
            "--include-package": [
                "test",
            ],
            "--no-pyi-file": None,
        }
    },
    version=versioneer.get_version(),
    packages=find_packages(
        exclude=[...]
    ),
    install_requires=[... ],
    python_requires=">=3.8",
    license_files=("LICENSE.txt",),
    license="Other/Proprietary License",
    cmdclass=cmdclass,
    build_with_nuitka=True,
)
```

I build it as such: `python setup.py bdist_nuitka clean --all`

I currently use [versioneer](https://github.com/python-versioneer/python-versioneer) to keep track of my package versions, but noticed that my installed package has an invalid version of `0+unknown` (the `__version__` attribute). After digging into this more, I see that versioneer changes `_version.py` when files are getting copied over to `build/` during the wheel generation. After digging into the Nuitka source code, I can see that Nuitka clears all the python files in `build/` and uses the source files instead for the wheel build instead, which is a little tricky for me. Is there a workaround for this? I can't seem to find a clear flag/option that handles this case. I'm able to work around this for now by modifying the source code during the wheel build myself, but would prefer to not do so.

kayhayen:
People are actively asking to call `clean` before building, which does that deletion too, does it not?

However, it sounds like Nuitka is compiling the wrong files, that would be interesting to fix. Check `nuitka\distutils\DistutilCommands.py` and maybe you can enhance it. It calls `addMainScriptDirectory` as a path to search, but otherwise, it will expect `sys.path` to provide the source, and for all I know it does. But quite possibly, we might have to only add the build folder, and might be able to find everything there? That would be great. Not sure about other modules, you e.g. choose to include, but it being the main script path, and not deleting the files until after compilation, would probably go ways.

pyrito:
[...]

PR #2223, opened 2023-05-08
Make Nuitka use the distribution build folder instead of the source files

pyrito:

# What does this PR do?
Makes Nuitka build from the `build` directory instead of the direct source files. Some projects such as versioneer change the composition of certain files during a distribution build. 

See: #2207 
# PR Checklist

      cover enough.


pyrito:
Hi @kayhayen , luckily versioneer.py is self contained source code so there is no need to add any other dependencies. I also truncated it to only have functionality needed to replicate my issue.

kayhayen:
Ah, wow. Can you clarify the license a bit though, I will have to list that for other downstreams.

kayhayen:
Also, tests passing, so I think are good to go with the license question answered, good job!

pyrito:
@kayhayen per the versioneer repo: https://github.com/python-versioneer/python-versioneer, it seems that the license is "License: Public Domain (Unlicense)", so it is effectively in the public domain and thus we don't have to do anything in particular I believe.

kayhayen:
Ok, but lets not be lazy and at least acknowledge it, putting it under Nuitka tests license otherwise, I will do that then on merge.

pyrito:
@kayhayen do you have any more info you can provide about the RecursionError? Do you think it's related to this PR?

kayhayen:
@pyrito I wrote this into the wrong window, sorry for the noise

PR #2226, opened 2023-05-12
dont follow symbolic links that point to other files in dist folder

SimonRelu:
# What does this PR do?

Before, Nuitka would always follow symbolic links when copying data files. This would mean that if there are multiple links that point to a file within the data dir, it would create multiple copies of that file. With this PR, it will detect when symbolic links point to a file within the dir and will instead copy them as a symbolic link decreasing the package size.

# TODOs

- [ ] I'm not familiar with how the final package is being compressed. If needed, we might have to set a flag in the compression tool to not follow symbolic links there as well.
- [ ] Not implemented for onefile

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1648,7 +1648,7 @@ LTO with MSVC with help this a lot, although for big compilations
 
 More ``anti-bloat`` work on more packages rounds up the work.
 
-For macOS specifically, the WebEngine support is cruical to some users,
+For macOS specifically, the WebEngine support is crucial to some users,
 and the new ``--macos-app-mode`` with more GUI friendly default resolves
 long standing problems in this area.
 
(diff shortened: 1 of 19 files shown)


kayhayen:
The `follow_symlinks` option, is not available in Python2. I am going to fix that myself on merge. Has also revealed that `PermissionError` is also a "recent" addition, but both no big deal.

kayhayen:
Ok, taking it from here. Symlinks are not well detected by older Python on Windows, and generally not as widespread there. I am therefore going to limit this to Linux, esp. as on Windows, before a certain Windows 10 version, it requires admin rights to create symlinks, which is very bad. Optimization in onefile format ought to make it such that the size of the onefile binary wouldn't be bigger at least, but I will delay that.

kayhayen:
I have manually merged this change, improving it, but I have not yet implemented the onefile part, it however just duplicates the symlinks.

I have tried and added support for upwards links as well, by making the link target anew as relative, and then checking if it's below the dist folder. Since newer Python only handles symlinks automatically, we use `os.symlink` with derived targets, and it's all good now. Thanks for your contribution, esp. with finding the easiest spot to add this. :)

This is right now on factory, and soon will be on staging (not in my latest push though) and thus 1.6 release, which I expect to happen within less than a week.

PR #2242, opened 2023-05-24
handle if standalone or onefile is specified in pyproject

martinmiglio:
# What does this PR do?
This change would allow for standalone or onefile builds using distutils and pyproject.toml.

Using distutils and pyproject, only module builds were allowed, this extends that functionality.

# PR Checklist

      cover enough.


martinmiglio:
This is not necessarily the best way of going about this, but I tried to maintain existing functionality as best as possible. Looking for feedback...

kayhayen:
It's actually completely wrong. The current code determines the top level packages and then compiles them all. It ignores the entry points entirely. Not sure if they exist at that build time already or not. I think wheels create the runners during installation maybe, not sure.

What this code is to do, is to recognize that an entirely different mode of operation is needed. The top level packages are probably what should be included, with `--include-module` or `--include-package` but a onefile or a dist folder is to be included, either as the runner, or to quickly launch the runner.

It's not about just switching out `--module` for `--onefile` or `--standalone`.

kayhayen:
So you gave up, right?

martinmiglio:
It's not something I can take on right now.

Issue #2253, opened 2023-05-28
(Linux) Patchelf doesn't skip DLL files


Euro-pol:
I have been trying to compile [tls-client](https://pypi.org/project/tls-client/) but it fails due to patchelf trying to patch .dll files. By the way, the same command works perfectly fine on windows and I could compile it, but not on linux.  
tls client version is 0.2.1 
- Nuitka version:
```
1.5.8
Commercial: None
Python: 3.11.3 (main, Apr  5 2023, 15:52:25) [GCC 12.2.1 20230201]
Flavor: Unknown
Executable: /home/user/coding/tls-test/env/bin/python
OS: Linux
Arch: x86_64
Distribution: Arch None
Version C compiler: /usr/bin/gcc (gcc).
```  
The error is `FATAL: Error, call to 'patchelf' failed: ['/usr/bin/patchelf', '--set-rpath', '$ORIGIN/../..', 'main.dist/tls_client/dependencies/tls-client-32.dll'] -> b'patchelf: not an ELF executable'.`  
Basically using --onefile breaks on linux since it tries using **patchelf** on .**dll** files. A potential fix would be skipping .dll files on a linux based system, and only apply it on **.so** files.  
Command used: `python -m nuitka --onefile main.py --follow-imports`

kayhayen:
It seems it is picking up Windows DLLs for no good reasons. I wonder if they are really used or not.

Can you enhance this, so it doesn't match these on non-Windows?

```
- module-name: 'tls_client.cffi'
  dlls:
    - from_filenames:
        relative_path: 'dependencies'
        # TODO: We might want to match OS and arches here closely, so we do not pick up
        # unnecessary arches.
        prefixes:
          - 'tls-client'
```

If you are available to do a PR, this can be fixed by adding to the yaml configuration, described here: [https://nuitka.net/doc/nuitka-package-config.html](https://nuitka.net/doc/nuitka-package-config.html)

And here is more of a walk through style post: [https://nuitka.net/posts/nuitka-package-config-kickoff.html](https://nuitka.net/posts/nuitka-package-config-kickoff.html)

Here are a example PRs for other packages:

[#2030](https://github.com/Nuitka/Nuitka/pull/2030) 
[#1993](https://github.com/Nuitka/Nuitka/pull/1993)

Euro-pol:
[...]

PR #2255, opened 2023-05-28
Fix tls-client different operating system dependencies

Euro-pol:
# What does this PR do?
Fixes tls-client compiling on differents operating systems. Tested this on both linux and windows, they work. Not sure about MacOS but should work too.  
due to issue #2253 (tls-client only compiling on windows)  
# PR Checklist


kayhayen:
Looks perfect to me, I am right now releasing 1.6, kind of locking things, but I will merge it for the hotfix soon after, 1.6.1.

PR #2354, opened 2023-07-25
Fix usage of deprecated sysconfig variable with mingw

Biswa96:
This fixes the following error with mingw python 3.11.

    nuitka/PythonFlavors.py", line 144, in isMSYS2MingwPython
        return "-mingw_" in sysconfig.get_config_var("SO")
    TypeError: argument of type 'NoneType' is not iterable

    SO config variable in sysconfig was removed in python 3.11. It was
    deprecated in Python 3.4 and was suggested for removal in Python 3.5.
    See this cpython commit
    https://github.com/python/cpython/commit/0860b26a4f7abeb61aad9f4b96de8e78eed8c12a

### DIFF
--- a/nuitka/PythonFlavors.py
+++ b/nuitka/PythonFlavors.py
@@ -141,7 +141,7 @@ def isMSYS2MingwPython():
 
     import sysconfig
 
-    return "-mingw_" in sysconfig.get_config_var("SO")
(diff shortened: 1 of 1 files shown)


kayhayen:
Going to merge that for factory and do add it for the 1.7.7 hotfix, that I intend to do shortly. I will most likely also make it check the python version for what variable to check, because I don't want to break old setups and am not inclined to really test that now.

Issue #2356, opened 2023-07-27
Add PEP 660 support?


shkey:
Recently when I was trying to use `pdm` to manage the packaging and build process of our company's project, I encountered an issue when using `nuitka.distutils.Build` as the compile and build backend. After I changed the `build-backend` in `pyproject.toml` to `nuitka.distutils.Build`, when I tried to use `pdm install` to install all the dependencies of the current project and install the current project itself in editable mode into the virtual environment, then used `pdm list` to check the installed dependencies, I found the dependencies were not properly installed in the virtual environment. The package names were shown as `UNKNOWN`(should be pdm-demo), which caused import errors in the code. Here is a simple example:

```bash
❯ pdm --version
PDM, version 2.8.0
❯ mkdir pdm-demo
❯ cd pdm-demo
❯ pdm init
❯ cat pyproject.toml
[project]
    authors         = [{ name = "shkey", email = "shkey325@gmail.com" }]
    dependencies    = []
    description     = ""
    dynamic         = ["version"]
    license         = { text = "MIT" }
    name            = "pdm-demo"
    readme          = "README.md"
    requires-python = ">=3.11"

[build-system]
    build-backend = "nuitka.distutils.Build"
    requires      = ["nuitka", "setuptools", "toml", "versioningit", "wheel"]
❯ pdm install -vv
......
All packages are synced to date, nothing to do.
Installing the project as an editable package...
pdm.termui: Preparing isolated env for PEP 517 build...
pdm.termui: Reusing shared build env: /var/folders/v6/sz7x176j2l10hdr9_j3d_d440000gn/T/pdm-build-env-zbhcceff-shared
pdm.termui: Running PEP 517 backend to build a wheel for <Link file:///Users/shkey/Downloads/pdm-demo (from None)>
pdm.termui: The build backend doesn't support PEP 660, falling back to setuptools-pep660
pdm.termui: Reusing shared build env: /var/folders/v6/sz7x176j2l10hdr9_j3d_d440000gn/T/pdm-build-env-wklmp4eq-shared
  ✔ Install pdm-demo 0.0.0 successful

🎉 All complete!
❯ pdm list
╭─────────┬─────────┬────────────────────────────────────╮
│ name    │ version │ location                           │
├─────────┼─────────┼────────────────────────────────────┤
│ UNKNOWN │ 0.0.0   │ -e /Users/shkey/Downloads/pdm-demo │
╰─────────┴─────────┴────────────────────────────────────╯
[...]

PR #2357, opened 2023-07-27
Add PEP 660 editable install support for nuitka.distutils.Build

shkey:
# What does this PR do?

 - Adds support for PEP 660 editable installs to nuitka.distutils.Build
 
  - To solve the problem that local project packages cannot be installed in editable mode.
  - Related to issue #2356

# PR Checklist

      cover enough.

### DIFF
--- a/nuitka/distutils/Build.py
+++ b/nuitka/distutils/Build.py
@@ -31,6 +31,9 @@ if not hasattr(setuptools.build_meta, "suppress_known_deprecation"):
 else:
     suppress_known_deprecation = setuptools.build_meta.suppress_known_deprecation
 
+SETUPTOOLS_ENABLE_FEATURES = os.getenv("SETUPTOOLS_ENABLE_FEATURES", "").lower()
+LEGACY_EDITABLE = "legacy-editable" in SETUPTOOLS_ENABLE_FEATURES.replace("_", "-")
+
 
 # reusing private "build" package code, pylint: disable=protected-access
 class NuitkaBuildMetaBackend(setuptools.build_meta._BuildMetaBackend):
@@ -52,3 +55,8 @@ get_requires_for_build_sdist = _BACKEND.get_requires_for_build_sdist
 prepare_metadata_for_build_wheel = _BACKEND.prepare_metadata_for_build_wheel
 build_wheel = _BACKEND.build_wheel
 build_sdist = _BACKEND.build_sdist
+
+if not LEGACY_EDITABLE:
+    get_requires_for_build_editable = _BACKEND.get_requires_for_build_editable
+    prepare_metadata_for_build_editable = _BACKEND.prepare_metadata_for_build_editable
+    build_editable = _BACKEND.build_editable


nuitka/distutils/Build.py
```diff
@@ -31,6 +31,9 @@ def suppress_known_deprecation():
 else:
     suppress_known_deprecation = setuptools.build_meta.suppress_known_deprecation
 
+SETUPTOOLS_ENABLE_FEATURES = os.getenv("SETUPTOOLS_ENABLE_FEATURES", "").lower()
+LEGACY_EDITABLE = "legacy-editable" in SETUPTOOLS_ENABLE_FEATURES.replace("_", "-")
```
kayhayen, line 35: I was wondering if this duplicates code from setuptools, but it's not too important I think.

kayhayen (approved):

nuitka/distutils/Build.py
shkey: This is for full compatibility with the `setuptools` configuration, if the user runs with `SETUPTOOLS_ENABLE_FEATURES="legacy-editable"` it may cause `_BACKEND.get_requires_for_build_editable`, `_BACKEND .prepare_metadata_for_build_editable` and `_BACKEND.build_editable` undefined because they are all declared in the `setuptools.build_meta` code under the `if not LEGACY_EDITABLE` condition. Also, since `LEGACY_EDITABLE` is not exported in `setuptools.build_meta`, the only way to get the truth of `LEGACY_EDITABLE` is to add the same two lines as in the `setuptools.build_meta` code here, So I wrote it like this. Or do you have a better way to do it?

kayhayen: Well, not being in `__all__` doesn't prevent usage. So, e.g. the current code does this:

```
if not hasattr(setuptools.build_meta, "suppress_known_deprecation"):
```

I believe that's also for working with older setuptools. In that way, older or newer setuptools decides stuff, and not us. 

So `LEGACY_EDITABLE = getattr(setuptools.build_meta, "LEGACY_EDITABLE")` would probably be more future proof. I will just change it on develop after the hotfix release for 1.8 to do it that way. Most important is that it works more often than before for the 1.7 release.

kayhayen: Like I said, no big thing. But in case they somehow change their mind, it's a good idea. And now e.g. Nuitka would attempt to use it with a setuptools that doesn't do it, but that's going to be rarely possible I guess, but like I said, just going to add that, and ask you to retest, hold on.

shkey: Okay, and I noticed that the commit for this PR that you merged yesterday is missing from the develop branch, so do I need to resubmit a new PR?

kayhayen: It's on its way into 1.7.8, right now being in hotfix/1.7.8 branch, which I intend to release shortly as well.

shkey: That's great, I'm waiting for this release.🥳

PR #2383, opened 2023-08-12
Added optional `sub_collect` to `_detectImports`

### LINKED ISSUE #2384: Collect submodules without follow imports in them.
in standalone mode currently it is blocking to add package folder manually after partial of package included in final executable.
Actually, I need `--collect-submodules` that will include other module files without checking for imports in them, what currently `--include-package` doing is also includes what's imported inside submodules.

This will prevent multiple import search paths (one in binary and one in directory):
```
Module A:
    a.py -> no depends
    b.py -> depends on large package B
```
here `app.py` only imports `A.a` and nuitka only collects this.
`app.py` also in the future would dynamically import `A.b`.
if try to create folder `A` and put `b.py` in executable dir, still you cannot import it, bcz namespace `A` already exists inside of executable (first path in `sys.path`).

I tried `--include-package-data`, I was expecting to include `.py` files, but it didn't.

byehack:
# What does this PR do?
fixes #2384

This will avoid to be multiple path search for interpreter.
for startup packages we should include all submodules not that parts only needed for interpreter (ref: [pyinstaller](https://github.com/pyinstaller/pyinstaller/blob/9bc447f29bbcd164a64f2d5dd3d38ecbcabf9ea2/PyInstaller/depend/analysis.py#L252-L253))

No, I didn't open issue.

# PR Checklist

      cover enough.


nuitka/freezer/ImportDetection.py
```diff
@@ -206,38 +206,16 @@ def _detectImports(command):
...
-    import_code = ";".join(
-        "import encodings.%s" % encoding_name
-        for encoding_name in sorted(encoding_names)
+    encodings_subs = ";".join(
+        "import encodings.%s" % m[1]
+        for m in pkgutil.iter_modules(sys.modules["encodings"].__path__)
     )
 
-    import_code += ";import locale;"
+    import_code = encodings_subs + ";import locale"
```
byehack, line 214: why we add `locale`?

nuitka/freezer/ImportDetection.py
```diff
@@ -39,11 +39,51 @@
 from nuitka.utils.FileOperations import areSamePaths
 from nuitka.utils.ModuleNames import ModuleName
 
+IMPORT_CODE = import_code = """
```
byehack, line 42:
```suggestion
IMPORT_CODE = """
```

nuitka/freezer/ImportDetection.py
```diff
@@ -39,11 +39,51 @@
...
+            del sys.modules[fail]
+"""
+
+
+# TODO: option to collect_submodules
+def _detectImports(imports):
     # This is pretty complicated stuff, with variants to deal with.
     # pylint: disable=too-many-branches,too-many-statements
 
+    command = IMPORT_CODE % imports
```
byehack, line 85:
```suggestion
    command = IMPORT_CODE % (imports,)
```

nuitka/freezer/ImportDetection.py
```diff
@@ -206,40 +246,23 @@ def _detectImports(command):
...
-    ):
-        if non_locale_encoding in encoding_names:
-            encoding_names.remove(non_locale_encoding)
-
-    import_code = ";".join(
-        "import encodings.%s" % encoding_name
-        for encoding_name in sorted(encoding_names)
+    # TODO: `_detectImports(code, collect_submoduls=True)` Could be used instead.
+    imports = tuple(
+        "import encodings.%s" % m[1]
```
byehack, line 251:
```suggestion
        "encodings.%s" % m[1]
```

nuitka/freezer/ImportDetection.py
kayhayen: The `locale` is used to react to the user settings for locale, which may not be used for the config Nuitka is running it. And the `inspect` needs to be monkey patched before it's loaded, to become tolerant of compiled  code. And `importlib` is needed to register Nuitka's loader with the application, without it, cannot register the loader for the package itself, i.e. we cannot load `importlib` if its not already there.

byehack: > And the `inspect` needs to be monkey patched before it's loaded, to become tolerant of compiled code. And `importlib` is needed to register Nuitka's loader with the application, without it, cannot register the loader for the package itself, i.e. we cannot load `importlib` if its not already there.

Correct, I noticed later when I tried to hard-code ignore `inspect`, I got `ModuleNotFoundError` at runtime.



> The `locale` is used to react to the user settings for locale, which may not be used for the config Nuitka is running it.

I couldn't understand, is it only for specified python versions?(I tried to ignore it in py37 and everything are well)

byehack:
IMO it's ready now, for further steps we can have cli option too.
Currently it's on by default for interpreter deps.

kayhayen:
Can you follow this to rebase: https://nuitka.net/doc/developer-manual.html#nuitka-git-github-workflow

I want to take a look, what this changes, and what to take from it, and why.

kayhayen (changes requested):
Needs a rebase to be able to look at it cleanly and compare with what current code does, to see what improvements it gives

kayhayen:
Closing for lack of feedback

PR #2434, opened 2023-09-11
Include pygeos' data directory in standard package config

a-gn:
# What does this PR do?

It tells Nuitka to include the `pyproj/proj_data` data directory using the standard package config file.

I didn't open an issue but this happens at runtime if we build `import pyproj` with `--standalone` on Linux (Nuitka-commercial 1.7.10, pyproj 3.4.1, Python 3.9):

```
ERROR 1: PROJ: proj_create_from_database: Cannot find proj.db
PROJ: proj_create_from_database: Cannot find proj.db
/usr/local/bin/myProgram.dist/pyproj/__init__.py:91: UserWarning: Valid PROJ data directory not found. Either set the path using the environmental variable PROJ_DATA (PROJ 9.1+) | PROJ_LIB (PROJ<9.1) or with `pyproj.datadir.set_data_dir`.
```

Without `proj.db`, `pyproj` can't do its job.

# PR Checklist

  but they might not cover enough. **(This is a small change so I rely on the Actions workflow.)**
- [x] Ideally new features or fixed regressions ought to be covered via new tests. **I didn't include a test for this because I assume `pyproj` won't be installed at test time. Let me know if and where I should add one.**
- [x] Ideally new or changed features have documentation updates. **I'm not really changing any features, but if I need to document this somewhere, please let me know.**


kayhayen:
The comment feels like a TODO, is there anything to be done about said environment variable, or is it how it's overriden, in which case, actually we probably want to disable that ability at some point, then it's just a TODO.

a-gn:
> The comment feels like a TODO, is there anything to be done about said environment variable, or is it how it's overriden, in which case, actually we probably want to disable that ability at some point, then it's just a TODO.

We don't seem to need that here: `PROJ_LIB` is read by the C++ PROJ library, but [the Python package has its own way of telling PROJ where to find the data](https://github.com/pyproj4/pyproj/blob/76b77c8586efa28565aaab2365fa459f75596043/pyproj/datadir.py#L73). (The process [starts here in `pyproj.__init__`](https://github.com/pyproj4/pyproj/blob/76b77c8586efa28565aaab2365fa459f75596043/pyproj/__init__.py#L89).)

In fact I always remove the `PROJ_LIB` variable if it exists when running a Python progra. Either it points to what pyproj could already find by itself, or it points to a different version and that messes with PROJ.

A test that checks that the warning is never sent would be good though, in case other versions or OSes don't work the same way. Just let me know how to add it (pyproj should be installed to run the test).

a-gn:
So: for a simple script containing

```python
import numpy as np
from pyproj import Transformer, transform

transformer = Transformer.from_crs(2263, 4326)
x_coords = np.random.randint(80000, 120000)
y_coords = np.random.randint(200000, 250000)

transform(2263, 4326, x_coords, y_coords)
transformer.transform(x_coords, y_coords)
```
if I build with `python3 -m nuitka --standalone --user-package-configuration-file=nuitka-package-config-file.yml import_pyproj.py`
where `nuitka-package-config-file.yml` contains:
```yaml
- module-name: 'pyproj'
  data-files:
    dirs:
      - 'proj_dir'
```
the error and warning disappears and the result runs correctly.

But in my bigger program (which I can't post here), I still get:
```
ERROR 1: PROJ: proj_create_from_database: Cannot find proj.db
PROJ: proj_create_from_database: Cannot find proj.db
```

a-gn:
I changed my mind: if we can write the proj_data folder somewhere and set PROJ_LIB to it, that should fix everything more simply. It would also ignore PROJ_LIB from the user which is good.

The only downside is that if we compile pyproj version X and set `PROJ_LIB` to pyproj X's data folder, and we then call `subprocess.run(["other_exe"])` and `other_exe` is linked to PROJ version Y, `other_exe` will inherit Python's `PROJ_LIB` and use the wrong PROJ data folder. That might cause problems. Not sure this is in scope for Nuitka to fix though, for example `cv2` changes the `PATH` in a way that messes with OpenCV in subprocesses and that's just accepted.

kayhayen:
Since it just works fine for me with `--include-package-data=pyproj` and --list-package-data looks fine, I am just going to cherry-pick your change for 1.9.2 now.

kayhayen:
One thing I noticed however, is that the data files include `.pyx` files for the extension modules, that is a code kind of data file, that needs to be ignored, and it still seems to work. I am going to credit the commit to you of course.

kayhayen:
So, manually merged this one. Sorry I didn't find the time to do it sooner. You message should not happen. If it does in any configuration that's based on factory branch or 1.9.2, then let me know, and I will look into it.

PR #2438, opened 2023-09-12
Reports: Fix get_distribution_license() for jinja2 templates

henzef:
This fixes the function for cases where the license is None instead of "UNKNOWN".


kayhayen:
This is great, it will be in 1.8.2 once it comes out as well.

Issue #2448, opened 2023-09-20
Package generated by "nuitka.distutils.Build" couldn't be installed to a virtualenv, but "setuptools.build_meta" do.


rushwing:
### Problem description

OS version
MacOS Ventura 13.5.2

Python version
Python 3.11.4

Pip version
pip 23.1.2

Nuitka version
1.8.1

Guide link
https://nuitka.net/doc/user-manual.html#use-case-5-setuptools-wheels

Problem description
I have created an setuptools wheel using Nuitka for a project which was used to parse configurations from json format files. (https://nuitka.net/doc/user-manual.html#use-case-5-setuptools-wheels). I used pyproject.toml as follows, and the only difference compared with using built-in Python setuptools is I used "nuitka.distutils.Build" rather than "setuptools.build_meta" in the "build-backend" section:

```
[project]
name = "serviceconfig"
version = "1.0.0"
description = "Reading service configurations for microservices"
readme = "README.rst"
requires-python = ">=3.8"

[project.scripts]
service_config = "configs.service_config:ServiceConfig"

[build-system]
# requires = ["setuptools>=42", "wheel", "toml"]
requires = ["setuptools>=42", "wheel", "nuitka", "toml"]
# build-backend = "setuptools.build_meta"
build-backend = "nuitka.distutils.Build"

[nuitka]
show-scons = true
include-package = "configs"
nofollow-import-to = ["*.tests", "*.distutils"]
```

and setup.py as follows:

```
from setuptools import setup

setup()
```

and I used cmake script to build the package:

```
include(${CMAKE_CURRENT_LIST_DIR}/env_variables.cmake)

if(DEFINED ENV{VIRTUAL_ENV} AND "$ENV{VIRTUAL_ENV}" MATCHES ".*${VENV_NAME}.*")
    set(PYTHON_EXE_PATH "python")
else()
    set(PYTHON_EXE_PATH ${PYTHON_EXECUTABLE})
endif()


# Silent clean the old build folders if exists
if(CLEAN_BUILD)
    file(REMOVE_RECURSE ${SERVICE_PATH}/build OPTIONAL)
    file(REMOVE_RECURSE ${SERVICE_PATH}/dist OPTIONAL)
    file(REMOVE_RECURSE ${SERVICE_PATH}/drivers.egg-info OPTIONAL)
endif()

if(NUITKA_BUILD)
    message("SYSTEM_PYTHON_PATH = ${SYSTEM_PYTHON_PATH}")
[...]

PR #2451, opened 2023-09-20
Fixing the issue #2448 of incorrect platform tag naming for the wheel package.

shkey:

# What does this PR do?
This PR is to fix the #2448 issue.

Hey, bro, you thought that issue #2448 was unrelated to Nuitka, so you closed it. However, after a period of breakpoint debugging, I discovered that the cause of this issue lies in a flawed logic within Nuitka's code. 
In the code of the wheel package, at https://github.com/pypa/wheel/blob/e76040b0d970a0fe6211c31888fc929078ba02d9/src/wheel/bdist_wheel.py#L252, when this line of code is executed, the variable `self.plat_name` is still `None`. Its value is initialized later at https://github.com/pypa/wheel/blob/e76040b0d970a0fe6211c31888fc929078ba02d9/src/wheel/bdist_wheel.py#L261. Therefore, at this point, the value of `self.plat_name_supplied` in line 252 is definitely `False`. However, in your Nuitka code, after calling `wheel.bdist_wheel.bdist_wheel.finalize_options(self)`, you run `self.plat_name_supplied = self.plat_name is not None`, which causes the value of `self.plat_name_supplied` to be inconsistent with the expected value. So, I submitted this PR to fix this behavior.

# PR Checklist

### DIFF
--- a/nuitka/distutils/DistutilCommands.py
+++ b/nuitka/distutils/DistutilCommands.py
@@ -379,7 +379,6 @@ class bdist_nuitka(wheel.bdist_wheel.bdist_wheel):
         wheel.bdist_wheel.bdist_wheel.finalize_options(self)
         # Force module to use correct platform in name
         self.root_is_pure = False
-        self.plat_name_supplied = self.plat_name is not None
 
     def write_wheelfile(self, wheelfile_base, generator=None):
         if generator is None:


kayhayen:
Woohoo, this comes at a surprise. I was thinking this wasn't possible, seems the time spent there was maybe not wasted after all.

Thanks for your contribution, I will make it part of the next hotfix if any occurs.

kayhayen:
I am moving this to the 1.8.3 hotfix branch, so it will re-appear on develop once that is merged only. You can use that branch to have your change on stable Nuitka right now.

PR #2453, opened 2023-09-21
Fix typos

omahs:
Fix typos


kayhayen:
Thanks for doing this, looks very good to me.

kayhayen:
I am moving this to the 1.8.3 hotfix branch, so it will re-appear on develop once that is merged only.

Issue #2471, opened 2023-10-02
Missing GDK DLL(s) for windows standalone build with msys2


hacklschorsch:
I am trying to package [gtimelog](https://github.com/gtimelog/gtimelog), a Python + GTK application, for Windows.
I got it running fine using [msys2](https://www.msys2.org/docs/python/), but it doesn't run standalone - it seems nuitka cannot find all relevant DLLs.

----

- Nuitka version, full Python version, flavor, OS, etc. as output by *this exact* command.

> python -m nuitka --version

```
$ python -m nuitka --version
1.8
Commercial: None
Python: 3.11.5 (main, Sep 28 2023, 19:43:22)  [GCC 13.2.0 64 bit (AMD64)]
Flavor: MSYS2 MinGW
Executable: C:/msys64/mingw64/bin/python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 10
Version C compiler: C:/msys64/mingw64/bin/gcc.exe (gcc).
```

- How did you install Nuitka and Python

1. I installed msys2 using their installer on Windows 11 pro and following [the msys2 guidelines](https://www.msys2.org/#installation) and [this micro howto to make sure GTK worked](https://stackoverflow.com/questions/33862049/python-cannot-install-pygobject/54795222#54795222).  I used the default paths (`C:\msys64\`).
2. In the msys2-provided mintty terminal emulator & default shell I cloned [gtimelog](https://github.com/gtimelog/gtimelog) into my $HOME
3. To make it run, I had to install a few libraries, most of which I found in the [pygobject docs](https://pygobject.readthedocs.io/en/latest/devguide/dev_environ.html#windows-dep):
```
pacman -S --needed --noconfirm base-devel mingw-w64-x86_64-toolchain git \
   mingw-w64-x86_64-python3 mingw-w64-x86_64-python3-cairo \
   mingw-w64-x86_64-gobject-introspection mingw-w64-x86_64-libffi
```
Additionally - not sure if all of these are required, but I think these are the ones I got (hope I am not missing anything):

```
pacman -S msys2-runtime-devel
pacman -S mingw-w64-x86_64-python-cairo
pacman -S mingw-w64-x86_64-libsecret
pacman -S mingw-w64-x86_64-libsoup
pacman -S mingw-w64-x86_64-python-nuitka
```
4. To get gtimelog to run:
In the cloned repository directory I ran `make run` to create .mo and GSettings schema files.
[...]

PR #2506, opened 2023-10-19
Add GI GTK/GDK/Cairo standalone test

hacklschorsch:
# What does this PR do?

This adds a regression test for 153c0463be and PR #2485.
It is almost identical with the [reproducer @kayhayen posted for #2471](https://github.com/Nuitka/Nuitka/issues/2471#issuecomment-1744525538).

I do not know what file preamble (copyright notices), meta tags (like ignoring warnings, configuring isort...) or changes to dependecy definitions (to make sure the test runners actually have Gtk, Gdk and Cairo) are required unfortunately.
Happy to add if given pointers, but of course please feel free to edit this as you see fit.

Issue #2471 triggered my looking into this.

# PR Checklist

  - Sorry, again used black manually.
  - I cannot run them on my setup still :(
  - This is trying to help with that - providing a test for an already fixed regression.


kayhayen:
It needs a `# nuitka-skip-unless-imports: cairo` or else it's going to fail on GitHub, where it's not installed. We currently only do virtualenv based tests in Nuitka-Watch, where I may add it once it supports MSYS2 as well, i.e. flavors.

hacklschorsch:
No need to have Cairo here really - the test would be almost as useful without it, no?

kayhayen:
You can of course make the `cairo` usage optional, but then `gi` is also not going to be installed everywhere, so lets not do it here. It's ok if the test is skipped in most environments, if private CI has that, or if it'S also added to Nuitka-Watch which I intend to do.

tests/standalone/GiUsing.py
```diff
@@ -0,0 +1,18 @@
...
+
+import cairo
+import gi
+
+gi.require_version("Gtk", "3.0")
+gi.require_version("Gdk", "3.0")
+
+from gi.repository import Gdk, Gtk
+
+# isort:start
```
kayhayen, line 11: It would appear that the isort start  moniker is something you wanted before the gi.repository import?

PR #2541, opened 2023-11-12
few latin1 -> utf-8 encode

shenjackyuanjie:
example broken module:

## this example will pass after this fix PR

[test.zip](https://github.com/Nuitka/Nuitka/files/13327787/test.zip)

## This test.zip will not pass and callout WerFault.exe on windows

[test.zip](https://github.com/Nuitka/Nuitka/files/13328002/test.zip)

### DIFF
--- a/nuitka/plugins/Plugins.py
+++ b/nuitka/plugins/Plugins.py
@@ -1344,7 +1344,7 @@ class Plugins(object):
     def encodeDataComposerName(cls, name):
         if str is not bytes:
             # Encoding needs to match generated source code output.
-            name = name.encode("utf-8")
+            name = name.encode("utf8")
             # for modules like abc.中文 or other need utf-8 to encode module name
 
         for plugin in getActivePlugins():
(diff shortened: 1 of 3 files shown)


kayhayen:
Since C compilers do not really default to UTF8, I really would like to avoid writing source code that is not latin1, or even ascii. That is why I changed the approach somewhat and use our identifier encoding. 

Thanks for your report, this is fixed on the factory branch, which is a development version under rapid development. You can try it out by going here: https://nuitka.net/doc/factory.html 

Feedback if this is working is very welcome, just please do not share plans of doing it, but rather confirmations or denials of it working.

shenjackyuanjie:
badnews
```
  File "V:\githubs\QT-EAS~1\build\test\MAIN~1.DIS\inspect.py", line 39, in <module>
ModuleNotFoundError: No module named 'collections.abc'
```
(with the second test.zip on factory

shenjackyuanjie:
upper issue is now fixed
but the segfault still exist
```
build\test\main.dist [ main][!?][🐍 v3.10.11(env)]
❯ .\main.exe
这里有个测试 here is a test in chinese module name
另一个测试 here is a anthor test in chinese file name
after define
```

kayhayen:
Using this as an input, I however avoided to write UTF8 source code, which found a few other errors in the code.

shenjackyuanjie:
\o/
passed!

PR #2589, opened 2023-12-01
Tentative Playwright support

### LINKED ISSUE #1029: Support for Playwright
**Problematic package: https://github.com/microsoft/playwright-python**
_Description:_  `playwright-python` is a browser-automation and testing framework maintained by Microsoft. It is a port from https://github.com/microsoft/playwright (nodejs) which finally allows us to use playwright in python. We can see this package is a highly improved version of [selenium](https://github.com/SeleniumHQ/selenium/tree/trunk/py) (of course internally it works completely different). It has a cleaner API, is much more stable and a lot of customizations are possible, that would not be possible with selenium.

After compiling and starting the excecutable via CMD, we get:
```
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\app.py", line 3, in <module>
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\playwright\sync_api\_context_manager.py", line 67, in __enter__
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\playwright\sync_api\_context_manager.py", line 48, in greenlet_main
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\asyncio\base_events.py", line 642, in run_until_complete
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\playwright\_impl\_connection.py", line 161, in run_as_sync
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\playwright\_impl\_connection.py", line 166, in run
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\playwright\_impl\_transport.py", line 60, in run
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\asyncio\subprocess.py", line 236, in create_subprocess_exec
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\asyncio\base_events.py", line 1661, in subprocess_exec
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\asyncio\windows_events.py", line 389, in _make_subprocess_transport
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\asyncio\base_subprocess.py", line 36, in __init__
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\asyncio\windows_events.py", line 885, in _start
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\asyncio\windows_utils.py", line 153, in __init__
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\subprocess.py", line 951, in __init__
  File "C:\Users\chwba\PYCHAR~1\NUITKA~3\APP~1.DIS\subprocess.py", line 1420, in _execute_child
FileNotFoundError: [WinError 2] The system cannot find the file specified
```
Compiled executable: https://nextcloud.mnwd.in/s/MzQYbnS4Q34wp5f
Compilation log: https://nextcloud.mnwd.in/s/BNawYSL4w9t9dNN

**Compiled with:**
(Tested with clang and msvc=14.2)
```
python -m nuitka --follow-imports^
                 --show-progress^
                 --show-scons^
                 --warn-unusual-code^
                 --assume-yes-for-downloads^
                 --warn-unusual-code^
                 --plugin-enable=pylint-warnings^
                 --standalone^
                 --clang^
                 app.py > compile.log
```

**Used requirements:**
```
playwright
```

**Code snippet to reproduce:**

Install playwright via `python -m pip install playwright` followed by `python -m playwright install`. - This triggers a download of the playwright browser executables, which are ultimately saved to: `C:\Users\%USERPROFILE%\AppData\Local\ms-playwright` as explained here: https://github.com/microsoft/playwright-python. 

I suppose nuitka can either not find the above mentioned playwright browser executables or the "driver executable" located in playwright/driver.

Also note, that playwright uses https://greenlet.readthedocs.io/en/latest/ (mentioning this, as I had noticed there is a plugin for gevent and eventlet).

```python
from playwright.sync_api import sync_playwright

with sync_playwright() as p:
	browser = p.chromium.launch(headless=False)
	page = browser.new_page()
	page.goto("http://playwright.dev")
	print(page.title())
	browser.close()
```

**Nuitka**
```
0.6.13
Python: 3.9.2 (tags/v3.9.2:1a79785, Feb 19 2021, 13:44:55) [MSC v.1928 64 bit (AMD64)]
Executable: C:\Users\chwba\PycharmProjects\nuitka_mp_pickling_error\venv\Scripts\python.exe
OS: Windows
Arch: x86_64

```
Installed via `python -m pip install nuitka`

KevinRodriguez777:
# What does this PR do?
 adds playwright support on windows
fixes #1029 
this is specific to windows as is pinned to the current version of the drivers that playwright uses and could be better, recursively looking for DLLs in `driver/package/.local-browsers/` will make this a cross platform solution regardless of what browser drivers the user has installed and as such potentially avoid bloat
# Note to users wanting to build an executable that imports playwright with Nuitka
even with this PR, you will need to have the following in your environment variables `PLAYWRIGHT_BROWSERS_PATH=0`
this tells playwright to install the browser drivers within your playwright package (`playwright/driver/.local-browsers/`) and will help Nuitka find the browsers much more easily
you will also need to run the `playwright install` command, however i suggest you install *only* the browsers you're using to avoid bloat (Bundling the browsers with the executables will generate bigger binaries. It is recommended to only bundle the browsers you use) by specifying it when installing e.g `playwright install chromium`

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,8 +1,8 @@
-nuitka (2.0~rc3+ds-1) unstable; urgency=medium
+nuitka (2.0~rc4+ds-1) unstable; urgency=medium
 
   * New upstream pre-release.
 
- -- Kay Hayen <kay.hayen@gmail.com>  Mon, 27 Nov 2023 04:45:24 +0100
+ -- Kay Hayen <kay.hayen@gmail.com>  Fri, 01 Dec 2023 19:23:16 +0100
 
 nuitka (1.9.3+ds-1) unstable; urgency=medium
 
--- a/nuitka/Version.py
+++ b/nuitka/Version.py
@@ -20,7 +20,7 @@
 """
 
 version_string = """\
-Nuitka V2.0rc3
+Nuitka V2.0rc4
 Copyright (C) 2023 Kay Hayen."""
 
 


kayhayen:
Environment variables can be set with pre and post load code for modules, consider e.g. this example that tries to limit the impact of the environment variable to timm, but that wouldn't be needed

```
- module-name: 'timm'
  implicit-imports:
    - pre-import-code:
        - |
          import os
          if os.environ.get('PYTORCH_JIT') is not None:
            os.environ['PYTORCH_JIT_OLD'] = os.environ.get('PYTORCH_JIT')
          os.environ['PYTORCH_JIT'] = "0"
      when: 'standalone'
    - post-import-code:
        - |
          import os
          if os.environ.get('PYTORCH_JIT_OLD') is None:
            del os.environ['PYTORCH_JIT']
          else:
            os.environ['PYTORCH_JIT'] = os.environ['PYTORCH_JIT_OLD']
            del os.environ['PYTORCH_JIT_OLD']
      when: 'standalone'
```

So you can add something for your variable like that easily. Benefit, malefit, if you do not consider presence of the variable, I think you should not consider it, so not even `.get is not None `check is needed.

KevinRodriguez777:
In this specific case for playwright, i think maybe it should be better to check for the environment variable, if it doesn't find it, raise an error with an link to this PR so that they know to set it and install (or reinstall) the drivers locally. thoughts?

KevinRodriguez777:
i want to catch the case where *no browsers* are installed.
>My question is, did the program work without the browsers installed locally then?
when no browsers are installed locally Playwright raises an error
image

KevinRodriguez777:
> If it didn't to begin with, you can expect to get the same error. It's not there to observe if you have done good things or not.
i hadn't thought of it that way, in which case just pushing the config yaml would be enough, and if people are having issues compiling the executable they can refer to the PR, alternatively i could ask playwright to include the same note for Nuitka that they do for [pyinstaller](https://playwright.dev/python/docs/library#pyinstaller) in regards for executables

kayhayen:
Sorry for the conflicts, in develop manual it explains how to rebase. 

Also, the playwright package, when it installs, you force it to do into the package dir, right? That's kind of unnatural. Would it normally be picked up from home directory? Because in that case, I believe a plugin is very much needed, so it can expose the browsers in there, and add options like `--include-playwright-browsers=all|chromium*|` etc. 

That's more along the lines of what the Qt plugins do for data files and DLLs, namely selective scanning from folders they know of, and probably even can ask playwright where it looks, so this is not super trivial. 

Coincidentally, that's another case of recursive and selective DLL inclusion there in Qt.

KevinRodriguez777:
> Also, the playwright package, when it installs, you force it to do into the package dir, right? 
yes, but that's also what Playwright tells people to do for [Pyinstaller](https://playwright.dev/python/docs/library#pyinstaller), why? i don't know, they don't say.

PR #2599, opened 2023-12-08
Yaml config additions

### LINKED ISSUE #2598: Specifying libusb backend and USB device not detected when compiling with nuitka
My application communicates with a specific device over USB. For Python to detect the device, I have to use https://github.com/pyocd/libusb-package and specify the backend. Below is my sample code: 

```python 
# Content of test_usb_comm.py
import libusb_package # https://github.com/pyocd/libusb-package
import usb.core
import usb.backend.libusb1
libusb1_backend = usb.backend.libusb1.get_backend(find_library=libusb_package.find_library)

def find_my_usb_devices(idVendor=0x0403, idProduct=0x90D9):
    device_list = []
    devices = usb.core.find(find_all=True, idVendor=0x0403, idProduct=0x90D9)
    for dev in devices:
        device_list.append(dev)
    return device_list

print(f'Listed of USB devices with the desired Vendor/Product ID:')
print(find_my_usb_devices(idVendor=0x0403, idProduct=0x90D9))
```

Below is the output when I run the python script: _test_usb_comm.py_.
```
Listed of USB devices with the desired Vendor/Product ID:
[<DEVICE ID 0403:90d9 on Bus 001 Address 008>]
```

**Problem:** After I compile my Python script, _test_usb_comm.py_, to an _test_usb_comm..exe_ file using Nuitka, my USB device is not being detected. I use this command to compile: `python -m nuitka --standalone .\test_usb_comm.py`. Below is the output I get when I run the .exe file. As you can see, my USB device is not detected. 

```
Listed of USB devices with the desired Vendor/Product ID:
[]
```

KevinRodriguez777:
#2589 accidentally closed, included it here.
added support for [playwright_stealth](https://github.com/AtuboDad/playwright_stealth)
added support for [libusb_package](https://github.com/pyocd/libusb-package), fixes #2598

### DIFF
--- a/nuitka/tools/watch/__main__.py
+++ b/nuitka/tools/watch/__main__.py
@@ -612,7 +612,9 @@ PR to create. Default not making a PR.""",
     for python_version in options.python_versions or reversed(
         getTestExecutionPythonVersions()
     ):
-        installed_pythons[python_version] = findPythons(python_version)
+        installed_pythons[python_version] = findPythons(
+            python_version, module_name="pipenv"
+        )
 
     nuitka_binary = os.path.abspath(os.path.expanduser(options.nuitka_binary))
     assert os.path.exists(nuitka_binary)
--- a/nuitka/utils/InstalledPythons.py
+++ b/nuitka/utils/InstalledPythons.py
@@ -181,32 +181,40 @@ def _getPythonInstallPathsWindows(python_version):
             yield candidate
 
 
-def findPythons(python_version):
+def findPythons(python_version, module_name=None, module_version=None):
     """Find all Python installations for a specific version."""
 
-    if python_version in _installed_pythons:
-        return _installed_pythons[python_version]
+    if python_version not in _installed_pythons:
+        result = OrderedSet()
 
-    result = OrderedSet()
+        if python_version == python_version_str:
+            result.add(
+                InstalledPython(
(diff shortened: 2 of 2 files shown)


kayhayen:
```
nuitka/plugins/standard/standard.nuitka-package.config.yml
  Warning: 2930:13 [comments] too few spaces before comment
  Warning: 2932:6 [comments] missing starting space in comment
```

So these warnings are not treated as errors, that is of course terrible, I didn't notice that yet. GitHub displays them when viewing the diffs. I will try and make a change for develop that turns them into proper errors.

KevinRodriguez777:
Playwright downloads Chromium, WebKit and Firefox browsers into the OS-specific cache folders:
```
%USERPROFILE%\AppData\Local\ms-playwright on Windows
~/Library/Caches/ms-playwright on MacOS
~/.cache/ms-playwright on Linux
```
this can be overriden via the `PLAYWRIGHT_BROWSERS_PATH` env variable
an solution to this would be for nuitka to first check the `PLAYWRIGHT_BROWSERS_PATH` env, if it's set to 0 (background can be found here: https://github.com/Nuitka/Nuitka/pull/2589#issue-2020704147) , then we can look inside driver/package/.local-browsers/, if it's other than 0, then whatever path it's provided to the env variable, else look into the paths i've referenced above.

Issue #2603, opened 2023-12-12
How can I change C compiler options, e.g. workaround default -fcf-protection option in the GCC compiler?


parkilwoo:
### Version
- 2.0rc5
- Commercial: None
- Python: 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0]
- Flavor: Debian Python
- Executable: /usr/bin/python
- OS: Linux
- Arch: x86_64
- Distribution: Ubuntu (based on Debian) 22.04.2
- Version C compiler: /usr/bin/gcc (gcc 11).

hello. I am using a .so file with Nuitka. I am using Thales solution for licensing the .so file, but I am not able to use Thales solution because the -fcf-protection option in GCC compiler is set to full, which is the default. Currently, I am using the SconsCompilerSettings.py file as a workaround and using the sed command to temporarily put in that part.
Is there a better way?

`sed -i 's/# Support for gcc and clang, restricting visibility as much as possible./env.Append(CCFLAGS=["-fcf-protection=none"])/' /usr/local/lib/python${PYTHON_VERSION}/dist-packages/nuitka/build/SconsCompilerSettings.py`

kayhayen:
https://nuitka.net/doc/user-manual.html#providing-extra-options-to-nuitka-c-compilation

kayhayen:
I saw your PR, very good. I welcome making these features directly accessible.

PR #2604, opened 2023-12-12
Added select -fcf-protection mode for the GCC compiler

parkilwoo:
# What does this PR do?
Enable integration with other third parties by selecting -fcf-protection for the GCC compiler, allowing you to choose the right trade-off between security and performance. 

#2603 

# PR Checklist

### DIFF
--- a/nuitka/OptionParsing.py
+++ b/nuitka/OptionParsing.py
@@ -953,12 +953,14 @@ and "auto" (when it's known to work). Defaults to "auto".""",
 )
 
 c_compiler_group.add_option(
-    "--unlock-fcf-protection",
-    action="store_true",
-    dest="unlock_fcf_protection",
-    default=False,
+    "--fcf-protection",
+    action="store",
+    dest="fcf_protection",
+    metavar="choice",
+    default="full",
+    choices=("full", "branch", "return", "none", "check"),
     help="""\
-Unlock fcf-protection""",
+Choice fcf-protection mode""",
 )
 
 
--- a/nuitka/OptionParsing.py
+++ b/nuitka/OptionParsing.py
@@ -952,6 +952,15 @@ Use static link library of Python. Allowed values are "yes", "no",
 and "auto" (when it's known to work). Defaults to "auto".""",
 )
 
+c_compiler_group.add_option(
+    "--unlock-fcf-protection",
+    action="store_true",
+    dest="unlock_fcf_protection",
+    default=False,
+    help="""\
+Unlock fcf-protection""",
+)
+
 
 del c_compiler_group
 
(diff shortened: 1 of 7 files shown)


nuitka/build/SconsCompilerSettings.py
```diff
@@ -628,6 +628,8 @@ def setupCCompiler(env, lto_mode, pgo_mode, job_count, onefile_compile):
             # compilation more memory hungry, but also faster.
             env.Append(CCFLAGS="-pipe")
 
+        # gcc complier fcf_protection option
+        env.Append(CCFLAGS=[f"-fcf-protection={env.fcf_protection}"])
```
kayhayen, line 632: We cannot use f-strings in Nuitka, but I can change that easily.

nuitka/OptionParsing.py
```diff
@@ -952,6 +952,17 @@
...
 
+c_compiler_group.add_option(
+    "--fcf-protection",
+    action="store",
+    dest="fcf_protection",
+    metavar="choice",
+    default="full",
+    choices=("full", "branch", "return", "none", "check"),
+    help="""\
+Choice fcf-protection mode""",
```
kayhayen, line 963: Enabling by default conflicts with performance goals of course, I would use `none` there. This most definitely needs a link to the GCC documentation for docs I guess.

nuitka/build/SconsCompilerSettings.py
```diff
@@ -628,6 +628,8 @@ def setupCCompiler(env, lto_mode, pgo_mode, job_count, onefile_compile):
             # compilation more memory hungry, but also faster.
             env.Append(CCFLAGS="-pipe")
 
+        # gcc complier fcf_protection option
```
kayhayen, line 631: Also unclear if clang has it, and if there is maybe a GCC version limitation, although I think I am about to drop very old gcc support, but it might have to be made conditional on that.

kayhayen (approved):

kayhayen:
I am going to merge it, but will then move it to the factory branch for the next pre-release, and make a few changes to it, so it is not active by default. I understand that you want to make it provide the option always, so it overrides. I would tend to have an "auto" value, and you would use "none". In that way, I avoid the issue of non-support in older gcc.

parkilwoo:
Thank you for your quick action.
I'm currently using the commercial version of Nuitka and it's working great.
Glad to be able to contribute to the product :)

kayhayen:
It's now on both factory and staging (commercial) branches. I chose to rename the option to `--cf-protection` and since the "-f" prefix is just how gcc passes these flags. Default is `auto` which indeed just doesn't provide the option. I also checked on Windows with MinGW64 and clang on Linux, all worked no problem. Thanks for your contribution.

PR #2605, opened 2023-12-13
Fixed Improper Method Call: Replaced `mktemp`

fazledyn-or:
# PR Checklist

## Summary
This PR fixes an unsafe method (mktemp) with a safer one (mkstemp). [Using an unsafe API to create temporary files (CWE-377)](https://cwe.mitre.org/data/definitions/377.html) can leave application and data vulnerable to attack.

## Description
While triaging your project, our bug fixing tool generated the following message(s)-

> In file: [win32.py](https://github.com/Nuitka/Nuitka/tree/develop/nuitka/build/inline_copy/lib/scons-2.3.2/SCons/Platform/win32.py#L133), there is a method that creates a temporary file using an unsafe API `mktemp`. The use of this method is discouraged in the [Python documentation](https://docs.python.org/3/library/tempfile.html#tempfile.mktemp). iCR suggested that a temporary file should be created using mkstemp which is a [safe API](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp). iCR replaced the usage of mktemp with mkstemp.

> In file: [win32.py](https://github.com/Nuitka/Nuitka/tree/develop/nuitka/build/inline_copy/lib/scons-3.1.2/SCons/Platform/win32.py#L187), there is a method that creates a temporary file using an unsafe API `mktemp`. The use of this method is discouraged in the [Python documentation](https://docs.python.org/3/library/tempfile.html#tempfile.mktemp). iCR suggested that a temporary file should be created using mkstemp which is a [safe API](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp). iCR replaced the usage of mktemp with mkstemp.

**Here Are Some Resources That Can Help You Understand**
- [POC - Improper Method Call - python - mktemp.mp4](https://drive.google.com/file/d/16YKuKgv7iItWcac2biRyi2Cm8upJ3HP_/view?usp=share_link)
- [MISC - Python - Taking a peek under the tempfile library.mp4](https://drive.google.com/file/d/1ns2_x_ZDvzFwj0eNNxyay-HUMclLasNT/view?usp=share_link)

It appers that the SCons library inside your repository is an outdated one. According to `setup.py` and `scons.py`, the outdated versions are used for POSIX based ssytems such as Unix and Linux. Since the original repo of SCons has updated their `mktemp` methods with `mkstemp`, it's better to adapt the changes to your project too.

I have updated the methods according to the [SCons source code](https://www.github.com/SCons/scons/commit/9368c9247952ff8a8b40da6cd15cd662cfcfdf0d).

## Previously Found & Fixed
Below is a list of open-source projects where this same bug was found and fixed-
- https://www.github.com/Azure/azure-linux-extensions/pull/1816
- https://www.github.com/celery/billiard/pull/394
- https://www.github.com/invesalius/invesalius3/pull/679
- https://www.github.com/spcl/dace/pull/1428
- https://www.github.com/scylladb/scylla-cluster-tests/pull/6795
- https://www.github.com/dimagi/commcare-hq/pull/33725
- https://www.github.com/SpikeInterface/spikeinterface/pull/2195
- https://www.github.com/nipy/nipy/pull/552
- https://www.github.com/imageio/imageio/pull/1055
- https://www.github.com/celery/kombu/pull/1828
- https://www.github.com/OCA/reporting-engine/pull/809

## CLA Requirements
*This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.*

All contributed commits are already automatically signed off.

> The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
>
> \- [Git Commit SignOff documentation](https://developercertificate.org/)

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/nuitka/build/inline_copy/lib/scons-2.3.2/SCons/Platform/win32.py
+++ b/nuitka/build/inline_copy/lib/scons-2.3.2/SCons/Platform/win32.py
@@ -130,8 +130,10 @@ def piped_spawn(sh, escape, cmd, args, env, stdout, stderr):
         return 127
     else:
         # one temporary file for stdout and stderr
-        tmpFileStdout = os.path.normpath(tempfile.mktemp())
-        tmpFileStderr = os.path.normpath(tempfile.mktemp())
+        tmpFileStdout, tmpFileStdoutName = tempfile.mkstemp(text=True)
+        os.close(tmpFileStdout)
+        tmpFileStderr, tmpFileStderrname = tempfile.mkstemp(text=True)
+        os.close(tmpFileStderr)
 
         # check if output is redirected
         stdoutRedirected = 0
@@ -147,9 +149,9 @@ def piped_spawn(sh, escape, cmd, args, env, stdout, stderr):
 
         # redirect output of non-redirected streams to our tempfiles
         if stdoutRedirected == 0:
-            args.append(">" + str(tmpFileStdout))
+            args.append(">" + str(tmpFileStdoutName))
         if stderrRedirected == 0:
-            args.append("2>" + str(tmpFileStderr))
+            args.append("2>" + str(tmpFileStderrname))
 
         # actually do the spawn
         try:
--- a/nuitka/build/inline_copy/lib/scons-3.1.2/SCons/Platform/win32.py
+++ b/nuitka/build/inline_copy/lib/scons-3.1.2/SCons/Platform/win32.py
@@ -183,8 +183,10 @@ def piped_spawn(sh, escape, cmd, args, env, stdout, stderr):
         return 127
     else:
         # one temporary file for stdout and stderr
-        tmpFileStdout = os.path.normpath(tempfile.mktemp())
-        tmpFileStderr = os.path.normpath(tempfile.mktemp())
+        tmpFileStdout, tmpFileStdoutName = tempfile.mkstemp(text=True)
+        os.close(tmpFileStdout)
+        tmpFileStderr, tmpFileStderrName = tempfile.mkstemp(text=True)
+        os.close(tmpFileStderr)
 
         # check if output is redirected
         stdoutRedirected = 0
@@ -199,9 +201,9 @@ def piped_spawn(sh, escape, cmd, args, env, stdout, stderr):
 
         # redirect output of non-redirected streams to our tempfiles
         if stdoutRedirected == 0:
-            args.append(">" + str(tmpFileStdout))
+            args.append(">" + str(tmpFileStdoutName))
         if stderrRedirected == 0:
-            args.append("2>" + str(tmpFileStderr))
+            args.append("2>" + str(tmpFileStderrName))
 
         # actually do the spawn
         try:
@@ -300,7 +302,7 @@ def get_system_root():
             except:
                 pass
 
-    # Ensure system root is a string and not unicode 
+    # Ensure system root is a string and not unicode
     # (This only matters for py27 were unicode in env passed to POpen fails)
     val = str(val)
     _system_root = val


nuitka/build/inline_copy/lib/scons-2.3.2/SCons/Platform/win32.py
```diff
@@ -130,8 +130,10 @@ def piped_spawn(sh, escape, cmd, args, env, stdout, stderr):
         return 127
     else:
         # one temporary file for stdout and stderr
-        tmpFileStdout = os.path.normpath(tempfile.mktemp())
-        tmpFileStderr = os.path.normpath(tempfile.mktemp())
+        tmpFileStdout, tmpFileStdoutName = tempfile.mkstemp(text=True)
```
kayhayen, line 133: I don't see how this is any safer, after closing, the race is back on. Note that this code is used with Python 2.6 only.

nuitka/build/inline_copy/lib/scons-3.1.2/SCons/Platform/win32.py
```diff
@@ -183,8 +183,10 @@ def piped_spawn(sh, escape, cmd, args, env, stdout, stderr):
         return 127
     else:
         # one temporary file for stdout and stderr
-        tmpFileStdout = os.path.normpath(tempfile.mktemp())
-        tmpFileStderr = os.path.normpath(tempfile.mktemp())
+        tmpFileStdout, tmpFileStdoutName = tempfile.mkstemp(text=True)
```
kayhayen, line 186: See my other comment no more safe.

nuitka/build/inline_copy/lib/scons-3.1.2/SCons/Platform/win32.py
```diff
@@ -199,9 +201,9 @@ def piped_spawn(sh, escape, cmd, args, env, stdout, stderr):
 
         # redirect output of non-redirected streams to our tempfiles
         if stdoutRedirected == 0:
-            args.append(">" + str(tmpFileStdout))
+            args.append(">" + str(tmpFileStdoutName))
```
kayhayen, line 204: The actual execution would have to use these filenames, and ideally not use files at all, but just pipes I guess.

kayhayen (changes requested):
This will on the surface hide the unsecure use of a temp filename by using a secure function to produce it, but otherwise adds no value.

The 4.3.0 version of scons apparently doesn't do it, but its use is limited to Windows and Python3 or higher. Maybe we should update Scons some time again and use it with all newer Pythons, but I didn't see a need so far. There was however a reason, it was not using the new Scons on non-Windows, but I don't really remember now, but Scons probably improved somewhat anyway.

However, these changes are not achieving anything good as they are, sorry to say.

nuitka/build/inline_copy/lib/scons-2.3.2/SCons/Platform/win32.py
fazledyn-or: The `tempfile.mkstemp` method is available and works for Python 2 too.

nuitka/build/inline_copy/lib/scons-3.1.2/SCons/Platform/win32.py
fazledyn-or: Let's say we have `tmpFileStdoutName = "/tmp/random"`

And previously, the `mktemp` method would just give you a random name for a temp file, won't create it. If we were to create the file ourselves, we could use `open` or a subprocess to let the shell write to it. Between these two steps, an attacker can (hypothetically) create a file with such name and modify the contents.

But, in case of `mkstemp`, the method creates the file for you, with a strict permission of 600. Meaning no other user except the user will be able to modify it. What if an attacker creates a file with that name at first? In that case, `mkstemp` will look if there's an existing file with that name. If there is, it'll create another random name to suggest and so on.

Which is why, even though the temp file is not being created by the Python program, we should use `mkstemp`- as long as it's being used as a "file". However, if `mktemp` is used as a random name generator (where no file operation are required), in those cases we can ignore such fixes.

nuitka/build/inline_copy/lib/scons-2.3.2/SCons/Platform/win32.py
kayhayen: Yes, that's true, I had checked myself too, also to see if `text` is a good mode, of that I wasn't sure, but since it's only using the filename, the mode doesn't really matter anyway. So not sure why it's given.

nuitka/build/inline_copy/lib/scons-3.1.2/SCons/Platform/win32.py
kayhayen: Ok, I get it, my bad, makes sense. Going to merge this one as well.

PR #2608, opened 2023-12-13
Fix bug: incorrectly access `pyproject.toml` `tool.nuitka` field

fynnsu:
# What does this PR do?

The `tool.nuitka` field from `pyproject.toml` gets loaded as a nested dict into the `toml_options` variable. 

Therefore `toml_options['tool.nuitka']` is empty and instead we need to access `toml_options['tool']['nuitka']`.

Despite using `v1.9.4` which includes d249f42645b16a0036d5730b24faffcf317fee07, nuitka didn't recognize the `tool.nuitka` field in my `pyproject.toml`.

# PR Checklist

Note: I didn't run the test suite for this small change, however tested on a local repo that this fixed my issue.

### DIFF
--- a/nuitka/distutils/DistutilCommands.py
+++ b/nuitka/distutils/DistutilCommands.py
@@ -323,7 +323,9 @@ class build(distutils.command.build.build):
                 for option, value in toml_options.get("nuitka", {}).items():
                     command.extend(self._parseOptionsEntry(option, value))
 
-                for option, value in toml_options.get("tool.nuitka", {}).items():
+                for option, value in (
+                    toml_options.get("tool", {}).get("nuitka", {}).items()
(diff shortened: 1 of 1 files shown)


nuitka/distutils/DistutilCommands.py
```diff
@@ -323,7 +323,9 @@ def _build(self, build_lib):
                 for option, value in toml_options.get("nuitka", {}).items():
                     command.extend(self._parseOptionsEntry(option, value))
 
-                for option, value in toml_options.get("tool.nuitka", {}).items():
+                for option, value in (
```
kayhayen, line 326: I see, it might not be there, thanks a lot for submitting this as a PR

Issue #2611, opened 2023-12-15
Nuitka doesn't handle using Python from compiled source checkout and cannot tell the location of Python.h


flozz:
Hello,

I am working on porting an application to macOS and for some reasons I had to build my own Python version (from a release tag of the official CPython Git repository). When I tried to compile my app, I had the following error:

```
Nuitka:INFO: Running C compilation via Scons.
Nuitka-Scons:INFO: Backend C compiler: clang (clang 15.0.0).
FATAL: Error, no 'Python.h' development headers can be found at '['/usr/local/include/python3.11', '/usr/local/develop/headers/python3.11']', dependency not satisfied!
```

I was able to fix my issue and to successfully build my app by editing the [`nuitka/build/Backends.scons`](https://github.com/Nuitka/Nuitka/blob/develop/nuitka/build/Backend.scons#L598-L606) file to add the path of the `Include/` folder of my compiled CPython.

* Is it possible/worthwhile to add a way to tel Nuitka where the `Python.h` file is located?
* What is the best way to achieve this? (CLI option? dedicated env var? parsing "common" env vars like CPATH/CCPATH/CFLAGS/...?)

I can work on a PR to implement whatever solution you think is the best to achieve this (if you think this feature is valuable for Nuitka).

## Nuitka / OS version

```
1.9.4
Commercial: None
Python: 3.11.7 (tags/v3.11.7:fa7a6f2303, Dec 14 2023, 15:44:57) [Clang 15.0.0 (clang-1500.1.0.2.5)]
Flavor: Unknown
Executable: /Users/fabien/myproject/build/__env__/bin/python
OS: Darwin
Arch: arm64
Version C compiler: /usr/bin/clang (clang 15.0.0).
```

## Steps to reproduce

I first installed XCode, HomeBrew and the following packages with `brew`:

```
brew install pkg-config openssl@3.0 xz gdbm tcl-tk
```

Then I built Python:

```bash
PATH="/opt/homebrew/bin/:$PATH"

cd ~/myproject/
mkdir -p build/

git clone https://github.com/python/cpython.git build/cpython.git
cd build/cpython.git
git checkout v3.11.7

GDBM_CFLAGS="-I$(brew --prefix gdbm)/include"
GDBM_LIBS="-L$(brew --prefix gdbm)/lib -lgdbm"
./configure \
    --with-openssl="$(brew --prefix openssl@3.0)"
make -j4

cd -
```

[...]

PR #2613, opened 2023-12-15
Handles virtualenv of not installed, self-compiled Python versions

flozz:
# What does this PR do?

This PR allows Nuitka to work with virtualenvs created from not-installed, self-compiled Python version on Unix systems (tested and works on macOS).

This PR is related to issue #2611.

# PR Checklist

---

NOTE: I successfully ran all the tests on Linux. On macOS they fail (either on my branch and on `develop`): 

```
+Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
```

### DIFF
--- a/nuitka/PythonVersions.py
+++ b/nuitka/PythonVersions.py
@@ -331,15 +331,9 @@ def getSystemPrefixPath():
 
             sys_prefix = getDirectoryRealPath(sys_prefix)
 
-        # Virtualenv of a not installed self-compiled Python version
-        if (
-            os.name != "nt"
-            and os.path.islink(sys.executable)
-            and os.path.isdir(
-                os.path.join(
-                    os.path.dirname(os.path.realpath(sys.executable)), "PCbuild"
-                )
-            )
+        # Self-compiled Python version in source tree
+        if os.path.isdir(
+            os.path.join(os.path.dirname(os.path.realpath(sys.executable)), "PCbuild")
         ):
             sys_prefix = os.path.dirname(os.path.realpath(sys.executable))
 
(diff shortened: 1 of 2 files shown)


kayhayen (changes requested):
I think the "islink" may not actually be necessary, if you just create the realpath and check it, you will also cover non-virtualenv usage as well, and the code will be not repeating itself, and I think, for Windows you need not exclude it at all, it will be good for detecting that one as well.

kayhayen:
One merged, I will retry with Windows and see how that helps. The detection of the PythonXY.LIB file kind of does similar things, but probably can be made into redundant code with a correct Python prefix. We might be able to drop the `os.name` exclusion, but I like it for now until it's also attempted.

kayhayen:
Looks like it can be 1.9 material to me right now, going to merge it to stable.

flozz:
I updated as requested :)

kayhayen (approved):

kayhayen:
Good job, I will actually benefit from this, since so far, when I (very rarely) did try to debug things by changing Python source code, I had to do an install step, that made things more clumsy, but now I expect it to work well.

I am however sceptical that your macOS build will work well for loading extension modules with Nuitka compiled binaries. I have seen problems with that, and mean to extract from somewhere what the proper way is. Ideally on macOS, and I do not know if it's possible, also it would be good to avoid "Python" as a DLL for macOS, for smaller deployments, and just generally because there is no real need for it, at least on Linux that works very well. I know on Windows DLL is required there, but I would assume that is not the case on macOS or is it.

Anyway, I will try it on Linux, and probably even merge to the next hotfix, as it seems very harmless, even if not super urgent stuff.

PR #2621, opened 2023-12-23
Added implicit imports for scipy

KevinRodriguez777:


kayhayen:
Now this conflicts of course, so the new checksum mechanism needs to die real soon now.

However, how do we handle this, how do we know that `scipy.cluster` can be used by `scipy`. Also, this looks like a lazy loader to me, notice the `def __getattr__` implementation. We should run time query `scipy.submodules` to know its submodules. That will also avoid the bloat concerns.

Something very similar to the code for pydantic should be applied:

```
        if module_name == "pydantic":
            # Pydantic has its own lazy loading, spell-checker: ignore pydantic
            if "def __getattr__(" in source_code:
                pydantic_info = self.queryRuntimeInformationSingle(
                    setup_codes="import pydantic",
                    value="pydantic._dynamic_imports",
                    info_name="pydantic_lazy_loader",
                )

                pydantic_lazy_loader_info = {}
                for key, value in pydantic_info.items():
                    # Older pydantic had only a string for the attribute.
                    if type(value) is tuple:
                        value = "".join(value).rstrip(".")

                    if value not in pydantic_lazy_loader_info:
                        pydantic_lazy_loader_info[value] = []
                    pydantic_lazy_loader_info[value].append(key)

                self._addLazyLoader(
                    module_name=module_name,
                    submodules=(),
                    submodule_attrs=pydantic_lazy_loader_info,
                )

```

Obviously `scipy` will be simpler there, since there will be less variants, and here it's the submodules field and no attrs to be given, but that's very much what is needed. Then we do `scipy` optimal per program as well.

PR #2622, opened 2023-12-24
added kiwisolver missing DLL

KevinRodriguez777:
fixes error due to missing DLL with kiwisolver


kayhayen (approved):

kayhayen:
What I really hate about the new checksums, is that on merge, they will very likely become wrong. I will really push to the per-item checksum instead, since that's just crazy source of merge conflicts and inconsistencies. But with a per module entry checksum, your merge could never cause a conflict, and we don't really care about the whole file being checked as much, only the items used.

kayhayen:
The checksums are a goner, hopefully they hit develop today

PR #2623, opened 2023-12-24
Fix --static-libpython=yes in pyenv

Eeems:
# What does this PR do?

It adds a new search path to get the static libpython when running in a pyenv.

I was messing around with running nuitka with python installed with pyenv for a github action (https://github.com/Eeems-Org/remarkable-nuitka-build-action) and it was failing to compile when `--static-libpython` was set to `yes` due to not finding the `libpython*.a` file.

# PR Checklist

### DIFF
--- a/nuitka/utils/StaticLibraries.py
+++ b/nuitka/utils/StaticLibraries.py
@@ -20,14 +20,12 @@
 """
 
 import os
-import subprocess
 
 from nuitka.containers.OrderedSets import OrderedSet
 from nuitka.PythonFlavors import (
     isAnacondaPython,
     isDebianPackagePython,
     isNuitkaPython,
-    isPyenvPython,
 )
 from nuitka.PythonVersions import (
     getPythonABI,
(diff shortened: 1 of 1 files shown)


Eeems:
I'm working around this for now by just linking the `libpython*.a` file to one of the locations it currently searches. I'm not really sure if this change deserves a test, it probably does, but I'm not really sure where to place it as I didn't see any existing test for the getSystemStaticLibPythonPath method after a quick scan.

nuitka/utils/StaticLibraries.py
```diff
@@ -189,6 +191,23 @@ def _getSystemStaticLibPythonPath():
                 # about that anyway.
                 pass
 
+        if isPyenvPython():
+            try:
+                machine = subprocess.check_output(["gcc", "-dumpmachine"])
```
kayhayen, line 196: I wonder if this lives somewhere in `sysconfig`. Usage or availability of GCC seems not a given. The value of `LIBPL`  could be that directory.

On my macOS it seems that way, and on Linux self compiled I got that:

```
>>> import sysconfig
>>> sysconfig.get_config_var("LIBPL")
'/opt/python3110/lib/python3.11/config-3.11-x86_64-linux-gnu'
```

And actually the static library lives there, so maybe it's not even specific.

kayhayen: So, I am thinking, you can make this not depend on the flavor at all, and I think it will immediately optimize self-compiled Python on Linux properly.

kayhayen (approved):
If you change it to the suggested folder, I will merge it gladly, this seems super nice.

kayhayen:
Just beware, need to test `sysconfig.get_config_var("LIBPL")` if it's not None, that is the case on at least Windows, but also others, and `os.path.join` is not going to like that value.

kayhayen:
I do not require such tests, it seems easy to test it for my self compiled Linux, and for what I know, pyenv is for the large part not all that difference, except maybe that having the static libpython is more rare. But with this, I wonder if it also automatically enables static libpython, probably not, right?

I am going to merge like that, just waiting for CI to see auto-format issues, etc. but to me that looks good. I also forgot that I kind of used this fore Debian before, I wonder where that `-pic` comes from though.

Eeems:
Sounds good! This was a pleasant experience 🙂

> But with this, I wonder if it also automatically enables static libpython, probably not, right?

I'm not sure, as I didn't look too closely at the code handling the auto detection. I would assume that if it's just checking to see if this returns a result, it would enable it.

PR #2632, opened 2023-12-29
Enhance Clang component not found exit info

shenjackyuanjie:
I change the `clang-cl.exe` not found error message
to make it more clearly

### DIFF
--- a/nuitka/build/SconsUtils.py
+++ b/nuitka/build/SconsUtils.py
@@ -434,7 +434,7 @@ def addClangClPathFromMSVC(env):
 
     if cl_exe is None:
         scons_logger.sysexit(
-            "Error, Visual Studio required for using ClangCL on Windows."
+            "Error, You need to install MSVC to use ClangCL on Windows."
         )
 
     clang_dir = os.path.join(cl_exe[: cl_exe.lower().rfind("msvc")], "Llvm")
@@ -465,7 +465,9 @@ def addClangClPathFromMSVC(env):
     if clangcl_path is None:
         # llvm project on windows also include clang-cl.exe
         scons_details_logger.sysexit(
-            "clang-cl.exe not found at '%s' or in %PATH%." % clang_dir
+            """clang-cl.exe not found in Visual Studio '%s' or in PATH.\n
+            "Install it from Visual Studio installer or from LLVM project release"""
+            % clang_dir
         )
 
     env["CC"] = "clang-cl"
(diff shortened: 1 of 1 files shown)


kayhayen:
I am not sure if MSVC is understood by more people than Visual Studio. It seems to me, while I use that term a lot, Python beginners not familiar with C development, will not know what it is. Probably having both mentioned is a good thing, maybe Nuitka should always say "MSVC (known as Visual Studio)" when it refers to that. These things could also very well have mnemonic pages, such that we can give more information, e.g. links to User Manual where it talks about that, direct download link and what not. The `mnemonic` argument would be good for that.

For the LLVM I am not sure if it gets added to PATH, probably the default installation path could be added too, when searching for it. However, I believe that pure LLVM is not working for Nuitka, since Windows SDK is not included there, and needs to be adapted for it to work, at least that was when I tried to use `clang.exe` maybe I never tried `clangcl.exe` where chances are that the Windows SDK needs no changes for it. However, without finding MSVC, I don't think it finds a Windows SDK in Scons, or does it, or does LLVM contain a windows.h file now?

shenjackyuanjie:
In my case, I install llvm manually to have the latest version, and add llvm into PATH
and install msvc with windows sdk by vs installer
which work fine with me

the point I want to change is that, you can install clang-cl.exe by VS or by llvm, so don't say "can't find Visual Studio", but tell the user what is lost, like `clang-cl.exe`

kayhayen:
I am not sure, that clang can work without MSVC installed, however, so it may well be what's missing.

kayhayen:
Actually in Scons, nothing is going to setup the environment in which `clangcl.exe` operates, what flags is uses, without first finding a cl.exe and having that work, and then the path is replaced. So it's misleading to say `or`. It seems you found out that instead of from MSVC you can install clangcl.exe from LLVM, but there is not a lot of benefit from that, except that we don't know if it's working as well as the package provided by Microsoft for that purpose.

Unless somebody is a huge LLVM fan that needs the very latest release to be used, I do not see the value in using it. If of course it worked standalone, sure, but I see no way of that happening unless we modify Scons very heavily, and I believe there is no Windows SDK in Clang distributions, so we would have to start and detect that, and, and... for what precisely?

shenjackyuanjie:
> I am not sure, that clang can work without MSVC installed, however, so it may well be what's missing.

yes, clang can't work without mingw or msvc, but the point is, clang-cl and cl.exe are different, and they can be install independently, so it is possible that a user install msvc, but didn't install clang, so it is useful to tell the user what is missing.
will this be more logical to you?

kayhayen:
Your current text has "or LLVM" which is absolutely untrue for all I know. The mention of the installer is better, but the "Clang component" thing is not mentioned anymore. The installer installs MSVC, you need to configure it as a component there.

kayhayen:
Is there anything forthcoming from this, the text can be enhanced, it just cannot become wrong.

PR #2700, opened 2024-02-08
PonyORM implicit dependencies

### LINKED ISSUE #44: PonyORM incompatible with Nuitka
I've found a runtime failure when trying to use a module compiled with nuitka which uses PonyORM for managing a sqlite database.
[PonyORM - Python ORM with beautiful query syntax](https://ponyorm.com/)

```
  File "C:\APP\app\resultsdb\__init__.py", line 109, in load
  File "<string>", line 2, in select
  File "C:\APP\app_packages\pony\utils\utils.py", line 58, in cut_traceback
    return func(*args, **kwargs)
  File "C:\APP\app_packages\pony\orm\core.py", line 5160, in select
    return make_query(args, frame_depth=3)
  File "C:\APP\app_packages\pony\orm\core.py", line 5147, in make_query
    tree, external_names, cells = decompile(gen)
  File "C:\APP\app_packages\pony\orm\decompiling.py", line 28, in decompile
    else: throw(TypeError)
  File "C:\APP\app_packages\pony\utils\utils.py", line 98, in throw
    raise exc
TypeError
```

Some initial diagnostics shows that it's really not likely to work, probably ever.... pony does some crazy introspection on the code to be able to use really convenient python syntax to generate sql queries.

Here's a really good description of how it work: https://stackoverflow.com/questions/16115713/how-pony-orm-does-its-tricks

And specifically:
https://github.com/ponyorm/pony/blob/orm/pony/orm/decompiling.py#L17
```
def decompile(x):
    cells = {}
    t = type(x)
    if t is types.CodeType: codeobject = x
    elif t is types.GeneratorType: codeobject = x.gi_frame.f_code
```
In my case, when it's running as a compiled module, "t" above is ` <class 'compiled_generator'>`
Even if I modify pony to accept this as a `types.GeneratorType` the object does not have the `x.gi_frame` attribute. 

I'm not particularly asking for this to be fixed, I've moved away from using PonyORM already (not even using a database anymore, switched to yaml with my `structured_config` module to manage)

I just wanted to document my findings in case others run into the same issue.

KevinRodriguez777:
fixes #44 
works against
```
from pony.orm.core import *

db = Database()


class Numbers(db.Entity):
    _table_ = "Numbers"
    id = PrimaryKey(int, auto=True)
    int8 = Required(int, size=8)  # TINYINT
    int16 = Required(int, size=16)  # SMALLINT
    int24 = Required(int, size=24)  # MEDIUMINT
    int32 = Required(int, size=32)  # INTEGER
    int64 = Required(int, size=64)  # BIGINT
    uint8 = Required(int, size=8, unsigned=True)  # TINYINT UNSIGNED
    uint16 = Required(int, size=16, unsigned=True)  # SMALLINT UNSIGNED
    uint24 = Required(int, size=24, unsigned=True)  # MEDIUMINT UNSIGNED
    uint32 = Required(int, size=32, unsigned=True)  # INTEGER UNSIGNED
    # uint64 = Required(int, size=64, unsigned=True)  # BIGINT UNSIGNED, supported by MySQL and Oracle


sql_debug(True)  # Output all SQL queries to stdout

db.bind("sqlite", "test_numbers.sqlite", create_db=True) # TESTED BY KRRT
# db.bind("mysql", host="localhost", user="pony", passwd="pony", db="test_numbers") # TESTED BY KRRT
# db.bind("postgres", user="pony", password="pony", host="localhost", database="test_numbers") #UNTESTED
# db.bind("oracle", "test_numbers/pony@localhost") # UNTESTED

db.drop_table("Numbers", if_exists=True, with_all_data=True)
db.generate_mapping(create_tables=True)


@db_session
def populate_database():
    lo = Numbers(int8=-128, int16=-32768, int24=-8388608, int32=-2147483648, int64=-9223372036854775808, uint8=0, uint16=0, uint24=0, uint32=0)  # , uint64=0)
    hi = Numbers(int8=127, int16=32767, int24=8388607, int32=2147483647, int64=9223372036854775807, uint8=255, uint16=65535, uint24=16777215, uint32=4294967295)
    # uint64=18446744073709551615)
    commit()


@db_session
def test_data():
    for n in Numbers.select():
        print(n.id, n.int8, n.int16, n.int24, n.int32, n.int64, n.uint8, n.uint16, n.uint24, n.uint32)  # , n.uint64)


if __name__ == "__main__":
    populate_database()
    test_data()

```

### DIFF
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -3103,14 +3103,10 @@
         'polyfactory.pytest_plugin': 'ignore'
       when: 'not use_pytest'
 
-- module-name: 'pony' # checksum: 2a40ab64
+- module-name: 'pony' # checksum: 3776d9aa
   implicit-imports:
     - depends:
-        - 'pony.orm.dbproviders'
-        - 'pony.orm.dbproviders.sqlite'
-        - 'pony.orm.dbproviders.mysql'
-        - 'pony.orm.dbproviders.postgres'
-        - 'pony.orm.dbproviders.oracle'
+        - 'pony.orm.dbproviders.*'
 
 - module-name: 'pooch' # checksum: e07f91ff
   anti-bloat:
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -3103,6 +3103,15 @@
         'polyfactory.pytest_plugin': 'ignore'
       when: 'not use_pytest'
 
+- module-name: 'pony' # checksum: 2a40ab64
+  implicit-imports:
+    - depends:
+        - 'pony.orm.dbproviders'
+        - 'pony.orm.dbproviders.sqlite'
+        - 'pony.orm.dbproviders.mysql'
+        - 'pony.orm.dbproviders.postgres'
+        - 'pony.orm.dbproviders.oracle'
+
 - module-name: 'pooch' # checksum: e07f91ff
   anti-bloat:
     - description: 'remove pytest reference'


nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -3103,6 +3103,15 @@
         'polyfactory.pytest_plugin': 'ignore'
       when: 'not use_pytest'
 
+- module-name: 'pony' # checksum: 2a40ab64
+  implicit-imports:
+    - depends:
+        - 'pony.orm.dbproviders'
```
kayhayen, line 3109: Maybe use `.*` notation for catch them all?

kayhayen (changes requested):

nuitka/plugins/standard/standard.nuitka-package.config.yml
KevinRodriguez777: done.

kayhayen (approved):

PR #2702, opened 2024-02-08
h2o support

### LINKED ISSUE #654: compiling h2o project with Nuitka
Hi,

i open this issue because i have problems to compile an H2O project with Nuitka.
https://www.h2o.ai/
H2O is a framework to do machine learning and artificial intelligence model. It works in a simply mode, after installing H2O in python env, we need to write 2 rows to use it:

```
import h2o
h2o.init(ip="localhost", port=54321, max_mem_size="8G", nthreads=-1)
```

With the second line, a cluster H2O is started, or if already started, it connect to him. The cluster (1 node in local pc) is the central element that permit to prepare data and build models. how? there's a jar file that is used (h2o.jar) that is searched in a python path.

Now, i tried in several ways to build a test project with this 2 lines. Here the code:

```
import h2o

def talk(message):
    return "Talk " + message

def main():
    print( talk("Hello World"))

if __name__ == "__main__":
    h2o.init(ip="localhost", port=54321, max_mem_size="8G", nthreads=-1)
    main()
```

When i finish to compile, and tried to start my executable, the h2o.jar file that is used to start the cluster, is not included. I tried to manually create the folder and moved this file in the path when it's searched, but then the problem is about the H2O version of jar file that mismatch. While the H2O.jar is version 3.28.0.3 (in python env first build), in the compiled project, the version that is registered is 0.0.local.
Someone had experience with this problem?
I tried command line like --include-package=h2o
-- follow-imports
--follow-imports --recurse-all
-- include plugin-directory

Follow now my step that i did, to reproduce the problem:

python version:
Python 3.6.9

I created a virtualenv:
virtualenv -p /usr/bin/python3 nuitka-venv

pip install h2o
pip install pandas
pip install -U nuitka

Now, opening a python shell, this 2 lines works

```
import h2o
h2o.init(ip="localhost", port=54321, max_mem_size="8G", nthreads=-1)
```

but when inserted in hello.py and compiled, something of h2o is not included in the compiled version.
Other info:
python -m nuitka --version
0.6.7
Python 3.6.9
Executable /home/valerio/Scrivania/test/nuitka-venv/bin/python
OS: Linux
Arch: x86_64

My ubuntu is 18.04.4 LTS

any kind of help is appreciated, thanks!

KevinRodriguez777:
fixes #654 
the `*.txt` catches `version.txt` and `buildinfo.txt` both of which h2o breaks without.
the `*.csv` catches two csv that h2o likes having.

### DIFF
--- a/nuitka/tools/testing/Common.py
+++ b/nuitka/tools/testing/Common.py
@@ -1186,9 +1186,7 @@ def getTestingCacheDir():
 def getTestingCPythonOutputsCacheDir():
     cache_dir = getCacheDir("cpython_outputs")
 
-    result = os.path.join(
-        cache_dir, os.getenv("NUITKA_TEST_SUITE", "")
-    )
+    result = os.path.join(cache_dir, os.getenv("NUITKA_TEST_SUITE", ""))
 
     makePath(result)
     return result


nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -1426,6 +1426,22 @@
...
+- module-name: 'h2o' # checksum: a246b3c9
+  data-files:
+    dirs:
+      - 'h2o_data/'
+    patterns:
+      - '*.txt'
+      - '*.csv'
+  dlls:
+    - from_filenames:
+        relative_path: 'backend/bin/'
```
kayhayen, line 1438: The trailing slash in these paths should not be, I need to figure out how to make auto-format remove them, or the schema to mark them as an error, because they all are supposed to be directory names.

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -1426,6 +1426,22 @@
...
+        prefixes:
+          - 'h2o'
+        suffixes:
+          - 'jar'
+        executable: 'yes'
```
kayhayen, line 1443: Are jar files really executables, or should this not be data files too?

kayhayen (changes requested):
Not sure about the jar file yet, if these really are needing an x-bit that would be something new to me. Otherwise I feel they are going to have issues with DLL dependency scans, since they are not actually binaries. dependency walker, ldd, otool, all might hate them.

kayhayen:
This one didn't make the cut for 2.0.2

nuitka/plugins/standard/standard.nuitka-package.config.yml
KevinRodriguez777: do you consider `.py` files executables? if you you'd execute one via python helloworld.py right?
for `jar` files, you'd do the same `java -jar helloworld.jar`
there's also the fact that jar files can have an manifest file containing the main entry point and those sort of jars are considered executable, i think h2o executes a subprocess to run this file, so it should be considered an executable.

nuitka/plugins/standard/standard.nuitka-package.config.yml
KevinRodriguez777: resolved.

nuitka/plugins/standard/standard.nuitka-package.config.yml
kayhayen: A jar file is not an executable by itself in my book, just because it is input to the `java` executable, to Nuitka is a data file totally, since we are not going to do anything but copying it and maybe embedding it. But I was wondering if the "executable" flag means that e.g. on Linux, an x-bit would be required, like if `helloworld.jar` would be executable somehow, and since these are zip files, I imagine they can `actually` have a runner in front, in which case, they actually are executables. That runner would have to be made portable, signed, etc potentially.

I think Python files are too messy as an example to discuss. I believe they are never data files, but currently we do not have the deployment options that would compile them to Python code for inclusion. That really must change, since providing editable plugins is a valid use case, but it doesn't get done so far.

kayhayen:
I think you need to reopen this one or create a new PR?

PR #2739, opened 2024-03-03
add metada for toga on windows & linux

KevinRodriguez777:


kayhayen (changes requested):
It seems macOS is missing, but I can add it, right?

KevinRodriguez777:
> It seems macOS is missing, but I can add it, right?

i don't have a MacOS machine so i can't help there

kayhayen:
I am merging this, not sure when I will get to doing it for macOS, but your Nuitka-Watch PR will drive that naturally once I merge that one as well. But I might take a moment, so I can wait for 2.1.1 with that.

Issue #2746, opened 2024-03-05
Unable to obtain driver from chrome using Selenium Manager when onefile and standalone


Diegocndd:
Hello! I didn't use chromedriver in my project. With the new version of Selenium (4), it is no longer necessary to specify a chromedriver. However, this chromedriver management is not working.

My code is as simples as:

``` 
from selenium import webdriver

chrome_options = webdriver.ChromeOptions()
driver = webdriver.Chrome(options=chrome_options)
driver.get('https://www.google.com/')
```

The code runs on the system fine and when i don't use onefile.

Compiled it with: python -m nuitka --follow-imports  --standalone --onefile --include-package=selenium  main.py

When executing main.exe, I receive the following error:

```
Traceback (most recent call last):
File "C:\Users\diego\AppData\Local\Temp\OND5FF~1\selenium\webdriver\common\driver_finder.py", line 38, in get_path
File "C:\Users\diego\AppData\Local\Temp\OND5FF~1\selenium\webdriver\common\selenium_manager.py", line 89, in driver_lo
cation
File "C:\Users\diego\AppData\Local\Temp\OND5FF~1\selenium\webdriver\common\selenium_manager.py", line 73, in get_binar

selenium. common. exceptions. WebDriverException: Message: Unable to obtain working Selenium Manager binary; C:\Users\diego
\AppData\Local\Temp\OND5FF~1\selenium\webdriver\common\windows\selenium-manager. exe

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
File "C:\Users\diego\AppData\Local\Temp\OND5FF~1\p.py", line 8, in <module>
driver = webdriver. Chrome(options=chrome_options)
File "C:\Users\diego\AppData\Local\Temp\OND5FF~1\selenium\webdriver\chrome\webdriver.py", line 45, in __ init
File "C:\Users\diego\AppData\Local\Temp\OND5FF~1\selenium\webdriver\chromium\webdriver.py", line 49, in __ init_
File "C:\Users\diego\AppData\Local\Temp\OND5FF~1\selenium\webdriver\common\driver_finder.py", line 41, in get_path
selenium. common. exceptions. NoSuchDriverException: Message: Unable to obtain driver for chrome using Selenium Manager. ; F
or documentation on this error, please visit: https://www.selenium.dev/documentation/webdriver/troubleshooting/errors/dr
iver_location
```

# Nuitka

Version 2.1
Non commercial

# pip freeze

```
[...]

PR #2758, opened 2024-03-10
selenium support

KevinRodriguez777:


KevinRodriguez777:
fixes https://github.com/Nuitka/Nuitka/issues/2746

.gitignore
```diff
@@ -67,8 +67,6 @@ doxygen-*
 .coverage*
 *.tmp
 *.dat
-.vscode/ipch
-.vscode/tags
-.vscode/.ropeproject
 Nuitka-*
 nuitka/tools/podman/containers/requirements-devel.txt
+.vscode/
```
kayhayen, line 72: I want to know about files there, that are to be ignored, so as to not missing things not to ignore, this counteracts it, so I won't merge this part.

kayhayen (approved):

kayhayen:
I am merging this, but I will remove the gitignore immediately, and put it on factory and hotfix branch, so I can have it for 2.1.2 release.

.gitignore
KevinRodriguez777: so you'd rather gitignore specific files?

kayhayen: Yes, already doing that, I think you removed some, those ones are vscode state files, but every now and then, an extension creates a config file, and that I would hate to accidentally ignore. So whatever brought you to ignoring that, it's worth considering if that file is for git to contain or not on a merits base.

furjac:
Is selenium working now

Issue #792, opened 2020-07-21
Option "--recompile-c-only" misleading.


belonesox:
While reading from option description

 "--recompile-c-only" —  "… Takes existing files and simply compile them as C again. …" 

someone (as me) may thought that is The Option that opposite to "--generate-c-only" that skip VERY LONG 
```Optimization.optimize(main_module.getOutputFilename())``` and only call Scons with clcache to recompile already generated C files (for hacking, experiments with compiler/linker options, etc).

But unfortunately, it is not true, Nuika still takes hours to "Optimization.optimize"
and there is no other option "--please-just-make-scons-and-recompile-already-generated-files-and-nothing-more".

Please, let  "--recompile-c-only" will only recompile C code, or.... if I missed the goal of the option, just make another option for this purpose.

kayhayen:
Unfortunately, for scons to work, Nuitka needs the information about the modules, and this would have to come from compilation or a cache. The later is the actual solution in my mind, but it doesn't exist, but for this I see this issue as a duplicate.

What this option does is to skip generating the C files, which is only useful for manual editing while debugging. Not a user facing option at all.

belonesox:
So, can we discuss here or on another issue other possibility, like option `--stop-on-generating-scons-file-and-shell-script-for-compilation`?

This option would stop Nuitka, generate shell file with Scons command for compilation and leaves it for researcher (experiments with compilation flags, etc)?

kayhayen:
Nuitka puts things in environment variables, on the command line when it calls things, in include files, etc. and is all dirty about this folder contents. This could get cleaned up, but it would also increase the maintenance burden for what exactly. I would much rather see time invested in proper caching, such that it would be close to that in effect.

belonesox:
All these stuff (environment variables, etc) also can be freezed on some  `set-up-envs-and-build-generated-cpp-sources-using-scons.sh` file.

I like to experiment with cpp compile phase (different gcc/ld options, using SAST tools for CPP, etc), without regenerating sources fron Nuitka. May be I would write dirty-hacking-monkey-patching plugin for this, but official option of course would be much better.

kayhayen:
Adding plugin interfaces of general usablity would be great.

[...]

PR #2770, opened 2024-03-16
Fix obvious typo, that breaks idempotency of scons build process — af…

belonesox:
# What does this PR do?

Fix obvious typo, that breaks idempotency of scons build process — after renaming `.c` → `.cpp`, in next Scons call (if we have to debug scons process, for example) it wrongly skipped `.cpp` files because of this typo.

Yes, I wish to make scons compilation reproduceable and debuggable without source code generation phase https://github.com/Nuitka/Nuitka/issues/792#issuecomment-1601361367

# PR Checklist

- [x] All tests still pass. Check the Developer Manual about `Running the Tests`. (Actually, some test on python 2 failed, but they failed without this patch).
- [ ] Ideally new features or fixed regressions ought to be covered via new tests. → I cannot generation of debug script for for Scons build without idempotency of build process.
- [ ] Ideally new or changed features have documentation updates. → Not needed

### DIFF
--- a/nuitka/build/SconsUtils.py
+++ b/nuitka/build/SconsUtils.py
@@ -613,7 +613,7 @@ def scanSourceDir(env, dirname, plugins):
 
         # Only C files are of interest here.
         if not hasFilenameExtension(
-            filename_base, (".c", "cpp")
+            filename_base, (".c", ".cpp")
         ) or not filename_base.startswith(("module.", "__", "plugin.")):
             continue
 


kayhayen:
Big words like idempotency used. :-)

Thanks a lot, this will go to the next hotfix directly.

kayhayen:
This actually is kind of relevant to us for zig as well, since we cannot use C11 mode there too. Also there is the splash screen stuff for Windows, that is C++ to begin with.

Issue #2407, opened 2023-08-25
Compatibility with bitsandbytes


hugoabonizio:
I'm having some trouble trying to compile a code that uses the [transformers](https://github.com/huggingface/transformers) library with [bitsandbytes](https://github.com/TimDettmers/bitsandbytes) for quantization. Without the flag `load_in_4bit=True`, which uses `bitsandbytes`, everything works fine. But when I enable it I get the error described bellow.

Environment setup:
```
python3 -m venv env1
source env1/bin/activate
python -m pip install nuitka transformers==4.30 bitsandbytes accelerate scipy
python -m pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu118
```

Version:
```
1.7.10
Commercial: None
Python: 3.8.10 (default, May 26 2023, 14:05:08) 
Flavor: Debian Python
Executable: /hugo/maritaca/testes/nuitka/test1/env1/bin/python
OS: Linux
Arch: x86_64
Distribution: Ubuntu (based on Debian) 20.04.6
Version C compiler: /usr/bin/gcc (gcc).
```

Installed packages:
```
accelerate==0.22.0
bitsandbytes==0.41.1
certifi==2023.7.22
charset-normalizer==3.2.0
cmake==3.25.0
filelock==3.12.2
fsspec==2023.6.0
huggingface-hub==0.16.4
idna==3.4
Jinja2==3.1.2
lit==15.0.7
MarkupSafe==2.1.2
mpmath==1.2.1
networkx==3.0
Nuitka==1.7.10
numpy==1.24.4
ordered-set==4.1.0
packaging==23.1
Pillow==9.3.0
psutil==5.9.5
PyYAML==6.0.1
regex==2023.8.8
requests==2.31.0
safetensors==0.3.2
scipy==1.10.1
sympy==1.11.1
tokenizers==0.13.3
torch==2.0.1+cu118
torchaudio==2.0.2+cu118
torchvision==0.15.2+cu118
tqdm==4.66.1
transformers==4.30.0
triton==2.0.0
typing-extensions==4.7.1
urllib3==2.0.4
zstandard==0.21.0
```

Script:
```python
from transformers import AutoModelForCausalLM, AutoTokenizer

load_in_4bit = True

model = AutoModelForCausalLM.from_pretrained(
    'facebook/opt-125m',
    load_in_4bit=load_in_4bit,
)
if not load_in_4bit:
	model.to('cuda:0')
[...]

PR #2798, opened 2024-03-31
Update TransformersPlugin.py

KRRT7:
adds missing models


KevinRodriguez777:
closes #2407

kayhayen:
Looks good to me, apparently so sorted the items, makes sense, I think this might have been file system sorting from my WSL or so.

kayhayen:
My only concern is that bitsandbytes changed their wheel packaging for a version, if possible, I would love to figure out the version that did it. Let me try, it's easy enough with --list-package-dlls and some bisecting.

kayhayen:
I had manually merged this one and then forgotten to close it.

PR #2802, opened 2024-04-03
⚡️ Speed up `_encodePythonStringToC()` by 80% in `nuitka/utils/CStrings.py`

aphexcx:
### 📄 `_encodePythonStringToC()` in `nuitka/utils/CStrings.py`

📈 Performance improved by **`80%`** (**`0.80x` faster**)

⏱️ Runtime went down from **`1289.80μs`** to **`716.90μs`**
### Explanation and details

<details>
<summary>(click to show)</summary>

Your current code is processing the string character by character, causing a lot of operations and memory allocations, which makes the code inefficient. We could eliminate some of these operations by handling the string as a whole.

Some Improvements.

- String concatenations are expensive operations. We can use a list instead and join it at the end.
- We can use a dictionary to map those special characters with its corresponding encoded values.
- `str is bytes` will always be false, so this check is not really beneficial in your case.

Here's your refactored code.

</details>

### Correctness verification

The new optimized code was tested for correctness. The results are listed below.
#### ✅ 31 Passed − 🌀 Generated Regression Tests
<details>
<summary>(click to show generated tests)</summary>

```python
# imports
import pytest  # used for our unit tests

# function to test is already provided, so we do not need to redefine it here.

# unit tests

def test_normal_ascii_characters():
    assert _encodePythonStringToC(b'Hello, World!') == '"Hello, World!"'
    assert _encodePythonStringToC(b'Python 3.8') == '"Python 3.8"'

def test_special_characters_escaped():
    assert _encodePythonStringToC(b'\tTabbed') == r'"\011Tabbed"'
    assert _encodePythonStringToC(b'\rCarriage\rReturn') == r'"\015Carriage\015Return"'
    assert _encodePythonStringToC(b'\nNew\nLine') == r'"\012New\012Line"'
    assert _encodePythonStringToC(b'Double"Quote') == r'"Double\042Quote"'
    assert _encodePythonStringToC(b'Back\\slash') == r'"Back\134slash"'
    assert _encodePythonStringToC(b'Question?Mark') == r'"Question\077Mark"'

def test_non_printable_ascii_characters():
    assert _encodePythonStringToC(b'\x00\x01\x02\x03') == r'"\000\001\002\003"'
    assert _encodePythonStringToC(b'\x7f') == r'"\177"'

def test_extended_ascii_characters():
    # Extended ASCII characters are not valid in a C string, so we expect an AssertionError
    with pytest.raises(AssertionError):
        _encodePythonStringToC(b'\x80\x9f\xa0\xff')

def test_binary_data_with_null_bytes():
    assert _encodePythonStringToC(b'\x00\x00\x00ABC\x00\x00') == r'"\000\000\000ABC\000\000"'

def test_octal_escape_sequence_followed_by_ascii_digits():
    assert _encodePythonStringToC(b'\x03123') == r'"\003""123"'
    assert _encodePythonStringToC(b'\x04123456789') == r'"\004""123456789"'

def test_empty_string():
    assert _encodePythonStringToC(b'') == '""'

def test_long_strings():
    assert _encodePythonStringToC(b'A' * 1000) == '"' + 'A' * 1000 + '"'
    assert _encodePythonStringToC(b'\n' * 1000) == r'"\012' + r'\012' * 999 + '"'

def test_unicode_characters_outside_ascii_range():
    # Unicode characters are not valid in a C string, so we expect an AssertionError
    with pytest.raises(AssertionError):
        _encodePythonStringToC('你好'.encode('utf-8'))
    with pytest.raises(AssertionError):
        _encodePythonStringToC('café'.encode('utf-8'))

# edge cases

def test_consecutive_special_characters_escaped():
    assert _encodePythonStringToC(b'\\\t\\') == r'"\134\011\134"'
    assert _encodePythonStringToC(b'\r\n\r\n') == r'"\015\012\015\012"'

def test_octal_escape_sequence_near_boundary_of_printable_ascii():
    assert _encodePythonStringToC(b'\x7e0') == r'"\176""0"'
    assert _encodePythonStringToC(b'\x20123') == r'"\040""123"'

def test_escaped_characters_followed_by_similar_unescaped_characters():
    assert _encodePythonStringToC(b'\\\\"') == r'"\134\134\042"'
    assert _encodePythonStringToC(b'\\\\"\\\\"') == r'"\134\134\042\134\134"'

def test_strings_end_with_octal_escape_sequence():
    assert _encodePythonStringToC(b'Ending with escape\x03') == r'"Ending with escape\003"'
    assert _encodePythonStringToC(b'Another\x07') == r'"Another\007"'

def test_octal_escape_sequence_followed_by_non_octal_digit():
    assert _encodePythonStringToC(b'\x038') == r'"\003""8"'
    assert _encodePythonStringToC(b'\x039') == r'"\003""9"'

def test_input_with_only_escape_characters():
    assert _encodePythonStringToC(b'\\\\') == r'"\134\134"'
    assert _encodePythonStringToC(b'\t\t\t\t') == r'"\011\011\011\011"'

def test_input_with_mix_of_octal_escape_sequences_and_printable_characters():
    assert _encodePythonStringToC(b'\x01Printable\x7fCharacters') == r'"\001Printable\177Characters"'

def test_input_with_escaped_characters_not_special_in_c_strings():
    assert _encodePythonStringToC(b'\x1b') == r'"\033"'
    assert _encodePythonStringToC(b'\x7e\x7f') == r'"\176\177"'
```
</details>

This optimization was discovered by [Codeflash AI](https://codeflash.ai) ⚡️

# PR Checklist

### DIFF
--- a/nuitka/utils/CStrings.py
+++ b/nuitka/utils/CStrings.py
@@ -19,41 +19,32 @@ def _identifierEncode(c):
 
 
 codecs.register_error("c_identifier", _identifierEncode)
-
-
 def _encodePythonStringToC(value):
-    """Encode a string, so that it gives a C string literal.
-
-    This doesn't handle limits.
-    """
-    assert type(value) is bytes, type(value)
-
-    result = ""
+    """Encode a string, so that it gives a C string literal"""
+    assert isinstance(value, bytes), type(value)
+    
+    special_chars = {92: r"\134", 9: r"\011", 13: r"\015", 10: r"\012", 63: r"\077", 34: r"\042"}
+    result = []
     octal = False
-
     for c in value:
-        if str is bytes:
-            cv = ord(c)
-        else:
-            cv = c
+        cv = c
 
-        if c in b'\\\t\r\n"?':
-            result += r"\%03o" % cv
-
+        if c in special_chars.keys():
+            result.append(special_chars[c])
             octal = True
+            
         elif 32 <= cv <= 127:
             if octal and c in b"0123456789":
-                result += '" "'
+                result.append(r'" "')
 
-            result += chr(cv)
-
+            result.append(chr(cv))
             octal = False
         else:
-            result += r"\%o" % cv
+            result.append(r"\%o" % cv)
 
             octal = True
 
-    result = result.replace('" "\\', "\\")
+    result = "".join(result).replace('" "\\', "\\")
 
     return '"%s"' % result
 
--- a/nuitka/utils/CStrings.py
+++ b/nuitka/utils/CStrings.py
@@ -19,20 +19,25 @@ def _identifierEncode(c):
 
 
 codecs.register_error("c_identifier", _identifierEncode)
+
+
 def _encodePythonStringToC(value):
-    """Encode a string, so that it gives a C string literal"""
-    assert isinstance(value, bytes), type(value)
-    
+    """Encode a string, so that it gives a C string literal.
+
+    This doesn't handle limits.
+    """
+    assert type(value) is bytes, type(value)
+
     special_chars = {92: r"\134", 9: r"\011", 13: r"\015", 10: r"\012", 63: r"\077", 34: r"\042"}
-    result = []
+    result = ['"']
     octal = False
     for c in value:
         cv = c
 
-        if c in special_chars.keys():
+        if c in special_chars:
             result.append(special_chars[c])
             octal = True
-            
+
         elif 32 <= cv <= 127:
             if octal and c in b"0123456789":
                 result.append(r'" "')
@@ -44,9 +49,10 @@ def _encodePythonStringToC(value):
 
             octal = True
 
+    result.append('"')
     result = "".join(result).replace('" "\\', "\\")
 
-    return '"%s"' % result
+    return result
 
 
 def encodePythonUnicodeToC(value):


nuitka/utils/CStrings.py
```diff
@@ -28,34 +28,31 @@ def _encodePythonStringToC(value):
     """
     assert type(value) is bytes, type(value)
 
-    result = ""
+    special_chars = {92: r"\134", 9: r"\011", 13: r"\015", 10: r"\012", 63: r"\077", 34: r"\042"}
```
kayhayen, line 31: For Python2, this would have to be different, since there `bytes` is not a number.

nuitka/utils/CStrings.py
```diff
@@ -28,34 +28,31 @@ def _encodePythonStringToC(value):
...
-
     for c in value:
-        if str is bytes:
-            cv = ord(c)
-        else:
-            cv = c
-
-        if c in b'\\\t\r\n"?':
-            result += r"\%03o" % cv
+        cv = c
```
kayhayen, line 35: This removes Python2 compatibility, there `ord(c)` is really needed. For performance, of this individual function, giving up on detecting the need inside a loop would probably be best.

nuitka/utils/CStrings.py
```diff
@@ -28,34 +28,31 @@ def _encodePythonStringToC(value):
...
 
+        if c in special_chars:
```
kayhayen, line 37: The special_chars idea is half-way. A prepared dictionary for all 256 values, on the module level, would be easy to make Python2/Python3 difference, and then translate things. If you mixed "octal" into the key and value of that dict, it would be better.

I would criticize that the idea is good, but the execution could be better.

kayhayen (changes requested):

kayhayen:
Rejection this for now as it breaks Nuitka with Python2 which we still care about, and this is not really a good code change anyway.

PR #2814, opened 2024-04-17
Cross compilation to WebAssembly / WASI

syrusakbary:
This PR aims to support to enable cross-compilation to other architectures, and more namely WebAssembly and WASI.
We've detailed a bit more in the announcement blogpost (which should be released soon: https://wasmer.io/posts/py2wasm-a-python-to-wasm-compiler )

Ideally, we would like most of the changes to be in Nuitka, and py2wasm being a very thin layer on top of that. 
More patches will be needed but this PR more or less what enables the bulk work (I do not want to overwhelm the maintainer with too many changes at once).

To summarize, this PR should achieve the following:
* Being able to cross-compile / serialize&desserialize to 32 bit architecture (Wasm32, WASI)
* Added support for some hacks required to compile to WASI


nuitka/PythonVersions.py
```diff
@@ -388,7 +388,7 @@ def isPythonValidDigitValue(value):
     return -_float_digit_boundary <= value <= _float_digit_boundary
 
 
-sizeof_clong = ctypes.sizeof(ctypes.c_long)
+sizeof_clong = 4  # ctypes.sizeof(ctypes.c_long)
```
kayhayen, line 391: Does this really mean, ctypes is wrong there for that target?

nuitka/build/static_src/HelpersFilesystemPaths.c
```diff
@@ -618,6 +624,10 @@ char const *getBinaryFilenameHostEncoded(bool resolve_symlinks) {
     // Resolve any symlinks we were invoked via
     resolveFileSymbolicLink(binary_filename_target, binary_filename_target, buffer_size, resolve_symlinks);
 
+#elif defined(__wasi__)
+    const char *wasi_filename = "program.wasm";
```
kayhayen, line 628: That looks wrong to me, hardcoding that, just as an experiment?

nuitka/build/static_src/HelpersFilesystemPaths.c
```diff
@@ -889,15 +899,18 @@ bool expandTemplatePath(char *target, char const *source, size_t buffer_size) {
                 appendStringSafe(target, pid_buffer, buffer_size);
             } else if (strcasecmp(var_name, "HOME") == 0) {
                 char const *home_path = getenv("HOME");
-
                 if (home_path == NULL) {
+#ifdef __wasi__
```
kayhayen, line 903: That's going to cause issues with much of standalone, etc. but I guess, that's not a topic for that platform?

nuitka/tools/data_composer/DataComposer.py
```diff
@@ -135,9 +135,9 @@ def _writeConstantValue(output, constant_value):
             _writeConstantValue(output, element)
     elif constant_type is long:
         if isPythonValidCLongValue(constant_value):
-            output.write(b"l" + struct.pack("l", constant_value))
+            output.write(b"l" + struct.pack("=l", constant_value))
```
kayhayen, line 138: I am confused why the representation would have to change without a change on the decoding side. Are we saying that network byte order vs. native order is something the decoder doesn't do properly?

kayhayen (commented):
This looks like it's not a lot yet, I hope there is instructions somewhere how to reproduce it, I am confused about the representation changes and the hard coded long change, that Python doesn't know about in ctypes, is that also unsupported as a module?

kayhayen:
So, cross compilation has not been supported for a long time, however, this looks like a good case for getting it started, esp. if you are willing to help. I invite you to reach out to me in private email if you can allocate a bit of resources to it, and we can split to work.

In my plan, it takes a slave Python process, that is capable of answering questions. Things like `os.name` and so on, that are compile time evaluated, are currently all hidden behind "hard imports" that e.g. indicate it to be  constant value, and then there are modules, allow listed to be used with getattr that turn the result into compile time values.

There are also functions like

```
def getComputationResult(node, computation, description, user_provided):
    """With a computation function, execute it and return constant result or
    exception node.

    """
```

that are used by the core of Nuitka optimization a bunch. Other places that you found, would e.g. have to cover encoding a value for the target platform with `struct.pack` and the like. Many places are already allowing for proper abstraction of things, and the challenge would be to find ways to not loose a bunch of compile time performance for the normal case.

The plan would be to make an interface, that gets asked questions, and responds with either a slave Python used, which in your case would be a WASI one, where computations are actually executed. Communication to this slave Python (in my goal set) is like is in the transport level SSH and/or subprocess input/output, and cached. This interface can then be chosen. You can lie in that interface all you want, or hard code things if you feel the need for it, but I would prefer caching to be used over hard coding, if we discover this to be slow. For the standard case, said interface would be trivial, but it would provide the hooks needed, to know what WASI gives for `os.name` without us knowing what it does.

This interface could then be implemented by something that implements more or less things. For Nuitka, I am also interested in cross version compilation. Like being able to compile Python 3.7 as a target Python on 3.13 with no-GIL threads and being able to move the Nuitka core from 2.6 / 3.x dual lingo, to more modern Python.

Another side of the interface would be to also plug Scons C compiler selection and configuration.

I would envision that Nuitka plugins end up doing these things, with the limitation probably that only one plugin can handle the slave Python interface. The plugins interface of Nuitka however, should probably provide the infrastructure for communication between the two Python interpreters. The pickle might be a good start, but I seem to recall that it is incapable of doing everything, we might have to do something, that allows for the host Python to speak of values and types, that only the slave Python actually knows for real, but this can be expanded over time. I am assuming, that you can live with matching the Python version, and I can later figure out how to unlock current code to produce 3.13 programs when Nuitka is running on 2.6 and vice versa.

nuitka/tools/data_composer/DataComposer.py
kayhayen: I now figured out, that you didn't run Nuitka on WASI, which is fine by me, but normally would be the recommended approach.

syrusakbary:
Awesome @kayhayen . I'll send you an email soon!

jedisct1:
`dlopen()` is always going to fail, so I'd suggest adding something like:

```diff
--- a/nuitka/build/static_src/MetaPathBasedLoader.c
+++ b/nuitka/build/static_src/MetaPathBasedLoader.c
@@ -1181,6 +1181,7 @@ static PyObject *_nuitka_loader_load_module(PyObject *self, PyObject *args, PyOb
     PyThreadState *tstate = PyThreadState_GET();

 #ifndef _NUITKA_STANDALONE
+# ifndef __wasi__
     if (installed_extension_modules != NULL) {
         PyObject *extension_module_filename = DICT_GET_ITEM0(tstate, installed_extension_modules, module_name);

@@ -1189,6 +1190,7 @@ static PyObject *_nuitka_loader_load_module(PyObject *self, PyObject *args, PyOb
             return callIntoInstalledExtensionModule(tstate, module_name, extension_module_filename);
         }
     }
+# endif
 #endif

     return IMPORT_EMBEDDED_MODULE(tstate, name);
@@ -1598,6 +1600,7 @@ static PyObject *_nuitka_loader_exec_module(PyObject *self, PyObject *args, PyOb
     // During spec creation, we have populated the dictionary with a filename to load from
     // for extension modules that were found installed in the system and below our package.
 #ifndef _NUITKA_STANDALONE
+# ifndef __wasi__
     if (installed_extension_modules != NULL) {
         PyObject *extension_module_filename = DICT_GET_ITEM0(tstate, installed_extension_modules, module_name);

@@ -1613,6 +1616,7 @@ static PyObject *_nuitka_loader_exec_module(PyObject *self, PyObject *args, PyOb
             return callIntoInstalledExtensionModule(tstate, module_name, extension_module_filename);
         }
     }
+# endif
 #endif

     return EXECUTE_EMBEDDED_MODULE(tstate, module);
```

This allows `zig cc` to be used as the wasm compiler.

kayhayen:
For zig cc we did a bunch of work.  For 3.11 or higher we got a workaround that uses their C++ compiler or what we call non C11 mode

bitnom:
very interested in this

lum1n0us:
It seems there is no update and comment recently. May I ask about the merging plan and blocking points about this feature? It will be very helpful for Wasm world to have python as a new guest language. We are very looking forward to it.

kayhayen:
I am currently in a crunch for 2.3 release, I will get in touch with the PR authors once I have the time for it after release. Until then 3.12 support was more important.

syrusakbary:
No worries @kayhayen , just ping me once things are ready to go!

derekbar90:
Awesome work! Very exciting to see this coming through 💪

kayhayen:
I scheduled a conference, let me rebase the branch for you to 2.3, if I can, I do rebase for bisection at release times

kayhayen:
Seems, I didn't get the right to change the PR branch, so what I am going to do, is to cherry pick the parts that conflicted into factory, so that you can just drop those C parts with wasm defines. The OpenBSD changes overtook you there.

kayhayen:
So, my merge results are on the factory branch. When you git rebase to the factory branch, simply `git rebase --skip`, and you should be good. I would find it ideal for the telco to see this already applied. For the C changes that I merge, I will ask you some questions, to add comments or clarify error messages. I didn't merge your Python parts, for which we will need to do some abstraction.

AlmogBaku:
any update?

kayhayen:
We are working on it, parts were merged for last release, making many of the original changes unnecessary, and there are other PRs to make this more viable. It's a long term thing though, but the current result is available from them via PyPI even.

lum1n0us:
Is there a command line option(of nuitka) to enable cross-compilation? At the moment, it seems like the only way I can do this is by modifying the py2wasm repository directly

ceelian:
This is great work! Thx a lot. Looking forward to using it!

farhaanbukhsh:
Hey @kayhayen @syrusakbary do we have some steps to test this PR or working example of compiling python to wasm?

lum1n0us:
I hope it can be useful. https://github.com/lum1n0us/wamr-with-py

kayhayen:
How synced up with 2.7, i.e. current develop branch is it? @lum1n0us so I can maybe merge some more to it, I haven't had the time for it recently, but I will have in the coming months

farhaanbukhsh:
Thanks @lum1n0us that worked :)

maxrossi:
Hello, any update on Nuitka cross compilation to WASI?

polarathene:
> How synced up with 2.7, i.e. current develop branch is it?

It appears to run from Nuitka 2.5 and stuck with Python 3.11 apparently. I can confirm it was successful with `python-build-standalone` from UV on a Fedora 42 container.

The current status in Nuitka officially is just a placeholder `--target=wasi` right?
- I had attempted to build with the current `develop` branch but it did not seem to output any wasm build.
- Additionally with Zig wrapper script workarounds (_Nuitka has some integration bugs I'll report_), it was unsuccessful at building for the `wasm32-wasi` target due to some C source. Zig was otherwise successful at building non-wasm target (_it didn't like `-l:libatomic.a` however, but that appears to be an upstream Zig bug_).

PR #2894, opened 2024-05-31
Fix --enable-console warning message

jbjd:
# What does this PR do?

Swaps "disable" and "force" in the warning message for the deprecated --enable-console and --disable-console flags

# Why was it initiated?

The warning currently says this:
"The old console option '--enable-console' should not be given anymore, use '--windows-console-module=disable' instead. It also has the extra mode 'attach' to consider."

It should say --windows-console-module=force when the --enable-console flag is present and vice versa.

### DIFF
--- a/nuitka/Options.py
+++ b/nuitka/Options.py
@@ -1200,7 +1200,7 @@ instead. It also has the extra mode 'attach' to consider."""
                         else "--enable-console"
                     ),
                     "--windows-console-module=%s"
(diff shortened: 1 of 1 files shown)


kayhayen:
I am so shockingly bad at booleans, luckily they don't matter often in computer science. :-)

kayhayen:
I will put this into the hotfix branch directly and on factory now.

PR #2900, opened 2024-06-05
fix "FATAL: Failed unexpectedly in Scons C backend compilation." build error.

CrossDarkrix:
sorry, I will cancel it now.
this pull is for will delete "Compare & pull request" to my project.

My PC is now able to build successfully with this commit.
but Until then I couldn't build anything.
At the time I was plagued by this error: "FATAL: Failed unexpectedly in Scons C backend compilation."
The temporary solution to this error was to ignore the error in the first place.
I was able to build successfully by ignoring the error.


kayhayen:
If you have an issue, raise it

PR #2918, opened 2024-06-12
fix Windows build failure on mapped network drives

jimkring:
`drive` has a value of something like `"Y:"` so when we prepend this to a subpath, which will not include and leading slash, we need to add the path separator.

# What does this PR do?

# PR Checklist


kayhayen (approved):
I am kind of surprised to see this, but I have to admit my testing was not too deep, and `Y:path` can work depending on where you are I guess. Thanks so much for your contribution.

PR #2947, opened 2024-06-24
pygame-ce compat with weird missing dlls

oddbookworm:

# What does this PR do?
[pygame-ce](https://github.com/pygame-community/pygame-ce) is a fork of pygame that uses a custom SDL Image build. For whatever reason, `libpng16-16.dll` and `libjpeg-62.dll` don't get copied into the build. I'm hoping this is a temporary fix (I tested that this seems to still work fine with the original). We have our own issue [here](https://github.com/pygame-community/pygame-ce/issues/2954) that I'm hoping we can eventually resolve on the library side, rather than Nuitka's side. But for now, this is the only solution I've come up with that doesn't require us to point every user to include our vendored in binaries for png and jpeg

pygame-ce issue linked above, no Nuitka issue

# PR Checklist

### DIFF
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -4047,10 +4047,16 @@
     - dirs:
         - '.'
 
-- module-name: 'pygame' # checksum: 8c9db0be
+- module-name: 'pygame' # checksum: f29246e3
   data-files:
     - patterns:
         - 'freesansbold.ttf'
+
+  dlls:
+    - from_filenames:
+        relative_path: '.'
+        prefixes:
+          - ''
 
   options:
     checks:
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -4047,10 +4047,16 @@
     - dirs:
         - '.'
 
-- module-name: 'pygame' # checksum: f29246e3
+- module-name: 'pygame' # checksum: 67ff3d63
   data-files:
     - patterns:
         - 'freesansbold.ttf'
+    - patterns:
+        - 'pygame_icon_mac.bmp'
+      when: 'macos'
+    - patterns:
+        - 'pygame_icon.bmp'
+      when: 'not macos'
 
   dlls:
     - from_filenames:


oddbookworm:
[reproducer.zip](https://github.com/user-attachments/files/15950092/reproducer.zip)

Here's the sample project I was working with
Build command:
```
nuitka --standalone main.py --include-data-files=./alien1.png=alien1.png --include-data-files=./alien1.jpg=alien1.jpg --include-data-files=./Roboto-Regular.ttf=Roboto-Regular.ttf --include-data-files=./boom.wav=boom.wav 
```

oddbookworm:
It’s been pointed out to me that potentially more dlls will have issues, so I need to convert this to a draft

kayhayen:
I recommend installing the commit hook; it autoformats your changes at commit time, which will address the CI error.

oddbookworm:
[reproducer.zip](https://github.com/user-attachments/files/15962839/reproducer.zip)

New version of the reproducer. I ultimately decided that pygame(-ce) needs to glob all dlls because they're runtime dynamic linking. It would be great to be able to have package-level entrypoints like pyinstaller does so that this doesn't have to belong in Nuitka, but I suppose this is how it must be for now.

Note that this version of the reproducer fails on both pygame and pygame-ce, it just so happens that libpng is fine on the original because it's not dynamically linked by their version of sdl-image like pygame-ce's is. I could list out all the runtime dll dependencies for one or the other here, but that would break the other version, so wildcard glob is the best choice here

Starbuck5:
> custom SDL Image build

It’s not that it’s custom, it’s that it’s new. And it also doesn’t include webp and tiff DLLs. And also DLLs for loading audio formats.

KRRT7:
the reproducer doesn't reproduce, i get this
```
❯ python main.py
pygame-ce 2.5.0 (SDL 2.30.3, Python 3.11.9)
  File "C:\Users\NuitkaDevOps\Desktop\Nuitka-Issues\pyg_ce\main.py", line 17, in <module>
    pygame.mixer.music.load("MIDI_sample.mid")
pygame.error: mixer not initialized
```

oddbookworm:
> the reproducer doesn't reproduce, i get this
> 
> ```
> ❯ python main.py
> pygame-ce 2.5.0 (SDL 2.30.3, Python 3.11.9)
> Traceback (most recent call last):
>   File "C:\Users\NuitkaDevOps\Desktop\Nuitka-Issues\pyg_ce\main.py", line 17, in <module>
>     pygame.mixer.music.load("MIDI_sample.mid")
> pygame.error: mixer not initialized
> ```



> the reproducer doesn't reproduce, i get this
> 
> ```
> ❯ python main.py
> pygame-ce 2.5.0 (SDL 2.30.3, Python 3.11.9)
> Traceback (most recent call last):
>   File "C:\Users\NuitkaDevOps\Desktop\Nuitka-Issues\pyg_ce\main.py", line 17, in <module>
>     pygame.mixer.music.load("MIDI_sample.mid")
> pygame.error: mixer not initialized
> ```

That definitely shouldn't be happening, line 5 of `main.py` initializes mixer in the call to `pygame.init()`. Here's what I get using unmodified nuitka
```
pygame-ce 2.5.0 (SDL 2.30.3, Python 3.12.0)
  File "C:\Users\andre\Projects\temp\MAIN~1.DIS\main.py", line 37, in <module>
  File "C:\Users\andre\Projects\temp\MAIN~1.DIS\main.py", line 23, in __init__
pygame.error: Failed loading libpng16-16.dll: The specified module could not be found.
```
Oh and here's the build command I used
```
nuitka --standalone main.py --include-data-files=./alien1.jpg=alien1.jpg --include-data-files=./Roboto-Regular.ttf=Roboto-Regular.ttf --include-data-files=./boom.wav=boom.wav --include-data-files=./boom.ogg=boom.ogg --include-data-files=./alien1.png=alien1.png --include-data-files=./bg.tiff=bg.tiff --include-data-files=./MIDI_sample.mid=MIDI_sample.mi
```

KRRT7:
i'm not using nuitka at all there, simply doing `python main.py`
```
Administrator in Desktop\Nuitka-Issues\_pygrepro via 🐍 v3.11.9 (venv)
❯ type main.py
import pygame
import pygame.freetype
import pygame.midi

pygame.init()
pygame.freetype.init()

screen = pygame.display.set_mode((500, 500))
clock = pygame.time.Clock()

# bg = pygame.image.load("bg.tiff")
# bg = pygame.transform.scale(bg, screen.get_size())

# pygame.Font()
pygame.freetype.Font("Roboto-Regular.ttf")

pygame.mixer.music.load("MIDI_sample.mid")
pygame.mixer.music.play(-1)

class Player:
    def __init__(self, pos: pygame.Vector2) -> None:
        self.pos = pos.copy()
        self.image = pygame.image.load("alien1.png")
        other_image_for_test = pygame.image.load("alien1.jpg") #noqa
        self.speed = 200 # pixels per second

    def move(self, direction: pygame.Vector2, dt: float) -> None:
        if direction:
            direction.normalize_ip()

        self.pos += direction * self.speed * dt

    def draw(self, screen: pygame.Surface) -> None:
        screen.blit(self.image, self.pos)

player_pos = pygame.Vector2(screen.get_rect().center)
player = Player(player_pos)

running = True
while running:
    dt = clock.tick() / 1000

    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            running = False

    direction = pygame.Vector2()
    keys = pygame.key.get_pressed()
    if keys[pygame.K_LEFT] or keys[pygame.K_a]:
        direction.x -= 1
    if keys[pygame.K_RIGHT] or keys[pygame.K_d]:
        direction.x += 1
    if keys[pygame.K_UP] or keys[pygame.K_w]:
        direction.y -= 1
    if keys[pygame.K_DOWN] or keys[pygame.K_s]:
        direction.y += 1

    player.move(direction, dt)

    screen.fill("black")
    # screen.blit(bg, (0, 0))
    player.draw(screen)
    pygame.display.flip()

```

this is the error that happens when i do the command 
```
Administrator in Desktop\Nuitka-Issues\_pygrepro via 🐍 v3.11.9 (venv)
❯ python main.py
pygame-ce 2.5.0 (SDL 2.30.3, Python 3.11.9)
  File "C:\Users\NuitkaDevOps\Desktop\Nuitka-Issues\_pygrepro\main.py", line 17, in <module>
    pygame.mixer.music.load("MIDI_sample.mid")
pygame.error: mixer not initialized
```
my main.py is the same as your main.py

oddbookworm:
Alright, you can just comment out these lines, they're not important to the reproducer. That's super weird though
```
pygame.mixer.music.load("MIDI_sample.mid")
pygame.mixer.music.play(-1)
```

KRRT7:
in which case i suggest using 
```yaml
- module-name: 'pygame'
  data-files:
    patterns:
      - 'freesansbold.ttf'
  dlls:
    - from_filenames:
        prefixes:
          - 'lib'
          - 'SDL2'
          - 'freetype'
          - 'portmidi'
  options:
    checks:
      - description: 'PyGame is a GUI framework'
        macos_bundle: 'recommend'
```
rather than the glob pattern, this also pulls in 

>  webp and tiff DLLs. And also DLLs for loading audio formats.

KRRT7:
doing a quick scan for DLLs in pygame lists all these DLLs
```
freetype.dll
libjpeg-62.dll
libogg-0.dll
libopus-0.dll
libopusfile-0.dll
libpng16-16.dll
libtiff-5.dll
libwavpack-1.dll
libwebp-7.dll
libwebpdemux-2.dll
libxmp.dll
portmidi.dll
SDL2.dll
SDL2_image.dll
SDL2_mixer.dll
SDL2_ttf.dll
```
all of which are pulled by the yaml config above

Starbuck5:
Doing a search on “lib” is pretty similar to just grabbing all executables. It’s not like it adds more protection or anything?

But if we or SDL change how they build those, the lib prefix may not be included in the future (it’s really more of a Linux thing), so this seems more liable to break.

pygame-ce won’t ship DLLs in our wheels unless they are needed.

KRRT7:
in which case 
```yaml
- module-name: 'pygame'
  data-files:
    patterns:
      - 'freesansbold.ttf'
  dlls:
    - from_filenames:
        relative_path: '.'
        prefixes:
          - ''
  options:
    checks:
      - description: 'PyGame is a GUI framework'
        macos_bundle: 'recommend'
```
but if you look at all our DLL configurations, nuitka doesn't like using glob patterns for DLLs, even if supported, maybe we should make that more explicit, we prefer being explicit and only including the DLLs we know that are absolutely needed

oddbookworm:
> but if you look at all our DLL configurations, nuitka doesn't like using glob patterns for DLLs, even if supported, maybe we should make that more explicit, we prefer being explicit and only including the DLLs we know that are absolutely needed

I would generally agree with this, and I hate wildcarding it, but I don't see much of a choice here that will satisfy everyone. The ideal solution IMO would be for nuitka to have entry points that 3rd party libraries can implement themselves where they can handle the specification of dynamic libraries that don't get caught by the dependency checker in nuitka. However, that isn't an option right now. We have the complicating consideration that we have to share a config between pygame and pygame-ce because they both install to the pygame directory. Sure, we can do explicit like you suggest, but that's got two big problems for us to consider:
1) we have absolutely no say in what pygame does or what dependencies they use, so if they end up with a dependency we don't have that doesn't happen to match any pattern, then their stuff will break. I don't want to end up with Nuitka being in the middle of any future drama between pygame and pygame-ce that may result from one or the other breaking the dependencies of the other
2) we have a lot of dependencies that are constantly changing, and soon we'll have to worry more about what SDL 3.0.0+ is doing, which is going to be a lot. I don't want to have to come make a pull request to nuitka every time SDL changes something about their dynamic dependencies.

KRRT7:
> but I don't see much of a choice here that will satisfy everyone.

the empty relative dir and prefix will pull in *all* DLLs within the module. that should sastify everyone i believe.

> pygame-ce won’t ship DLLs in our wheels unless they are needed.
provided that this keeps being true, i think we will be happy with the above approach, we just don't like including bloat DLLs as this greatly increases the final binary size

oddbookworm:
> the empty relative dir and prefix will pull in _all_ DLLs within the module. that should sastify everyone i believe.

out of curiosity, what's the difference between this and what I have? I don't mind making the change as long as the behavior is the same

oddbookworm:
Alright, I've made the changes. Thanks @KRRT7!

oddbookworm:
> ```
> Administrator in Desktop\Nuitka-Issues\_pygrepro via 🐍 v3.11.9 (venv)
> ❯ python main.py
> pygame-ce 2.5.0 (SDL 2.30.3, Python 3.11.9)
> Traceback (most recent call last):
>   File "C:\Users\NuitkaDevOps\Desktop\Nuitka-Issues\_pygrepro\main.py", line 17, in <module>
>     pygame.mixer.music.load("MIDI_sample.mid")
> pygame.error: mixer not initialized
> ```

As for this, I'm slightly concerned about it. Was this run on a local machine, or was it run though something that doesn't have audio capability?
It's possible mixer failed to initialize and the return value of `pygame.init()` would have revealed at least one module failing to initialize, but I don't immediately know what else could have caused that. If it was local, could you provide the output of `pygame.print_debug_info()`? (that function only exists in pygame-ce). I'd at least like to be able to open an issue about it on our repo so that it can be investigated if it's a legitimate issue

KRRT7:
> was it run though something that doesn't have audio capability?
that'd do it, it's running on an Azure windows server.


```
pygame-ce 2.5.0 (SDL 2.30.3, Python 3.11.9)
Platform:               Windows-10-10.0.22631-SP0
System:                 Windows
System Version:         10.0.22631
Processor:              Intel64 Family 6 Model 85 Stepping 7, GenuineIntel
Architecture:           Bits: 64bit     Linkage: WindowsPE

Python:                 CPython 3.11.9 (tags/v3.11.9:de54cf5, Apr  2 2024, 10:12:12) [MSC v.1938 64 bit (AMD64)]
pygame version:         2.5.0
SDL versions:           Linked: 2.30.3  Compiled: 2.30.3
SDL Mixer versions:     Linked: 2.8.0   Compiled: 2.8.0
SDL Font versions:      Linked: 2.22.0  Compiled: 2.22.0
SDL Image versions:     Linked: 2.8.2   Compiled: 2.8.2
Freetype versions:      Linked: 2.11.1  Compiled: 2.11.1

Display Driver:         Display Not Initialized
Mixer Driver:           Mixer Not Initialized
```

oddbookworm:
Apologies, I missed one additional step in there. Can you call pygame.display.init() and pygame.mixer.init() before the print?

KRRT7:
```
pygame-ce 2.5.0 (SDL 2.30.3, Python 3.11.9)
  File "C:\Users\NuitkaDevOps\Desktop\Nuitka-Issues\_pygrepro\main.py", line 6, in <module>
    pygame.mixer.init()
pygame.error: WASAPI can't find requested audio endpoint: Element not found.
```

oddbookworm:
Thanks. Let's move this back on topic lol. Lemme know if I should make any further changes

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -3823,11 +3823,17 @@
     dirs:
       - '.'
 
-- module-name: 'pygame' # checksum: 5fbb83cf
+- module-name: 'pygame' # checksum: f29246e3
   data-files:
     patterns:
       - 'freesansbold.ttf'
```
ankith26, line 3829: This isn't related to the DLL issue, but is something else I noticed that would be nice to fix.

Here is the logic we use in the pygame-ce pyinstaller hook for including the pygame icon file, would be nice to have this in here as well, maybe just include anything that matches the `pygame_icon` prefix?

```py
if platform.system() == "Darwin":
    _append_to_datas("pygame_icon_mac.bmp")
else:
    _append_to_datas("pygame_icon.bmp")
```

Maxwell175: The proper way is to make separate blocks for each platform so just the necessary files are used.

kayhayen: Actually the schema outlaws that, but the code will work with it, but I will change that for 2.4 release.

Something like this ought to work:

```
data-files:
    - patterns:
        - 'freesansbold.ttf'
    - patterns:     
         - 'pygame_icon_mac.bmp'
      when: 'macos'
    - patterns:     
         - 'pygame_icon.bmp'
      when: 'not macos'
```

Previously we never had a need for this to be a list, but apparently, we should. I will make a transition for develop right before 2.4 releases, you can already use the code in the PR and igore the checker error.

oddbookworm: > Actually the schema outlaws that, but the code will work with it, but I will change that for 2.4 release.
> 
> Something like this ought to work:
> 
> ```
> data-files:
>     - patterns:
>         - 'freesansbold.ttf'
>     - patterns:     
>          - 'pygame_icon_mac.bmp'
>       when: 'macos'
>     - patterns:     
>          - 'pygame_icon.bmp'
>       when: 'not macos'
> ```
> 
> Previously we never had a need for this to be a list, but apparently, we should. I will make a transition for develop right before 2.4 releases, you can already use the code in the PR and igore the checker error.

I might be misunderstanding what you're saying. Using this version, the commit hook yells at me that it's not valid, so it won't even let me commit. If I remove the `-` from before `patterns`, it then spits out a duplicate key error because `patterns` is repeated

oddbookworm: Here's what I mean:
```yaml
- module-name: 'pygame' # checksum: f29246e3
  data-files:
    - patterns:
      - 'freesansbold.ttf'
    - patterns:
      - 'pygame_icon_mac.bmp'
      when: 'macos'
    - patterns:
      - 'pygame_icon.bmp'
      when: 'not macos'

  dlls:
    - from_filenames:
        relative_path: '.'
        prefixes:
          - ''

  options:
    checks:
      - description: 'PyGame is a GUI framework'
        macos_bundle: 'recommend'
```
yields this error 
```
  File "C:\Users\andre\Projects\Nuitka\bin\autoformat-nuitka-source", line 20, in <module>
    main()
  File "C:\Users\andre\Projects\Nuitka\nuitka\tools\quality\auto_format\__main__.py", line 105, in main
    autoFormatFile(desc["src_path"], git_stage=desc)
  File "C:\Users\andre\Projects\Nuitka\nuitka\tools\quality\auto_format\AutoFormat.py", line 798, in autoFormatFile
    formatYaml(tmp_filename, ignore_diff=ignore_yaml_diff)
  File "C:\Users\andre\Projects\Nuitka\nuitka\tools\quality\auto_format\YamlFormatter.py", line 313, in formatYaml
    sorted_entry["data-files"] = _reorderDictionary(
                                 ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\andre\Projects\Nuitka\nuitka\tools\quality\auto_format\YamlFormatter.py", line 159, in _reorderDictionary
    entry._items(),
    ^^^^^^^^^^^^
AttributeError: 'CommentedSeq' object has no attribute '_items'
```

kayhayen:
Yeah, the schema update is very much needed, I will do it for 2.4 really soon. The commit hook is designed to catch such violations, but once I make the change, you can do it. You can trick it with `COMMIT_UNCHECKED=1` environment, but that's not the way it's supposed to be of course.

Thanks for your patience.

oddbookworm:
I'll definitely wait until after the schema update is made to push the extra data files. If I skip the pre-commit hook, then it'll still be blocked by the schema stuff unless you want every pull request between this one and the schema to have to skip the hook too

oddbookworm:
As far as I know, this should be good to go. Need a maintainer to run the CI though to see if I borked anything

oddbookworm:
I verified locally that it worked with my reproducer

kayhayen:
This is going to be in 2.4.8, I made the invasive hotfix 2.4.6 to make that easy to do, now that's of course bound to happen, but I need to accumulate more important changes to actually push that out, but usually that happens in a week max.

PR #3008, opened 2024-07-15
improved arcade configuration

DragonMoffon:
# What does this PR do?
The arcade package was bundling megabytes on unnecessary data which is now excluded

# PR Checklist

### DIFF
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -163,7 +163,11 @@
 - module-name: 'arcade' # checksum: 3f9f0b7d
   data-files:
     dirs:
-      - 'resources'
+      - 'resources\system'
+  anti-bloat:
+    - no-auto-follow:
+        'examples': 'ignore'
+        'experimental': 'ignore'
 
 - module-name: 'aspose' # checksum: 9babf530
   data-files:
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -163,7 +163,7 @@
 - module-name: 'arcade' # checksum: 3f9f0b7d
   data-files:
     dirs:
-      - 'resources/system'
+      - 'resources/system/'
   anti-bloat:
     - no-auto-follow:
         'examples': 'Arcade examples are not be available unless you do "--include-module=acade.examples"'
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -163,11 +163,11 @@
 - module-name: 'arcade' # checksum: 3f9f0b7d
   data-files:
     dirs:
-      - 'resources\system'
+      - 'resources/system'
   anti-bloat:
     - no-auto-follow:
-        'examples': 'ignore'
-        'experimental': 'ignore'
+        'examples': 'Arcade examples are not be available unless you do "--include-module=acade.examples"'
+        'experimental': 'Experimental features of Arcade will not be available unless you do "--include-module=acade.experimental"'
 
 - module-name: 'aspose' # checksum: 9babf530
   data-files:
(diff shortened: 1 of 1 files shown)


nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -163,7 +163,11 @@
 - module-name: 'arcade' # checksum: 3f9f0b7d
   data-files:
     dirs:
-      - 'resources'
+      - 'resources\system'
```
kayhayen, line 166: That slash form ought to be wrong, although I can imagine it works on Windows, and this might be a Windows only package, so I might want to do something about this, such that auto-format cleans this up or something like that.

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -163,7 +163,11 @@
...
+  anti-bloat:
+    - no-auto-follow:
+        'examples': 'ignore'
+        'experimental': 'ignore'
```
kayhayen, line 170: Just guessing here, but I assume experimental feature usage is rare, in that case, you should tell the user about it, and inform them that experimental features are not going to be available by not putting 'ignore' there.

An example of this would be

```
- module-name: 'joblib.externals.cloudpickle.cloudpickle' # checksum: 1d738df0
  anti-bloat:
    - no-auto-follow:
        'tornado': 'can break tornado integration of joblib'
```

Although, it probably should say. "Experimental features of arcade will not be available unless you do '--include-module=acade.experimental'" so the new user knows about this and also knows the remedy. I agree these texts are not generally as helpful as they ought to be.

nuitka/plugins/standard/standard.nuitka-package.config.yml
DragonMoffon: Yeah, this is actually going to require so fixing on Arcade's side because it uses Windows style slashes everywhere, so doing it correctly here causes issues.

thank you for the heads up

nuitka/plugins/standard/standard.nuitka-package.config.yml
DragonMoffon: That makes a lot of sense. I will also update Arcade's docs with the same info. Thank you.

DragonMoffon: `FATAL: Error, cannot add data file with 'arcade\resources/system\gamecontrollerdb.txt' path, as 'contains illegal character '/''`

nuitka/plugins/standard/standard.nuitka-package.config.yml
kayhayen: Ah, that's Nuitka internal, after changes I made it will just require you to change it over to `resources/system` and work fine. It's just because Windows accepts forward slashes, while Linux hates backlashes and os.path.normpath doesn't correct them there. You PR is not incorrect because of this, but I actually have checks for dest_path and the like that would have caught you well, if `posixpath.normpath` worked the way I think. I can switch this over with no effects. You code can do whatever it pleases in arcade.

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -163,7 +163,11 @@
 - module-name: 'arcade' # checksum: 3f9f0b7d
   data-files:
     dirs:
-      - 'resources'
+      - 'resources/system/'
```
kayhayen, line 166: That should actually be wrong, to have a trailing slash, pretty sure the checker catches it even.

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -163,7 +163,11 @@
...
+  anti-bloat:
+    - no-auto-follow:
+        'examples': 'Arcade examples are not be available unless you do "--include-module=acade.examples"'
```
kayhayen, line 169: Is arcade really doing imports of those, I seem to recall you saying it does not, then you just don't need to handle that. I understand that for experimental it might, there it makes sense to have it then.

kayhayen:
I did a rebase for the 2.4 release, do not worry, I will resolve this for you, provided I have push rights or merge it directly otherwise for the 2.5 release cycle and likely into 2.4.1 hotfix.

Issue #3012, opened 2024-07-19
[wasm32-wasi] About function pointer casting in build/static_src


lum1n0us:
I notice there is a widely coding pattern in *build/static_src/CompiledXType.c*. It will cause [a trap during execution](https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-control-mathsf-call-indirect-x-y)(Step 18.) when using as a wasm32-wasi target.

Take *CompiledFunctionType.c* as an example. 
- Define `Nuitka_Function_get_name` as `static PyObject *Nuitka_Function_get_name(struct Nuitka_FunctionObject *function)`. With one argument.
- When refer it in `static PyGetSetDef Nuitka_Function_getset[]`, cast it to `getter`(from Include/descrobject.h). It means cast `PyObject*(struct Nuitka_FunctionObject*)` -> `PyObject *(PyObject *, void *)`. 
- Therefore, callers will execute `Nuitka_Function_get_name` in a form likes `get(obj, closure)`(`getset_get()` in *Objects/descrobject.c*).

From caller side, callee is a function with two arguments. On the other side, callee is a function with one argument. It works well in C. But it confuses wasm.  

A call with function pointers will be transferred into `call_indirect` in Wasm, and pack the callee type together. During execution, there is a type check between expected function type(like `PyObject*(void*,void*)`) and actually function type(`PyObject*(void*)`) (all pointers are same types). Eventually, the checker will hit a trap because of different amount of arguments.

---
P.S.

I am thinking about using a different function signature for wasm32-wasi target like:

``` c
#if __wasi__
static PyObject *Nuitka_Function_get_name(struct Nuitka_FunctionObject *function, void *data) {
#else
static PyObject *Nuitka_Function_get_name(struct Nuitka_FunctionObject *function) {
#endif
```

kayhayen:
I am OK with these functions getting modified to specify an unused argument instead. But ideally we also it it in a way, where the cast to ``getter`` and ``setter`` becomes unnecessary, that means we accept PyObject *self and cast it to the function object instead, this can only be good at catching errors. If you were to do that, it will be very acceptable.

lum1n0us:
Something like?

``` c
#if __wasi__
static PyObject *Nuitka_Function_get_name(PyObject *self, void *data) {
#else
static PyObject *Nuitka_Function_get_name(PyObject *self,) {
#endif
[...]

PR #3029, opened 2024-07-28
follow PyCFunction signatures

lum1n0us:

# What does this PR do?

#3012

# PR Checklist

- ~~[ ] Ideally new features or fixed regressions ought to be covered via new tests.~~
- ~~[ ] Ideally new or changed features have documentation updates.~~

--- 
- [ ] naming 
- ~~[ ] TODOs in nuitka/build/static_src/InspectPatcher.c~~
- ~~[ ] TODOs in nuitka/build/static_src/MetaPathBasedLoader.c~~
- ~~[ ] TODOs in nuitka/build/static_src/MetaPathBasedLoaderResourceReader.c~~

### DIFF
--- a/nuitka/build/static_src/CompiledFrameType.c
+++ b/nuitka/build/static_src/CompiledFrameType.c
@@ -543,7 +543,7 @@ static int Nuitka_Frame_tp_traverse(struct Nuitka_FrameObject *frame, visitproc 
 
 #if PYTHON_VERSION >= 0x300
 
-static PyObject *Nuitka_Frame_clear(struct Nuitka_FrameObject *frame) {
+static PyObject *Nuitka_Frame_clear(struct Nuitka_FrameObject *frame, PyObject *unused) {
     assert(Nuitka_Frame_CheckExact((PyObject *)frame));
     CHECK_OBJECT((PyObject *)frame);
     assert(_PyObject_GC_IS_TRACKED(frame));
@@ -637,7 +637,7 @@ static inline Py_ssize_t Nuitka_Frame_GetSize(struct Nuitka_FrameObject *frame) 
 #endif
 }
 
-static PyObject *Nuitka_Frame_sizeof(struct Nuitka_FrameObject *frame) {
+static PyObject *Nuitka_Frame_sizeof(struct Nuitka_FrameObject *frame, PyObject *unused) {
     assert(Nuitka_Frame_CheckExact((PyObject *)frame));
     CHECK_OBJECT((PyObject *)frame);
     assert(_PyObject_GC_IS_TRACKED(frame));
--- a/nuitka/build/static_src/CompiledFunctionType.c
+++ b/nuitka/build/static_src/CompiledFunctionType.c
@@ -660,7 +660,7 @@ static PyGetSetDef Nuitka_Function_getset[] = {
      NULL},
     {NULL}};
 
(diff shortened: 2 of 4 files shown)


kayhayen:
Seems, sometimes I defined even kwds for functions with no arguments, see `Nuitka_ResourceReaderFiles_is_file` which is confusing. I would prefer instead of naming the added value `data` or `value` that it would be called `_unusable` in cases, and where it's missing, name it `args` e.g. `Nuitka_ResourceReaderFiles_exit` takes arguments, but ignores them.

lum1n0us:
There are three arguments in [Nuitka_ResourceReaderFiles_read_text()](https://github.com/Nuitka/Nuitka/blob/develop/nuitka/build/static_src/MetaPathBasedLoaderResourceReaderFiles.c#L170). All are used. What is your opinion if trying to change its signature to `PyObject *(*)(PyObject *, PyObject *)`?

lum1n0us:
> see Nuitka_ResourceReaderFiles_is_file which is confusing. I would prefer instead of naming the added value data or value that it would be called _unusable in cases

Like?

```c
static PyObject *Nuitka_ResourceReaderFiles_is_dir(struct Nuitka_ResourceReaderFilesObject *files, PyObject *args,
                                                   PyObject *kwds) {
```
->
```c
static PyObject *Nuitka_ResourceReaderFiles_is_dir(struct Nuitka_ResourceReaderFilesObject *files, PyObject *args_unusable,
                                                   PyObject *kwds_unusable) {

```

kayhayen:
As it is define `METH_NOARGS` I was thinking it would be different. Nuitka calls those like this:

```PyObject *result = (*method)(self, NULL);```

And the signature you use should only be used when `METH_VARARGS` is set. So either Nuitka makes calling code different there from CPython if you say it through that, or your example is wrong. Unused keyword arguments would work that way though, yes.

**Comment by kayhayen** [comment]
Eh, no, unusable keyword arguments should not exist, which is what your example states.

**Comment by lum1n0us** [comment]
The problem is found by below case(need to compile to .wasm):

``` python
import unittest

# A simple function to illustrate
def parse_int(s):
    return int(s)

class TestConversion(unittest.TestCase):
    # Testing that an exception gets raised
    def test_bad_int(self):
        self.assertRaises(ValueError, parse_int, "N/A")  # <| -- exception HERE!
```

It turns out
1. `parse_int()` throws `ValueError` exception because N/A is invalid
2. `unittest` catches the exception and handles it with `__exit()` of `_AssertRaisesContext`.
3. When `__exit()` uses `traceback.clear_frames()` to clear frames, it goes to `Nuitka_Frame_clear()`.
4. `Nuitka_Frame_clear(struct Nuitka_FrameObject *frame)` doesn't fit the expectation of `PyCFunction` which is `PyObject*(*PyCFunction)(PyObject*, PyObject*).

Since CompiledCellType, CompiledGeneratorType, CompiledFunctionType, CompiledMethodType and CompiledFrameType are widely used, their unsafe function pointer type castings(to PyCFunction) are fixed too.

kayhayen:
This looks good to me so far, let me know when you feel it's complete. I have to say that you ought to remove `(PyCFunction)` casts from Nuitka probably and then you will find them at compile time immediately and all of them. Basically all these casts are creating issues for WASI I guess.

lum1n0us:
Yes. They are.

That's all for the casting PR. It is good to go.

PR #3033, opened 2024-07-30
Add data file for `license-expression` package

kdeldycke:
# What does this PR do?

Adds the [`data` folder](https://github.com/nexB/license-expression/tree/main/src/license_expression/data) of the [`license-expression`](https://github.com/nexB/license-expression) package is where the license database is stored. While seems to be available is the source distribution, Nuitka has not enough metadata to import it naturally.

# Why was it initiated? Any relevant Issues?

I stumble upon this issue while trying to make a standalone executable of my [Meta Package Manager](https://github.com/kdeldycke/meta-package-manager) project, which depends on [`cyclonedx-python-lib`](https://github.com/CycloneDX/cyclonedx-python-lib) for SBOM export, itself depending on `license-expression`.

If [compiling of my CLI ends up without any error](https://github.com/kdeldycke/meta-package-manager/actions/runs/10150356524/job/28067304207#step:7:39), I get an [immediate issue while running it](https://github.com/kdeldycke/meta-package-manager/actions/runs/10150356524/job/28068453054#step:4:42):
```pytb
Traceback (most recent call last):
  File "/tmp/onefile_1690_1722281965_555845/__main__.py", line 49, in <module>
  File "/tmp/onefile_1690_1722281965_555845/__main__.py", line 43, in main
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "/tmp/onefile_1690_1722281965_555845/meta_package_manager/cli.py", line 79, in <module meta_package_manager.cli>
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "/tmp/onefile_1690_1722281965_555845/meta_package_manager/sbom.py", line 52, in <module meta_package_manager.sbom>
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "/tmp/onefile_1690_1722281965_555845/spdx_tools/spdx/writer/json/json_writer.py", line 10, in <module spdx_tools.spdx.writer.json.json_writer>
[...]

PR #3037, opened 2024-08-01
Typo in `license_expression` package ID

kdeldycke:
# What does this PR do?

This fix the bad ID for the `license-expression` package introduced in #3033.

I'd like to apologise, I mixed up my local copies of Nuitka fork and venv, so I pushed my previous #3033 PR with the wrong package ID. 🙇

I triple-checked this PR really fix the issue with `license-expression` package.

# PR Checklist

### DIFF
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -2140,7 +2140,7 @@
         prefixes:
           - 'libusb'
 
-- module-name: 'license-expression' # checksum: 810a705a
+- module-name: 'license_expression' # checksum: 810a705a
   data-files:
     dirs:
       - 'data'


kayhayen (approved):

kayhayen:
Oops, we should have caught that. The module names have no schema check at all.

kayhayen:
Adding it reveals another issue:

> Nuitka-Tools: Config for module 'license-expression' is using a package name rather than a module name.
> Nuitka-Tools: Config for module 'mozilla-ca' is using a package name rather than a module name.

So that would have been good during your development.

kayhayen (approved):

kdeldycke:
Thanks again @kayhayen for the fast merge!

kayhayen:
Going out in the next hotfix as well.

kayhayen:
Turns out mozilla-ca is a valid name, but I allow listed this one, and will error out in the future. I am also changing data files to be lists of items like the other types of values already are with that hotfix, so we avoid merge issues with hotfixes that add data files.

PR #3040, opened 2024-08-01
Standalone: removed numba references from shap

interseba5:
# What does this PR do?
This pr removes all the numba references in the latest shap package version

Newer versions of shap uses njit decorator instead of @jit. Updating the yaml to remove this new references when 'not use_numba' is specified


kayhayen:
Seems like a logical extension. I will include it in the next hotfix and resolve the whitespace error myself tomorrow.

kayhayen:
I rebased this to current develop and applied `./bin/autoformat-nuitka-source --yaml` but I recommend you install the git commit hook, as it does that at commit time automatically.

PR #3045, opened 2024-08-05
fix: tolerate decoding error in ccache logs

TargetLocked:
Thank your for contributing to Nuitka!

!! Please check that you select the **develop branch** (details see below link) !!

Before submitting a PR, please review the guidelines:
[Contributing Guidelines](https://github.com/Nuitka/Nuitka/blob/develop/CONTRIBUTING.md)

# What does this PR do?

This PR plans to address the encoding issues in ccache logs.

I have done some simple tests locally and found ccache does not treat all encoding consistently in their logs (probably because of some unspecified behaviors in `std::filesystem::path::string` but I didn't investigate it further). My tests also find that all command and result lines (with which we are only concerned if I understand the code correctly) are encoded with local encoding.

Given the inconsistent in the log, I suggest we decode the log using the local encoding (which effectively reverts 0a55ff85dbead8519982da32a0f7cd540be0f575) and just ignore errors in the irrelevant parts.

# Why was it initiated? Any relevant Issues?

This should fix #3039.

# PR Checklist

- [x] Correct base branch selected? Should be `develop` branch.
- [x] Enabled commit hook or executed `./bin/autoformat-nuitka-source`.
- [x] All tests still pass. Check the Developer Manual about `Running the Tests`. There are GitHub
  Actions tests that cover the most important things however, and you are welcome to rely on those,
  but they might not cover enough.
- [x] Ideally new features or fixed regressions ought to be covered via new tests.
- [x] Ideally new or changed features have documentation updates.

PR #3046, opened 2024-08-05
fix: correct the lookup of `command`

TargetLocked:

# What does this PR do?

This PR corrects the lookup behavior of `command`.

When working with #3045 , I find the code here looks a bit strange. The values in `data` here should be results rather than commands, so it makes no sense for me to look for `command` in `data`.

I have not checked logic outside this function, so my view might be wrong.

No issues related.

# PR Checklist

### DIFF
--- a/nuitka/build/SconsCaching.py
+++ b/nuitka/build/SconsCaching.py
@@ -292,7 +292,7 @@ def _getCcacheStatistics(ccache_logfile):
                 result = result.strip()
 
                 try:
-                    command = data[commands[pid]]
+                    command = commands[pid]
                 except KeyError:
                     # It seems writing to the file can be lossy, so we can have results for
                     # unknown commands, but we don't use the command yet anyway, so just
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -2140,7 +2140,7 @@
         prefixes:
           - 'libusb'
 
-- module-name: 'license-expression' # checksum: 810a705a
+- module-name: 'license_expression' # checksum: 810a705a
   data-files:
     dirs:
       - 'data'


kayhayen:
Can you explain some more. Trying it out, just gives me incorrect result `Nuitka-Scons: Compiled 0 C files using ccache.` why you want to do this one?

TargetLocked:
> Can you explain some more. Trying it out, just gives me incorrect result `Nuitka-Scons: Compiled 0 C files using ccache.` why you want to do this one?

This is to me more of a slightly complicated typo rather than a bug fix. The code here means to match `command` and `result` by `pid`. `commands` should map `pid` to `command` while `data` should map `command` to `result`. So `data[commands[pid]]` is a `result` but assigned to `command`, which mixed them up. The reason it worked is the fallback branch `expect KeyError`, making `command` just a unique key of the `result` line, kind of like a hash. Fortunately, we don't use the value of `command` anywhere else, so this is covered up.

The regression is not expected since I thought it would be a simple fix. My apologies for not testing before submitting this PR, I will look into it later. There might be another issue in the related parts.

TargetLocked:
> The reason it worked is the fallback branch `expect KeyError`, making `command` just a unique key of the `result` line, kind of like a hash.

To check this, you may want to try to add some print.

```python
                try:
                    command = data[commands[pid]]
                except KeyError:
                    # It seems writing to the file can be lossy, so we can have results for
                    # unknown commands, but we don't use the command yet anyway, so just
                    # be unique.
                    command = "unknown command leading to " + line
                    scons_logger.warning("fallback")
                else:
                    scons_logger.warning("no fallback")
```

My results show that all lookups fall back.

TargetLocked:
Anyway, the original version works well and is expected to work well as long as we are not going to do something with the value of `command` in the future. If you think such a fix is not necessary, feel free to close this. 😉

PR #3049, opened 2024-08-06
Playwright support

### LINKED ISSUE #1029: Support for Playwright

KRRT7:
fixes #1029

### DIFF
--- a/nuitka/plugins/standard/PlaywrightPlugin.py
+++ b/nuitka/plugins/standard/PlaywrightPlugin.py
@@ -6,9 +6,10 @@
 spell-checker: ignore Playwright
 """
 
+import os
+
 from nuitka.Options import isStandaloneMode
 from nuitka.plugins.PluginBase import NuitkaPluginBase
-import os
 from nuitka.utils.Execution import NuitkaCalledProcessError
 
 
@@ -70,7 +71,7 @@ class NuitkaPluginPlaywright(NuitkaPluginBase):
         elif env_defined == "0":
             return result
         elif env_defined:
-            result = os.path(env_defined)
+            result = os.path.normpath(env_defined)
         else:
             cache_directory = ""
             if os.name == "posix":
@@ -112,29 +113,19 @@ class NuitkaPluginPlaywright(NuitkaPluginBase):
             return
 
         if not self.include_browsers:
-            self.warning(
-                "No browsers included. Use the option '--playwright-include-browser=browser_name' to include one. Use 'all' to include all installed ones."
+
+            self.sysexit(
+                "No browsers included. Use the option '--playwright-include-browser=browser_name' to include one. Use 'all' to include all installed ones."  # pylint: disable=C0301
             )
 
         self.get_Installed_Playwright_Browsers()
 
         if not self.installed_browsers:
-            self.warning(
+            self.sysexit(
                 "Error, no browsers found in the registry, if you're using playwright, make sure to install a browser."
             )
 
-        for browser in self.include_browsers:
-            if browser not in self.installed_browsers:
-
-                msg = (
-                    "Error, requested to include browser '%s' that was not found, the list of installed ones is '%s'."
-                    % (
-                        browser,
-                        ", ".join(self.installed_browsers),
-                    )
-                )
-                self.sysexit(msg)
-
+        self.info("Including browsers: %s" % ", ".join(self.include_browsers))
         if "all" in self.include_browsers:
             self.include_browsers = self.installed_browsers
         elif "ffmpeg" not in self.include_browsers and any(
@@ -149,6 +140,16 @@ class NuitkaPluginPlaywright(NuitkaPluginBase):
                     break
 
         for browser in self.include_browsers:
+            if browser not in self.installed_browsers:
+                msg = (
+                    "Error, requested to include browser '%s' that was not found, the list of installed ones is '%s'."
+                    % (
+                        browser,
+                        ", ".join(self.installed_browsers),
+                    )
+                )
+                self.sysexit(msg)
+
             msg = "Including '%s' from '%s'." % (
                 browser,
                 self.installed_browsers[browser].path,
--- a/nuitka/plugins/standard/PlaywrightPlugin.py
+++ b/nuitka/plugins/standard/PlaywrightPlugin.py
@@ -92,11 +92,13 @@ class NuitkaPluginPlaywright(NuitkaPluginBase):
 
         return result
 
-    def get_Installed_Playwright_Browsers(self):
+    def getInstalledPlaywrightBrowsers(self):
         registry_directory = self._getRegistryDirectory()
         if not os.path.exists(registry_directory):
             return
 
+        # TODO: Seems useless use of os.scandir, our helper for listing a directory
+        # would be better to use.
         browsers_installed = [
             browser
             for browser in os.scandir(registry_directory)
@@ -117,7 +119,7 @@ class NuitkaPluginPlaywright(NuitkaPluginBase):
                 "No browsers included. Use the option '--playwright-include-browser=browser_name' to include one. Use 'all' to include all installed ones."  # pylint: disable=C0301
             )
 
-        self.get_Installed_Playwright_Browsers()
+        self.getInstalledPlaywrightBrowsers()
 
         if not self.installed_browsers:
             self.sysexit(
--- a/nuitka/plugins/standard/PlaywrightPlugin.py
+++ b/nuitka/plugins/standard/PlaywrightPlugin.py
@@ -8,6 +8,7 @@ spell-checker: ignore Playwright
 from nuitka.Options import isStandaloneMode
 from nuitka.plugins.PluginBase import NuitkaPluginBase
 import os
+from nuitka.utils.Execution import NuitkaCalledProcessError
 
 
 class NuitkaPluginPlaywright(NuitkaPluginBase):
@@ -17,7 +18,7 @@ class NuitkaPluginPlaywright(NuitkaPluginBase):
     plugin_desc = "Required by 'playwright' package."
 
     def __init__(self, include_browsers):
-        self.include_browsers = include_browsers
+        self.include_browsers = list(include_browsers)
         self.installed_browsers = self.get_Installed_Playwright_Browsers()
 
         if "all" in self.include_browsers:
@@ -59,12 +60,32 @@ class NuitkaPluginPlaywright(NuitkaPluginBase):
             """,
         )
 
+    def _getPlaywrightInfo(self):
+        "Determine the path of the playwright module."
+        try:
+            info = self.queryRuntimeInformationMultiple(
+                info_name="playwright_info",
+                setup_codes="""
+from playwright import __file__ as playwright_file
+from os.path import dirname
+""",
+                values=(("playwright_file", "dirname(playwright_file)"),),
+            )
+        except NuitkaCalledProcessError as e:
+            self.debug("Exception during detection: %r" % e)
+            raise
+
+        if info is None:
+            self.sysexit("Error, it seems 'playwright' is not installed or broken.")
+
+        return info
+
     def getRegistryDirectory(self):
-        import playwright
 
         env_defined = os.environ.get("PLAYWRIGHT_BROWSERS_PATH")
         path_home = os.path.expanduser("~")
-        playwright_module_path = os.path.dirname(playwright.__file__)
+        playwright_module_path = self._getPlaywrightInfo().playwright_file
+
         result = os.path.join(
             playwright_module_path, "driver", "package", ".local-browsers"
         )
--- a/nuitka/plugins/standard/PlaywrightPlugin.py
+++ b/nuitka/plugins/standard/PlaywrightPlugin.py
@@ -1,4 +1,5 @@
 #     Copyright 2024, Kay Hayen, mailto:kay.hayen@gmail.com find license text at end of file
+#     Copyright 2024, Kevin Rodriguez, mailto:turcioskevinr@gmail.com find license text at end of file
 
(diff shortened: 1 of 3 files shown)


nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+
+class NuitkaPluginPlaywright(NuitkaPluginBase):
+    """This class represents the main logic of the plugin."""
+
+    plugin_name = "Playwright"
+    plugin_desc = "Required by 'playwright' package."
+
+    def __init__(self, include_browsers):
+        self.include_browsers = list(include_browsers)
+        self.installed_browsers = self.get_Installed_Playwright_Browsers()
```
kayhayen, line 22: Always enabled plugins cannot run code during init. I think you have to treat empty "include_browsers" special and NOT check for installed browsers. Pretty sure that should potentially crash when playwright is not installed.

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+        except NuitkaCalledProcessError as e:
+            self.debug("Exception during detection: %r" % e)
+            raise
+
+        if info is None:
+            self.sysexit("Error, it seems 'playwright' is not installed or broken.")
+
+        return info
+
+    def getRegistryDirectory(self):
```
kayhayen, line 83: Leading `_` would be good here too for the method name.

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+            """,
+        )
+
+    def _getPlaywrightInfo(self):
+        "Determine the path of the playwright module."
+        try:
+            info = self.queryRuntimeInformationMultiple(
+                info_name="playwright_info",
+                setup_codes="""
+from playwright import __file__ as playwright_file
```
kayhayen, line 69: Use `self.locateModule` instead of you only want a module location at compile time.

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+
+        result = os.path.join(
+            playwright_module_path, "driver", "package", ".local-browsers"
+        )
+        if os.path.exists(result) and next(os.scandir(result), False):
+            return result
+        elif env_defined == "0":
+            return result
+        elif env_defined:
+            result = os.path(env_defined)
```
kayhayen, line 97: Mixing branches that return and branches that do not is very bad style iny my book.

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+    def getRegistryDirectory(self):
+
+        env_defined = os.environ.get("PLAYWRIGHT_BROWSERS_PATH")
+        path_home = os.path.expanduser("~")
+        playwright_module_path = self._getPlaywrightInfo().playwright_file
+
+        result = os.path.join(
+            playwright_module_path, "driver", "package", ".local-browsers"
+        )
+        if os.path.exists(result) and next(os.scandir(result), False):
```
kayhayen, line 92: I think we should add a `isNonEmptydirectory` in `nuitka.utils.FileOperation` if that is what this code wants to test. The use of `False` as a `next` value is very creative.

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+            return result
+        elif env_defined == "0":
+            return result
+        elif env_defined:
+            result = os.path(env_defined)
+        else:
+            cache_directory = ""
```
kayhayen, line 99: Somewhere a comment that we cannot query this from playwright should be present, with it being JS I think, otherwise people will always wonder that's this.

It looks as if though, this is basically appdirs, so why don't we use our on AppDirs module, which probably covers more OSes even.

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+
+        return {browser.name: browser for browser in browsers_installed}
+
+    def considerDataFiles(self, module):
+
+        if module.getFullName() != "playwright" or not self.include_browsers:
+            return
+
+        for browser in self.include_browsers:
+            self.info(
```
kayhayen, line 140: Cannot use f-strings as they become SyntaxError for older Python versions.

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+        if not os.path.exists(registry_directory):
+            return {}
+
+        browsers_installed = list(os.scandir(registry_directory))
+        for browser in browsers_installed:
+            if browser.name == ".links":
+                browsers_installed.remove(browser)
+                break
+
+        return {browser.name: browser for browser in browsers_installed}
```
kayhayen, line 132: Cannot use dictionary contactions as they become SyntaxError for older Python versions, 2.6 only in this case. Maybe use the for-loop from above to build it incrementally.

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+        return result
+
+    def get_Installed_Playwright_Browsers(self):
+
+        registry_directory = self.getRegistryDirectory()
+
+        if not os.path.exists(registry_directory):
+            return {}
+
+        browsers_installed = list(os.scandir(registry_directory))
```
kayhayen, line 126: There is a reason, `getFileList` exists, it can e.g. ignore things, and it also gives a determinstic order, not dependent on the file system which I think should be strived for always.

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+        for browser in browsers_installed:
+            if browser.name == ".links":
+                browsers_installed.remove(browser)
+                break
+
+        return {browser.name: browser for browser in browsers_installed}
+
+    def considerDataFiles(self, module):
+
+        if module.getFullName() != "playwright" or not self.include_browsers:
```
kayhayen, line 136: Plugins really should register for module names, and they should be able to like disable themselves, if during init they find there won't be anything to do for me, like if no browsers are to be included. That is not on your plate though.

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
+#     Copyright 2024, Kay Hayen, mailto:kay.hayen@gmail.com find license text at end of file
```
kayhayen, line 1: You are welcome to put your name there

nuitka/plugins/standard/PlaywrightPlugin.py
```diff
@@ -0,0 +1,168 @@
...
+                    self.include_browsers.append(browser)
+                    self.info(
+                        "Including 'ffmpeg' for chromium-based browser. It is required by playwright."
+                    )
+                    break
+
+        for browser in self.include_browsers:
+            if browser not in self.installed_browsers:
+                self.sysexit(
+                    f"Error, requested to include browser '{browser}' that was not found, the list of installed ones is '{', '.join(self.installed_browsers.keys())}'."
```
kayhayen, line 40: Don't use `.keys` on dictionaries the dictionary by default iterates over them.

nuitka/plugins/standard/PlaywrightPlugin.py
KRRT7: yes, it aims to test that the result directory isn't empty

nuitka/plugins/standard/PlaywrightPlugin.py
KRRT7: we want to cover the OS'es playwright supports, using appdirs and supporting OS'es that playwright itself doesn't feels finicky

nuitka/plugins/standard/PlaywrightPlugin.py
KRRT7: i'm relying on os.scandir's Path-Like result's there

nuitka/plugins/standard/PlaywrightPlugin.py
KRRT7: this is a playwright specific quirk, if the env variable is 0 then we should look within the site-modules/playwright directory for the browsers, if not, then it is a path they want us to look into for the browsers i.e home directory or anywhere where the user wants

kayhayen:
As a I side note, ideally I resolve them comments, when I think the question asked has an answer that needs no answer. Otherwise this is useless to do, since email notifications would be the only thing informing me of there even being an answer, and even if I think it's not a good one, no way I can immediately react at all times, but email becomes unread.

kayhayen:
I don't quite like code using plugins. I think it would be better if we allowed plugins to provide variable values (on demand) and have configuration instead. For something like `_getPlaywrightRegistryDirectory` maybe we don't have to have it all as Python, but somehow provide it. But for now it's good enough. I am sometimes on a mission to reduce the number of plugins, and this will be on my mind surely as one that I believe ought not to be one.

PR #3064, opened 2024-08-13
Improve getFileContents by using pathlib instead of os

RMPR:

# What does this PR do?

Improve getFileContents by using pathlib instead of os

On Windows 11 under some conditions os.read can fail not finding the
path with an error about Unicode encoding. This does not fail with
pathlib. Additionally, since Python 3 'U' is the default behavior and
since 3.10 has no longer any effect, and is considered deprecated.

# PR Checklist

### DIFF
--- a/nuitka/importing/PreloadedPackages.py
+++ b/nuitka/importing/PreloadedPackages.py
@@ -116,7 +116,7 @@ def detectPthImportedPackages():
         for path, filename in listDir(prefix):
             if filename.endswith(".pth"):
                 try:
-                    for line in getFileContentByLine(path, "rU"):
+                    for line in getFileContentByLine(path, "r"):
                         if line.startswith("import "):
                             if ";" in line:
                                 line = line[: line.find(";")]
--- a/nuitka/utils/FileOperations.py
+++ b/nuitka/utils/FileOperations.py
@@ -14,6 +14,7 @@ import errno
 import fnmatch
 import glob
 import os
+import pathlib
 import pickle
 import shutil
 import stat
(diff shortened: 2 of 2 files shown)


KRRT7:
Nuitka officially supports Python2, so we need to keep backwards compatibility with it

kayhayen:
Also is it just me, or doesn't the code change look terribly more complex, by handling the open locally, and limiting what modes there are.

RMPR:
> Also is it just me, or doesn't the code change look terribly more complex, by handling the open locally,

Are you referring to using 2 open instead of one? I could have factored out the open and used pathlib on the file pointer instead no problem.

> and limiting what modes there are.

The function docstring states that the two modes are `r` and `rb` making the code explicitly fail when that contract is breached is a good practice IMHO.

kayhayen:
The function's job is not to enforce mode limitations on other code. If you think that's a good practice, raising a `RuntimeError` rather than `AssertionError` or `TypeError` or `ValueError`, that's not it.

Also, the universal line buffer mode is needed for compatibility with older Python and should, of course, not be removed just because it doesn't fit your idea of limited modes allowed.

As for the open functionality, we already got `openTextFile`, which dispatches between `codecs.open` and regular `open`, and just generally handles portability for us. I don't see `pathlib` contributing anything there that those cannot do, except maybe combining opening and reading into one thing, which honestly, doesn't get easier than this

```
        with openTextFile(filename, mode, encoding=encoding) as f:
            return f.read()
```

I feel pathlib might be a nice interface, but I don't think it's helpful with everything or that it helps portability across Python versions to use it at all. Until `codecs.open` get removed, there is nothing that speaks against using it.

PR #3066, opened 2024-08-14
Update Yaml.py to allow merging implicit-imports

jimkring:
Hi @kayhayen 

# What does this PR do?

This adds support for a custom package-configuration.yml to merge it's `implicit-imports` for a package with the `implicit-imports` for a package that's defined in the standard configuration here: `nuitka\plugins\standard\standard.nuitka-package.config.yml`

The current behavior is that if a package is already declared in the standard.nuitka-package.config.yml then trying to add it in a custom package-configuration.yml (that's referenced during a build) will raise and assertion.  However, the desired behavior (for my use case) is to simply append the dependency to the existing `implicit-imports`.

# PR Checklist

### DIFF
--- a/nuitka/utils/Yaml.py
+++ b/nuitka/utils/Yaml.py
@@ -105,9 +105,21 @@ Error, empty (or malformed?) user package configuration '%s' used."""
         # TODO: Full blown merging, including respecting an overload flag, where a config
         # replaces another one entirely, for now we expect to not overlap.
         for key, value in other.items():
-            assert key not in self.data, key
-
-            self.data[key] = value
+            # assert key not in self.data, key
+            if key in self.data:
+                new_implicit_imports = value.get('implicit-imports', None)
+                if new_implicit_imports:
+                    value.pop('implicit-imports')
+                    if self.data[key].get('implicit-imports', None) is None:
+                        self.data[key]['implicit-imports'] = new_implicit_imports
+                    else:
+                        self.data[key]['implicit-imports'].extend(new_implicit_imports)
+                if len(value) > 0:
+                    general.sysexit(
+                        f"Error, duplicate config for module name '{key}' encountered in '{self.name}'."
+                    )
+                else:
+                    general.info(f"merged implicit-imports for '{key}'.")
 
 
 def getYamlPackage():


kayhayen:
Very good, I think we can work off that. I think we should also allow (in the schema ideally) to specify the strategy for merging values, like you could say `merge: replace|extend|disallow` in your user values, and then it would adhere that, with that value being on the top level for the module, for the section, and for entries in the list.

kayhayen:
I rebased this PR to see what to do with it. My plans for working with @KRRT7 on this, didn't come to pass. I will do this one on my next PR day this month, and check in with what @KRRT7 has there too, making it a best of.

I think, I want a per-config element definition, if the current config is to replace or extend existing one, so that you can disable implicit-imports as well as adding more of them.

Also, what I want is like a merge tool, rather than doing this on the fly. In that world, several files are read and extended, and then it's capable of what systemctl cat can do. It can produce this kind of output:

```
$ systemctl cat httpd 
# /usr/lib/systemd/system/httpd.service     <--highlighted
....
[Unit]
Description=The Apache HTTP Server
...
[Service]
...
PrivateTmp=true
...

# /etc/systemd/system/httpd.service.d/override.conf  <--highlighted
[Service]
PrivateTmp=false
UMask=0002
```

So those comments, show where things are coming from. I really like that idea of being able to say `python -m nuitka --show-package-config=numpy` and it then producing something similar, ideally with line numbers, but I don't think, we can have those though, due to limitations of the cheap parser. Once Standalone Backend Python becomes a thing, we can have more powerful yaml tooling without impacting user compiles, then we can do that too.

I know, it's a bit of overengineering, but I hate it when I don't know the effective result, and cannot tell where it's coming from, and we could also record that in the build result, report, or whatever for reference. Not sure how hard that's going to be, but if we have that tooling, we can do the merge to file contents and read that instead.

kayhayen:
It seems the current code potentially is not correct, at least CI files on a case that uses config to include data files like this:

```
- module-name: 'data_files_package' # checksum: 5ee34e01
  data-files:
    - patterns:
        - '**/*.txt'
```

No implicit import, it seems this config got dropped apparently.

KRRT7:
My implementation is pretty close to this one I believe, good stuff

kayhayen:
I think, writing out config is crucial for testing this kind of thing. Like merge our config files into themselves and see no change, would be a prudent test. As with the current issue, it's otherwise way too difficult to know what's the outcome.

kayhayen:
I messed this up, won't happen so soon, was too ill when this was planned and didn't have the time since.

kayhayen (approved):
I am merging this, because while I got a lot one recently, I did not manage to come up with a more universal approach, and this will help some cases for now at least

PR #3129, opened 2024-09-25
Onefile: Preserve the original argv[0] in __compiled__.onefile_argv0

xen0n:
# What does this PR do?

Record the `argv[0]` in the onefile bootstrapper and expose it via `__compiled__.onefile_argv0`, to not regress advanced CLI use cases where the program needs this info to differentiate behavior.

In commit 9b0406b2a ("Fix, need to make sure `sys.argv[0]` is absolute for best usability"), the original `argv[0]` as received by the onefile binary is thrown away, breaking applications needing this information.

To support these cases, record the original `argv[0]` in the `NUITKA_ONEFILE_ARGV0` environment variable, for exposure via `__compiled__.onefile_argv0`. The field is made `None` if the onefile mode is not in use, or if the binary is not launched via the onefile bootstrapper.

# PR Checklist

## Summary by Sourcery

Preserve the original argv[0] in onefile mode by recording it in the NUITKA_ONEFILE_ARGV0 environment variable and exposing it through __compiled__.onefile_argv0, ensuring compatibility with advanced CLI use cases that require this information.

New Features:
- Introduce a new feature to record the original argv[0] in the onefile bootstrapper and expose it via __compiled__.onefile_argv0 for advanced CLI use cases.

Enhancements:
- Add a mechanism to store the original argv[0] in the NUITKA_ONEFILE_ARGV0 environment variable and retrieve it using the getOnefileArgv0Object function.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1404,6 +1404,31 @@ which you expect to be inside the onefile binary, access them like this.
    except NameError:
       open(os.path.join(os.path.dirname(sys.argv[0]), "user-provided-file.txt"))
 
+Note that when the program is launched from the onefile executable, the
+original ``sys.argv[0]`` from the invocation command line will not be
+preserved. For advanced use cases where one needs access to the original
+``sys.argv[0]``, it may be found at ``__compiled__.onefile_argv0``. The
+field will read back as ``None`` if the program is not launched from the
+onefile executable, thus not having gone through the onefile bootstrap
+stage; the original ``sys.argv[0]`` would be preserved as well in this
+case.
+
+.. code:: python
+
+   # Suppose the onefile binary is placed at /opt/abc/bin/foo, and it was
+   # symlinked to /usr/local/bin/bar, and invoked as `bar ...`:
+   assert sys.argv[0] == "/usr/local/bin/bar"
+   assert __compiled__.onefile_argv0 == "bar"
+
+   # If the onefile tempdir is overridden and the program is invoked
+   # directly from the unpacked location, sys.argv[0] would not be touched.
+   #
+   # Suppose the onefile tempdir is /home/xx/.cache/abc/0.1.2, and the
+   # foo.bin executable inside is symlinked to /usr/local/bin/baz, and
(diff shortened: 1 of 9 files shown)


kayhayen:
So, this looks great to me. I think it might not be all that specific to onefile, but probably could be considered `original_argv`, I have to checkout again, if we don't also do it for standalone mode, and maybe even accelerated mode, in which case that naming and applying might make a lot more sense. I can definitely take it from there. Good on you for extension the documentation, that's great.

kayhayen:
So, I did a bunch of refactoring, and now it works on Linux, I am a bit worried about the validity of the environment string after deleting it, probably needs a strdup, and I kind of feel that mixing filename and environment strings APIs is sort of doomed to not work in some configuration, but we shall see.

kayhayen:
So, can you try it out, and tell me if it still works for you? I renamed the attribute to `original_argv0` so you need to consider that of course. 

It works for accelerated and standalone mode too now, and also for cases where the unpacked folder of the onefile is called directly. @xen0n 

I would then merge this for 2.5 release if you confirm, or at least don't deny it working.

kayhayen:
Seems, I made a mistake for Windows and the unicode path conversion to object helper, I fixed that and rebased to latest pre-release.

xen0n:
Sorry for the late reply; I only had time to test this today (mostly traveling for the past 2 weeks).

After rebasing the branch onto the latest `develop` I can see the onefile binary working but the unpacked `*.bin` program segfaults. Not sure if I've done the rebasing incorrectly (I've simply skipped those earlier commits that seem to have already landed), but it'll be tomorrow before I can start debugging anyway.

kayhayen:
Seems, I need to do another rebase, GitHub Actions found stuff there, maybe there issues with develop.

kayhayen:
I tried Linux after rebase with no issue just now after a rebase, lets see what the actions builds say. If they work, it cannot be too bad I feel. That said, of course nothing tests the attribute value, but I think I saw it in earlier tests OK.

Issue #2038, opened 2023-02-02
Feature: Allow starting a windows executable with minimized console


deajan:
Hello,

I've lately designed a [backup program](https://github.com/netinvent/npbackup) which works both with a CLI and a GUI, working on both linux and windows.
The idea is to have a GUI when running the program without parameters, and use the CLI in order to achieve scheduled tasks etc.

When run on Windows without parameters, a command line window shows, then the GUI appears.
I have a "user" problem, being that some users just click into the command line window, which halts GUI execution.
I've seen a couple of users complaining about this so far.

Would there be perhaps a way to run the command line window minimized, via let's say a `--windows-minimize-console` switch or so ?
That would totally solve my "user related" issues.
Thanks.

PS: Willing to sponsor a bit on this

* Nuitka version, full Python version, flavor, OS, etc. as output by *this*
  command.

  > python -m nuitka --version

```
1.4.2
Commercial: 2.0.0
Python: 3.10.7 (tags/v3.10.7:6cc6b13, Sep  5 2022, 14:08:36) [MSC v.1933 64 bit (AMD64)]
Flavor: CPython Official
Executable: c:\Python310-64\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 10
Version C compiler: ~\AppData\Local\Nuitka\Nuitka\Cache\downloads\gcc\x86_64\11.3.0-14.0.3-10.0.0-msvcrt-r3\mingw64\bin\gcc.exe (gcc).
```
* How did you install Nuitka and Python

```
git clone https://github.com/nuitka/nuitka-commercial
python -m pip install c:\path\to\nuitka-commercial
```

* Also supply a [Short, Self Contained, Correct, Example](https://sscce.org/)

Run the following program without arguments, click into the console window, then the GUI freezes until someone types a key into the console window (the actual reason being that windows considers console clicks as select operation for copy/pasting text).

```python
from argparse import ArgumentParser
import PySimpleGUI as sg

def main_cli():
    print("I am running a CLI")

def main_gui():
    sg.Popup("Now the console window is open, users click inside, and hence will block GUI program")

if __name__ == "__main__":
    parser = ArgumentParser(prog="Some mixed gui and cli program")


    parser.add_argument(
        "--check", action="store_true", help="Do something fancy"
    )

    args = parser.parse_args()

    if args.check:
        main_cli()
    else:
        main_gui()
```

* Provide in your issue the Nuitka options used

```
[...]

PR #3182, opened 2024-11-04
add new choice 'hide' to --windows-console-mode flag

fallchildren2:
As discussed in #2038

~~Still a WIP for now as I have to test it beforehand.~~

## Summary by Sourcery

Introduce a new 'hide' option to the --windows-console-mode flag, enabling the console window to be hidden when spawned. Update the console handling logic to accommodate this new option, ensuring proper behavior on Windows systems.

New Features:
- Add a new 'hide' option to the --windows-console-mode flag, allowing the console window to be hidden if it is newly spawned.

Enhancements:
- Modify the console handling logic to support the new 'hide' option, ensuring that the console window is hidden when appropriate.

## Summary by Sourcery

Introduce a 'hide' option to the --windows-console-mode flag to allow hiding of newly spawned console windows on Windows systems. Update the console handling logic to accommodate this new option.

New Features:
- Add a new 'hide' option to the --windows-console-mode flag, allowing the console window to be hidden if it is newly spawned.

Enhancements:
- Modify the console handling logic to support the new 'hide' option, ensuring that the console window is hidden when appropriate.

### DIFF
--- a/nuitka/OptionParsing.py
+++ b/nuitka/OptionParsing.py
@@ -1390,14 +1390,16 @@ windows_group.add_option(
     "--windows-console-mode",
     action="store",
     dest="console_mode",
-    choices=("force", "disable", "attach"),
+    choices=("force", "disable", "attach", "hide"),
     metavar="CONSOLE_MODE",
     default=None,
     help="""\
 Select console mode to use. Default mode is 'force' and creates a
 console window unless the program was started from one. With 'disable'
 it doesn't create or use a console at all. With 'attach' an existing
-console will be used for outputs. Default is 'force'.
+console will be used for outputs. With 'hide' a newly spawned console
+will be hidden and an already existing console will behave like
+'force'. Default is 'force'.
 """,
 )
 
--- a/nuitka/build/Backend.scons
+++ b/nuitka/build/Backend.scons
@@ -562,6 +562,9 @@ if "linux" in sys.platform:
 if not env.msvc_mode:
     env.Append(LIBS=["m"])
 
+if console_mode == "hide":
+    env.Append(LIBS=["User32"])
+
 if python_debug:
     env.Append(CPPDEFINES=["Py_DEBUG"])
 
--- a/nuitka/build/Backend.scons
+++ b/nuitka/build/Backend.scons
@@ -562,9 +562,6 @@ if "linux" in sys.platform:
 if not env.msvc_mode:
     env.Append(LIBS=["m"])
 
-if console_mode == "hide":
-    env.Append(LIBS=["User32"])
-
 if python_debug:
     env.Append(CPPDEFINES=["Py_DEBUG"])
 
--- a/nuitka/build/SconsCompilerSettings.py
+++ b/nuitka/build/SconsCompilerSettings.py
@@ -787,7 +787,7 @@ def setupCCompiler(env, lto_mode, pgo_mode, job_count, onefile_compile):
 
     # Even if console is forced, for Win32 it means to specify Windows
     # subsystem, we can still attach or create.
-    if env.console_mode != "force":
+    if env.console_mode not in ["force", "hide"]:
         if env.mingw_mode:
             env.Append(LINKFLAGS=["-Wl,--subsystem,windows"])
             env.Append(CPPDEFINES=["_NUITKA_WINMAIN_ENTRY_POINT"])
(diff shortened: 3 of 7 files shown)


fallchildren2:
Works fine for me in my tests:

https://gitlab.com/modding-openmw/umo/-/commit/7ab90b17f6e26411e260b3e2e78a33ad0bcb1b2f

fallchildren2:
Or not... the User32 lib is not linked somehow...

fallchildren2:
Now it works with adding the User32 lib when console_mode is "hide". Still not too sure though about the scons stuff. No experience here.

kayhayen:
You did a fantastic job here, it's a shame I am releasing 2.5 without it, but from the looks of it, I will work on it a bit for minor cleanups and probably merge it for a hotfix, seems very complete and low risk to me.

fallchildren2:
Thanks for the kind words. Greeting from Hannover!

kayhayen:
Close to my emotional home, Hamburg.

nuitka/OptionParsing.py
```diff
@@ -1390,14 +1390,16 @@
     "--windows-console-mode",
     action="store",
     dest="console_mode",
-    choices=("force", "disable", "attach"),
+    choices=("force", "disable", "attach", "hide"),
```
kayhayen, line 1393: In a moment of bike shedding, I would like the name it `minimize_after_start` or such, or anything that conveys it better that it cannot start hidden.

nuitka/OptionParsing.py
```diff
@@ -1390,14 +1390,16 @@
...
     metavar="CONSOLE_MODE",
     default=None,
     help="""\
 Select console mode to use. Default mode is 'force' and creates a
 console window unless the program was started from one. With 'disable'
 it doesn't create or use a console at all. With 'attach' an existing
-console will be used for outputs. Default is 'force'.
+console will be used for outputs. With 'hide' a newly spawned console
+will be hidden and an already existing console will behave like
```
kayhayen, line 1401: Btw, I really like that you managed that.

nuitka/build/Backend.scons
```diff
@@ -562,6 +562,9 @@ if "linux" in sys.platform:
 if not env.msvc_mode:
     env.Append(LIBS=["m"])
 
+if console_mode == "hide":
```
kayhayen, line 565: I will want to move this to common code though.

nuitka/build/SconsCompilerSettings.py
```diff
@@ -787,7 +787,7 @@ def setupCCompiler(env, lto_mode, pgo_mode, job_count, onefile_compile):
 
     # Even if console is forced, for Win32 it means to specify Windows
     # subsystem, we can still attach or create.
-    if env.console_mode != "force":
+    if env.console_mode not in ["force", "hide"]:
```
kayhayen, line 790: Probably should list the modes that actually are not console, but actually want Windows subsystem instead. Also using a tuple over a list when doing this operation is something I regard cleaner.

nuitka/build/static_src/MainProgram.c
```diff
@@ -1191,6 +1191,9 @@ int __stdcall wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, wchar_t *lp
 #else
 #if defined(_WIN32)
 int wmain(int argc, wchar_t **argv) {
+#if defined(_NUITKA_HIDE_CONSOLE_WINDOW)
```
kayhayen, line 1194: The placement here is probably very agressive, but also why not, need to check though.

kayhayen: Probably might make sense to merge this with

```
    // Attach to the parent console respecting redirection only, otherwise we
    // cannot even output traces.
#if defined(_WIN32) && defined(_NUITKA_ATTACH_CONSOLE_WINDOW)
    inheritAttachedConsole();
#endif
```

kind of code and just have a `setupIneritedConsole()` helper that does these condition checks in the common code rather than per item.

kayhayen (approved):
I will take it from there for a hotfix, just keeping track of things I mean to change on top of what you did.

nuitka/OptionParsing.py
kayhayen: I will stick to `hide` for these options, actual reading will always be required. At least hide is a more obvious name than attach.

nuitka/build/static_src/MainProgram.c
kayhayen: I will probably clean this up at some point, but not for release, not really that much of an improvement right now.

PR #3186, opened 2024-11-07
Allow disabling nuitka in pyproject.toml builds.

### LINKED ISSUE #1600: Using nuitka to build wheels with setuptools.build_meta backend in pyproject.toml without changing code
I would like to use

```terminal
[build-system]
...
build-backend = "setuptools.build_meta"
```

in my Python package's `pyproject.toml`, but still be able to build wheels with nuitka analogously to the legacy `python setup.py bdist_nuitka` command. I tried using this `setuptools` backend with `build_with_nuitka=True` in the `setup.cfg`, but it seems to be ignored in PEP-517-compliant wheel-building commands such as `pip build .` (whereas `build-backend = "nuitka.distutils.Build"` does work with `pip build`).

I want to use only the `setuptools` backend because I also do "generic" builds without `nuitka`, and editing the `pyproject.toml` file in CI would screw up my automated version tracking with `setuptools_scm`.

```terminal
python -m nuitka --version
0.8.3
Commercial: None
Python: 3.9.7 (default, Sep 28 2021, 13:03:42) 
Flavor: pyenv
Executable: /home/campy/Documents/IMS/SolarPV/PVfit/venv/pvfit/bin/python
OS: Linux
Arch: x86_64
Distribution: Ubuntu (based on Debian) 18.04.6
```

mabl:
# What does this PR do?

When using `pyproject.toml` it is sometimes helpful to be able to switch between `setuptools`-based build and a Nuitka build without modifying the `pyproject.toml`.

Here, a configuration setting `no-nuitka` is added, which can be set to `true`, resulting in a bypass of nuitka during build.

This mechanism also works for the `build` package, using the `--config-setting` command line option.

Fixes #1600.

## Summary by Sourcery

New Features:
- Introduce a configuration setting 'no-nuitka' in pyproject.toml to allow bypassing Nuitka during the build process.

### DIFF
--- a/nuitka/distutils/Build.py
+++ b/nuitka/distutils/Build.py
@@ -23,16 +23,16 @@ class NuitkaBuildMetaBackend(setuptools.build_meta._BuildMetaBackend):
     def build_wheel(
         self, wheel_directory, config_settings=None, metadata_directory=None
     ):
-        # Allow falling back to setuptools when the `no-nuitka` configuration setting is set to true.
+        # Allow falling back to setuptools when the `build_with_nuitka` configuration setting is set to true.
         if config_settings:
-            no_nuitka = config_settings.pop("no-nuitka", "false").lower()
+            build_with_nuitka = config_settings.pop("build_with_nuitka", "true").lower()
 
-            if no_nuitka not in {"true", "false"}:
+            if build_with_nuitka not in ("true", "false"):
                 raise ValueError(
-                    f"When passing the  `no-nuitka` setting, it must either be `true` or `false`."
+                    "When passing the 'build_with_nuitka' setting, it must either be 'true' or 'false'."
                 )
 
-            if no_nuitka == "true":
+            if build_with_nuitka == "false":
                 return super().build_wheel(
                     wheel_directory, config_settings, metadata_directory
                 )
--- a/nuitka/distutils/Build.py
+++ b/nuitka/distutils/Build.py
@@ -23,6 +23,20 @@ class NuitkaBuildMetaBackend(setuptools.build_meta._BuildMetaBackend):
     def build_wheel(
         self, wheel_directory, config_settings=None, metadata_directory=None
     ):
+        # Allow falling back to setuptools when the `no-nuitka` configuration setting is set to true.
+        if config_settings:
+            no_nuitka = config_settings.pop("no-nuitka", "false").lower()
+
+            if no_nuitka not in {"true", "false"}:
+                raise ValueError(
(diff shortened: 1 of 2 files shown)


kayhayen:
I noticed this one relatively late only, my bad. I rebased it to current develop, after I made the release rebase for 2.5, and am going to check it out now.

nuitka/distutils/Build.py
```diff
@@ -23,6 +23,20 @@ class NuitkaBuildMetaBackend(setuptools.build_meta._BuildMetaBackend):
     def build_wheel(
         self, wheel_directory, config_settings=None, metadata_directory=None
     ):
+        # Allow falling back to setuptools when the `no-nuitka` configuration setting is set to true.
+        if config_settings:
+            no_nuitka = config_settings.pop("no-nuitka", "false").lower()
```
kayhayen, line 28: We got for `setup.cfg`, and `setup` already the naming 

```
   [metadata]
   build_with_nuitka = true
```

So I guess, I would like to stick that name. Ordinarily, you could just switch the backend back to disable it with the pyproject, but I can see how that's less obvious.

kayhayen (changes requested):
I might do the renaming myself if you don't want to do it yourself. Also we should update Nuitka-Website docs.

kayhayen:
Really liking it, would add consistency if all 3 was of doing it, supported it. The name should ideally be the same. It's a pity that the parsing has to be done manually, but I guess, that's how it goes if Nuitka doesn't enter the picture, what else could do it.

kayhayen:
I will put this into a hotfix of 2.5 easily, since it seems to have no risk.

nuitka/distutils/Build.py
mabl: So you propose to invert the logic and use `build_with_nuitka=false` as the config setting? I'm fine either way. Also, feel free to edit the MR as you see fit. I would really love to see this merged, since it would simplify our CI pipeline :-)

kayhayen: Yes, effectively that.

kayhayen:
I have a suspect, the setting ends up as a Nuitka option on the command line though, which then should be unknown, is that not true?

kayhayen:
I am ignore `build_with_nuitka` in the options parsing from TOML file, then it ought to be good to toggle it on/off without removing the config value from the file.

kayhayen:
Can you tell me if this works for you, then I can merge it.

kayhayen:
So, 2.5. is delayed due to a regression I will need to fix, taking this in right away therefore.

kayhayen (approved):

PR #3233, opened 2024-11-30
Fix: mode=app not working correct on macOS

chenmozhijin:
Fixes the issue where mode=app has no effect on macOS because options.create_app_bundle was not being used.

## Summary by Sourcery

Bug Fixes:
- Fix the issue where mode=app was not functioning correctly on macOS by ensuring the correct option is used to create an app bundle.


kayhayen (approved):

kayhayen:
This will go directly to 2.5.5 hotfix too

PR #3273, opened 2024-12-24
Fix/ignore f trace

Oblynx:

# What does this PR do?

It is part 1 of an effort to make Nuitka play well with debuggers. The scenario that we try to enable is to debug through compiled frames.

It fixes the error:

```
  File "/home/container_user/.local/share/code-server/extensions/ms-python.debugpy-2024.8.0-linux-x64/bundled/libs/debugpy/_vendored/pydevd/pydevd.py", line 2246, in set_trace_for_frame_and_parents frame.f_trace = self.trace_dispatch RuntimeError: f_trace is not writable in Nuitka
```

With this change, the debugger can step through compiled frames. At least in some cases. However, stopping at a breakpoint still doesn't work.

# PR Checklist

## Summary by Sourcery

Bug Fixes:
- Fixed a RuntimeError that prevented debuggers from setting f_trace on compiled frames, enabling stepping through compiled frames in some cases.

### DIFF
--- a/nuitka/build/static_src/CompiledFrameType.c
+++ b/nuitka/build/static_src/CompiledFrameType.c
@@ -268,11 +268,20 @@ static int _Nuitka_Frame_set_trace(PyObject *self, PyObject *value, void *data) 
     assert(Nuitka_Frame_CheckExact(self));
     CHECK_OBJECT(self);
     assert(_PyObject_GC_IS_TRACKED(self));
+#if !defined(_NUITKA_DEPLOYMENT_MODE) && !defined(_NUITKA_NO_DEPLOYMENT_FRAME_USELESS_SET_TRACE)
+    if (value == Py_None) {
+        return 0;
+    } else {
+        PyThreadState *tstate = PyThreadState_GET();
 
-    PyThreadState *tstate = PyThreadState_GET();
-
-    SET_CURRENT_EXCEPTION_TYPE0_STR(tstate, PyExc_RuntimeError, "f_trace is not writable in Nuitka");
-    return -1;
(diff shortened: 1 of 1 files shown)


kayhayen (approved):
I will take this, but will want to hide it behind a non-deployment handler, that we will give a warning if this is happening. Silently breaking user expectations is not fun for those trying to figure out, that this promise to call back is not kept.

kayhayen:
In preparation of merging this, I solved the warning about the return value (not an object, but int, should be 0 for "ok" I think) and rebased to current develop.

kayhayen:
So, I added support for `--no-deployment-flag=frame-useless-set-trace`. Also, if the value provided is `None`, i.e., it's not a tracing function, then we just ignore that, which in some cases also helps.

This gives me e.g. this:

```
  File "/home/hayen/repos/Nuitka-develop/Mini.py", line 4, in f
    sys._getframe().f_trace = False

RuntimeError: f_trace is not writable in Nuitka, ignore with '--no-deployment-flag=frame-useless-set-trace'
```

And using that flag, the write is now ignored. I consider it import for the users to be aware, that expectations may not be met. Whatever debugpy is doing there, it will not work. However, maybe it always only ever set `None` unless enabling something, so this may be all largely overkill, it would be nice if you confirmed that. I am merging this now (once CI likes it) and will put it on staging branch for you to use.

kayhayen (approved):
Thanks for taking this on

kayhayen:
Part of the 2.6 release that I just made, it will hit commercial once that release is proven to be not all that bad.

Issue #3303, opened 2025-01-22
Multidist binaries ignore the way they are called and check only the running binary name, making them hard to invoke from a single one


gracezheng9:
I created a virtual environment via `python -m venv testvenv`. 

Activated the virtual environment via `.\testvenv\Scripts\activate`.

I installed `nuitka` via `pip install nuitka`.

```
(testvenv) PS Z:\Users\gzheng\work> pip install nuitka
Collecting nuitka
  Using cached Nuitka-2.6-cp313-cp313-win_amd64.whl
Collecting ordered-set>=4.1.0 (from nuitka)
  Using cached ordered_set-4.1.0-py3-none-any.whl.metadata (5.3 kB)
Collecting zstandard>=0.15 (from nuitka)
  Downloading zstandard-0.23.0-cp313-cp313-win_amd64.whl.metadata (3.0 kB)
Using cached ordered_set-4.1.0-py3-none-any.whl (7.6 kB)
Downloading zstandard-0.23.0-cp313-cp313-win_amd64.whl (495 kB)
Installing collected packages: zstandard, ordered-set, nuitka
Successfully installed nuitka-2.6 ordered-set-4.1.0 zstandard-0.23.0
```

Below is the version information:
```
(testvenv) PS Z:\Users\gzheng\work\testvenv> python -m nuitka --version
2.6
Commercial: None
Python: 3.13.0 (tags/v3.13.0:60403a5, Oct  7 2024, 09:38:07) [MSC v.1941 64 bit (AMD64)]
Flavor: CPython Official
GIL: yes
Executable: Z:\Users\gzheng\work\testvenv\Scripts\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 11
Version C compiler: C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.42.34433\bin\Hostx64\x64\cl.exe (cl 14.3).
```

I am trying to reproduce the simple multidist example from stackoverflow here: https://stackoverflow.com/questions/78047572/how-do-i-produce-multiple-executables-using-nuitka-for-python. But it seems that it does not matter what `args` I pass to `subprocess`, the built exe cannot find the correct entrypoint; only the `executable=` is used even though Python reports that the `arg` is indeed what I passed in. What am I doing wrong? 

If it matters, I am on a Parallels Windows 11 VM running on a Macbook pro. 

I compiled the exe as follows:
```
[...]

PR #3305, opened 2025-01-22
Multidist binaries now use __compiled__.original_argv0

gracezheng9:
so that multiple entrypoint can be invoked from a single binary

# What does this PR do?
Use `__compiled__.original_argv0` instead of `sys.argv[0]`to allow multidist binaries to invoke from the original argv0 entrypoint.

See https://github.com/Nuitka/Nuitka/issues/3303. 

# PR Checklist

## Summary by Sourcery

Bug Fixes:
- Fix an issue where multidist binaries could not invoke the correct entrypoint when called with different names.

### DIFF
--- a/nuitka/tree/ReformulationMultidist.py
+++ b/nuitka/tree/ReformulationMultidist.py
@@ -36,7 +36,7 @@ def createMultidistMainSourceCode():
     source_code = renderTemplateFromString(
         r"""
 import sys, re, os
-main_basename = re.sub(r'(.pyw?|\.exe|\.bin)?$', '', os.path.normcase(os.path.basename(sys.argv[0])))
+main_basename = re.sub(r'(.pyw?|\.exe|\.bin)?$', '', os.path.normcase(os.path.basename(__compiled__.original_argv0)))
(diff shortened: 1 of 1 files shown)


kayhayen (approved):
Thanks for doing this, good job.

kayhayen:
I will take it from here, seems codespell got an update making it fail, although I feel that one is already resolved, probably because I need to rebase this one to 2.6 which I had forgotten to put out on develop too.

kayhayen:
So, that is actually wrong, merging it for the hotfix now, it will have the spelling fixes needed once merged to develop, we will be sane again

kayhayen:
Lovely, we don't have a real distutils test case unfortunately, which is really weak of Nuitka, but I guess, steams from the fact, that we don't have it easy to produce runners. On Windows, anything not a symlink is going to kind of be ugly, as Windows lacks "exec" support, otherwise I would have multidist and being used for wheel compilations already, your approach is nice. I need to check how well symlink support is now without admin rights on Windows, and maybe how to test for it during installation, that kind of thing.

PR #3332, opened 2025-02-07
Treat shared library dependencies with paths as like rpaths

### LINKED ISSUE #3325: Binaries compiled with a Python installed by `uv` fails on Ubuntu: `libpython3.13.so.1.0` not found
Tl;Dr: System Python installed by `actions/setup-python` is working but not the Python installed by `uv`.

I made up a simple project to demonstrate this issue: https://github.com/kdeldycke/nuitka-issue-3325

### Works with `actions/setup-python`

This project is a simple CLI compiled to a standalone binary on all major platforms (Ubuntu, macOS and Windows) and 2 different architectures (`x86_64` and `arm64`).

This is performed by a GitHub workflow available at: https://github.com/kdeldycke/nuitka-issue-3325/blob/main/.github/workflows/system-python.yaml

The process [works perfectly on all platforms](https://github.com/kdeldycke/nuitka-issue-3325/actions/runs/13107996796).

That's why I love Nuitka and exclusively rely on it! :D

Note that in the logs, Nuitka is [picking the right Python version from the system](https://github.com/kdeldycke/nuitka-issue-3325/actions/runs/13107996796/job/36565931945#step:7:14):

```text
Nuitka: Starting Python compilation with Nuitka '2.6.2' on Python (flavor GitHub Actions Python), '3.13' commercial grade 'not installed'.
```

### Issue with `uv`

Now let's modify the working workflow used above, and completely remove the use of `actions/setup-python`. Instead, we will rely on `uv` alone to install our target Python version, by invoking:

```shell-session
$ uv venv --python 3.13
```

That new workflow is available at: https://github.com/kdeldycke/nuitka-issue-3325/blob/main/.github/workflows/uv-python.yaml

Here is an explicit highlight of the differences between the two workflows:
```diff
$ diff -ru ./.github/workflows/system-python.yaml ./.github/workflows/uv-python.yaml        
--- ./.github/workflows/system-python.yaml	2025-02-03 10:49:36
+++ ./.github/workflows/uv-python.yaml	2025-02-03 10:49:36
@@ -6,7 +6,7 @@
 
 jobs:
 
-  system-python:
+  uv-python:
     strategy:
       matrix:
         os:
@@ -19,15 +19,12 @@
     continue-on-error: true
     steps:
       - uses: actions/checkout@v4.2.2
-      - uses: actions/setup-python@v5.4.0
-        with:
-          python-version: "3.13"
       - name: Install uv
         run: |
           python -m pip install "uv == 0.5.26"
       - name: Install Nuitka
         run: |
-          uv --no-progress venv
+          uv --no-progress venv --python 3.13
           uv --no-progress pip install "nuitka == 2.6.2"
 
       - name: uv run -m nuitka --version
```

Despite these differences, I expect no change in the results. But this is not the case as you can see in this run: https://github.com/kdeldycke/nuitka-issue-3325/actions/runs/13107996795

All the Ubuntu builds fail. The [binary is still produced on these platforms](https://github.com/kdeldycke/nuitka-issue-3325/actions/runs/13107996795/job/36565931926#step:6:24), but running them [ends up with this error message](https://github.com/kdeldycke/nuitka-issue-3325/actions/runs/13107996795/job/36565931926#step:8:5):

```text
my-cli-ubuntu-24.04.bin: error while loading shared libraries: libpython3.13.so.1.0: cannot open shared object file: No such file or directory
```

Note that in this case, Nuitka [report Python 3.13 flavor as unknown](https://github.com/kdeldycke/nuitka-issue-3325/actions/runs/13107996795/job/36565931926#step:6:6):

```text
Nuitka: Starting Python compilation with Nuitka '2.6.2' on Python (flavor Unknown), '3.13' commercial grade 'not installed'.
```

### Context

This has been tested with:
- Latest `nuitka`: 2.6.2
- Latest `uv`: 0.5.26
- Latest Python: 3.13
- Latest `actions/setup-python`: 5.4.0

geofft:
This fixes builds using python-build-standalone (the Python distribution installed by uv), which have an $ORIGIN-relative libpython dependency that should be treated like an $ORIGIN-relative rpath.

Fixes #3325.

## Summary by Sourcery

Treat shared library dependencies with paths relative to `$ORIGIN` as rpaths. This change ensures that dependencies are correctly resolved, even when the Python binary is installed in a non-standard location.

Bug Fixes:
- Fix builds using `python-build-standalone`, which have an `$ORIGIN`-relative `libpython` dependency that should be treated like an `$ORIGIN`-relative rpath.  Fixes #3325

Build:
- Treat shared library dependencies with paths as like rpaths when detecting binary path DLLs on POSIX systems.
- Handle multiple rpaths correctly when processing shared libraries on Darwin and ELF systems.

### DIFF
--- a/nuitka/freezer/DllDependenciesPosix.py
+++ b/nuitka/freezer/DllDependenciesPosix.py
@@ -22,10 +22,17 @@ from nuitka.utils.Utils import (
 from .DllDependenciesCommon import getLdLibraryPath
 
 # Detected Python rpath is cached.
-_detected_python_rpath = None
+_detected_python_rpaths = None
 
 # Cached ldd results.
 ldd_result_cache = {}
+
+
+def _resolveOriginValue(rpath, origin):
+    rpath = rpath.replace("$ORIGIN", origin)
+    rpath = os.path.normpath(os.path.join(origin, rpath))
+
+    return rpath
 
 
 def detectBinaryPathDLLsPosix(dll_filename, package_name, original_dir):
@@ -41,38 +48,26 @@ def detectBinaryPathDLLsPosix(dll_filename, package_name, original_dir):
     # This is the rpath of the Python binary, which will be effective when
     # loading the other DLLs too. This happens at least for Python installs
     # on Travis. pylint: disable=global-statement
-    global _detected_python_rpath
-    if _detected_python_rpath is None and not isPosixWindows():
-        _detected_python_rpath = getSharedLibraryRPATHs(sys.executable) or []
+    global _detected_python_rpaths
+    if _detected_python_rpaths is None and not isPosixWindows():
+        _detected_python_rpaths = getSharedLibraryRPATHs(sys.executable)
 
-        if _detected_python_rpath:
-            # Need to resolve a potential symlink.
-            if os.path.islink(sys.executable):
-                sys_executable = os.readlink(sys.executable)
-            else:
-                sys_executable = sys.executable
+        if os.path.islink(sys.executable):
+            sys_executable = os.readlink(sys.executable)
+        else:
+            sys_executable = sys.executable
 
-            for i, rpath in enumerate(_detected_python_rpath):
-                rpath = rpath.replace(
-                    "$ORIGIN", os.path.dirname(sys_executable)
-                )
+        sys_executable_dir = os.path.dirname(sys_executable)
 
-                rpath = os.path.normpath(
-                    os.path.join(os.path.dirname(sys.executable), rpath)
-                )
+        _detected_python_rpaths = [
+            _resolveOriginValue(rpath=detected_python_rpath, origin=sys_executable_dir)
+            for detected_python_rpath in _detected_python_rpaths
+        ]
 
-                _detected_python_rpath[i] = rpath
-
-    # Single one, might be wrong for Anaconda, which uses multiple ones on at least
-    # macOS.
-    python_rpaths = []
-    if _detected_python_rpath:
-        python_rpaths.extend(_detected_python_rpath)
+    python_rpaths = _detected_python_rpaths[:]
 
     if os.path.islink(dll_filename):
-        link_target_path = os.path.dirname(
-            os.path.realpath(dll_filename)
-        )
+        link_target_path = os.path.dirname(os.path.realpath(dll_filename))
         python_rpaths.append(link_target_path)
 
     # TODO: Actually would be better to pass it as env to the created process instead.
--- a/nuitka/freezer/DllDependenciesPosix.py
+++ b/nuitka/freezer/DllDependenciesPosix.py
@@ -12,7 +12,7 @@ from nuitka.containers.OrderedSets import OrderedSet
 from nuitka.PythonFlavors import isAnacondaPython
 from nuitka.Tracing import inclusion_logger
 from nuitka.utils.Execution import executeProcess, withEnvironmentPathAdded
-from nuitka.utils.SharedLibraries import getSharedLibraryRPATHS
+from nuitka.utils.SharedLibraries import getSharedLibraryRPATHs
 from nuitka.utils.Utils import (
     isAlpineLinux,
     isAndroidBasedLinux,
@@ -43,7 +43,7 @@ def detectBinaryPathDLLsPosix(dll_filename, package_name, original_dir):
     # on Travis. pylint: disable=global-statement
     global _detected_python_rpath
     if _detected_python_rpath is None and not isPosixWindows():
-        _detected_python_rpath = getSharedLibraryRPATHS(sys.executable) or []
+        _detected_python_rpath = getSharedLibraryRPATHs(sys.executable) or []
 
         if _detected_python_rpath:
             # Need to resolve a potential symlink.
--- a/nuitka/freezer/DllDependenciesPosix.py
+++ b/nuitka/freezer/DllDependenciesPosix.py
@@ -12,8 +12,7 @@ from nuitka.containers.OrderedSets import OrderedSet
 from nuitka.PythonFlavors import isAnacondaPython
 from nuitka.Tracing import inclusion_logger
 from nuitka.utils.Execution import executeProcess, withEnvironmentPathAdded
-from nuitka.utils.FileOperations import resolveSymlink
-from nuitka.utils.SharedLibraries import getSharedLibraryRPATH
+from nuitka.utils.SharedLibraries import getSharedLibraryRPATHS
 from nuitka.utils.Utils import (
     isAlpineLinux,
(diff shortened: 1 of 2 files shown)


nuitka/freezer/DllDependenciesPosix.py
```diff
@@ -44,32 +43,35 @@ def detectBinaryPathDLLsPosix(dll_filename, package_name, original_dir):
...
     global _detected_python_rpath
     if _detected_python_rpath is None and not isPosixWindows():
-        _detected_python_rpath = getSharedLibraryRPATH(sys.executable) or False
+        _detected_python_rpath = getSharedLibraryRPATHS(sys.executable) or []
 
         if _detected_python_rpath:
             # Need to resolve a potential symlink.
             if os.path.islink(sys.executable):
-                sys_executable = resolveSymlink(sys.executable)
+                sys_executable = os.path.realpath(sys.executable)
```
geofft, line 51: I think this is an existing bug that I can submit separately: resolveSymlink returns a relative path, not an absolute path, so it's not helpful here and the further filename manipulations below do the wrong thing.

nuitka/utils/SharedLibraries.py
```diff
@@ -368,21 +385,21 @@ def _getSharedLibraryRPATHDarwin(filename):
                 if str is not bytes:
                     result = result.decode("utf8")
 
-                return result
+                rpaths.append(result)
```
geofft, line 388: It's possible to have multiple `LC_RPATH` load commands (try `cc -shared -o blah -Wl,-rpath,a,-rpath,b; otool -L blah`). Since I had an excuse to make these functions return a list instead of a single value, I figured I'd make this change at the same time. I can submit it separately if we choose not to use this approach.

nuitka/freezer/DllDependenciesPosix.py
kayhayen: You are absolutely spot on, resolving it ourselves, makes only sense for things within rpaths, but not there. I am going to just fix it too.

nuitka/utils/SharedLibraries.py
kayhayen: Ah yes, that's just that I never saw a Python that does that.

nuitka/freezer/DllDependenciesPosix.py
```diff
@@ -44,32 +43,35 @@ def detectBinaryPathDLLsPosix(dll_filename, package_name, original_dir):
...
             else:
                 sys_executable = sys.executable
 
-            _detected_python_rpath = _detected_python_rpath.replace(
-                "$ORIGIN", os.path.dirname(sys_executable)
-            )
+            for i, rpath in enumerate(_detected_python_rpath):
```
kayhayen, line 55: I think I will strive to make this resolution a function, this happens in at least one other place too, and then we can just use a list contraction and don't need to modify the list.

kayhayen (changes requested):
I actually approve of it, but for the updating of rpaths and stylistically I am going to make a rebase after my next hotfix, and then I think we can even put this into a hotfix, seems like only an improvement.

kayhayen:
I am very confused now, why I saw this only this late, but recently I have been not keeping up it seems, sorry for the delay, this is very good work and very welcome. I normally am more up to date with reactions, I try.

For removing the rpath, it seems that `_setSharedLibraryRPATHDarwin` also needs to be adapted, since it needs to know the existing one before it can remove it. And my bike shedding would prefer the wording `RPATHs`, although yours is probably more English, is it not. At least Gemini says "The plural form of RPATH, when capitalized, would be RPATHs. Just add a lowercase "s" to the end." 

There may be other places, but I am sure this can be done like that.

I closed the issue referenced, because it started working for me, but maybe only because I used a venv activated, as that seems to make a difference somehow?

What I would do with this one, is to rebase and work myself a little bit on this, pushing into your branch if that's allowed, as it typically is.

kayhayen:
Seems I had forgotten to post my last comment, I rebased it just now, and will do a bit of said bike shedding cleanups.

nuitka/freezer/DllDependenciesPosix.py
kayhayen: This was a conflict, since that change was already in a hotfix in the hopes it already helps something

nuitka/utils/SharedLibraries.py
```diff
@@ -242,9 +243,24 @@ def _getSharedLibraryRPATHElf(filename):
...
 
-            return result
+            rpaths.append(result)
+        elif b"NEEDED" in line:
+            # If the Python binary has a library dependency like
+            # $ORIGIN/../lib/libpython.so, then treat it like it has an
+            # rpath of $ORIGIN/../lib. This is needed for
+            # python-build-standalone (used by uv).
+            result = line[line.find(b"[") + 1 : line.rfind(b"]")]
+            directory, slash, filename = result.rpartition(b"/")
```
kayhayen, line 253: Why is thos not using `os.path.dirname()` and simply checks if that is non-empty, and also I think, we might want to make sure we eliminate duplicates, as other code, e.g. the one removing rpaths on macOS may hate to try and remove one twice, so I am adding that.

kayhayen:
At least when removing the rpaths, these fake rpaths are an issue, because they are naturally very hard to remove, so I would expect that install_name_tool errors out on these, but we don't use it there, and get away with it with `patchelf --set-rpath` not needing it. However I don't quite like that, and will try and make these paths separate things, like returning a tuple, rpaths as a list and then a set of these directories, in then handle that the same for something and not others. We only really need that for detecting DLL dependencies on Posix it seems. But I am interested to hear back from you @geofft if this is all good in your book.

nuitka/utils/SharedLibraries.py
```diff
@@ -227,7 +227,8 @@ def _getDLLVersionWindows(filename):
 _readelf_usage = "The 'readelf' is used to analyse dependencies on ELF using systems and required to be found."
 
 
-def _getSharedLibraryRPATHElf(filename):
```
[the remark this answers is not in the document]
kayhayen: I don't see value in joining the two code paths as they deal with different tools

nuitka/utils/SharedLibraries.py
kayhayen: I made that change, hopefully it's good.

kayhayen:
So, this one seems good to go, and I think I found issues with torch and multiple rpaths it has, some of which need to be preserved, and since that will touch the very same codes, I want to merge this now, to make it part of the next hotfix.

I think since `:` is used in single rpath values as well as having multiple entries, it would be good to split rpaths up into an ordered set and combine them into one when writing out, but accept a list there as well. This is most definitely work for 2.7 though.

PR #3333, opened 2025-02-07
Scons: Add rpath for relocatable Python installations

### LINKED ISSUE #3325: Binaries compiled with a Python installed by `uv` fails on Ubuntu: `libpython3.13.so.1.0` not found

geofft:
Relocatable installations are not installed to a system-wide path and therefore compiling against their libpython requires using an rpath. At the moment the heuristic is whether the installation comes from python-build-standalone.

Fixes #3325.

### DIFF
--- a/nuitka/PythonFlavors.py
+++ b/nuitka/PythonFlavors.py
@@ -220,6 +220,19 @@ def isUninstalledPython():
     return isAnacondaPython() or "WinPython" in sys.version
 
 
+def isRelocatable():
+    """Is libpython not in the system global library directory."""
+    import sysconfig
+
+    if sysconfig.get_config_var("PYTHON_BUILD_STANDALONE") == 1:
+        return True
+
+    if isSelfCompiledPythonUninstalled():
+        return True
+
+    return False
+
+
 _is_win_python = None
 
 
--- a/nuitka/build/Backend.scons
+++ b/nuitka/build/Backend.scons
@@ -812,6 +812,8 @@ elif env.exe_mode:
     # For NetBSD the rpath is required, on FreeBSD it's warned as unused.
     if isNetBSD():
         env.Append(LINKFLAGS=["-rpath=" + python_lib_path])
+    elif env.relocatable:
+        env.Append(LINKFLAGS=["-Wl,-rpath=" + python_lib_path])
 
 if isMacOS():
     if env.module_mode or env.dll_mode:
--- a/nuitka/build/SconsInterface.py
+++ b/nuitka/build/SconsInterface.py
@@ -34,6 +34,7 @@ from nuitka.PythonFlavors import (
     isAnacondaPython,
     isMSYS2MingwPython,
     isNuitkaPython,
+    isRelocatable,
     isSelfCompiledPythonUninstalled,
 )
 from nuitka.PythonVersions import (
@@ -544,6 +545,9 @@ def getCommonSconsOptions():
 
     if isSelfCompiledPythonUninstalled():
         scons_options["self_compiled_python_uninstalled"] = asBoolStr(True)
+
+    if isRelocatable():
+        scons_options["relocatable"] = asBoolStr(True)
 
     cpp_defines = Plugins.getPreprocessorSymbols()
     if cpp_defines:
--- a/nuitka/build/SconsUtils.py
+++ b/nuitka/build/SconsUtils.py
@@ -250,6 +250,7 @@ def createEnvironment(
     env.self_compiled_python_uninstalled = getArgumentBool(
         "self_compiled_python_uninstalled", False
     )
+    env.relocatable = getArgumentBool("relocatable", False)
 
     # Non-elf binary, important for linker settings.
     env.noelf_mode = getArgumentBool("noelf_mode", False)


nuitka/PythonFlavors.py
```diff
@@ -220,6 +220,19 @@ def isUninstalledPython():
     return isAnacondaPython() or "WinPython" in sys.version
 
 
+def isRelocatable():
```
kayhayen, line 223: I think I have called these things "uninstalled". And for them, a shell script is created to execute them as of recently, this strives to also persist the PYTHONPATH at compile time, PYTHONHOME, and what not.

Setting an rpath to find the libpython for sure, is a good idea anyway. And actually maybe independent of the fact if it's a system installed Python or not.

As for `sysconfig.get_config_var("PYTHON_BUILD_STANDALONE")` I am starting to feel, that this is a Python flavor, and UV Python actually is not, but might be producing CPython or Standalone Python (I need to check for proper naming), and that I should have that as the flavor instead.

nuitka/build/Backend.scons
```diff
@@ -812,6 +812,8 @@ elif env.exe_mode:
     # For NetBSD the rpath is required, on FreeBSD it's warned as unused.
     if isNetBSD():
         env.Append(LINKFLAGS=["-rpath=" + python_lib_path])
+    elif env.relocatable:
```
kayhayen, line 815: I am thinking, we might just do this all the time, if the goal is to make a specific Python configuration executable. For the later I am talking of accelerated mode. And with the rpath feature being so universally available, there really is no hard need to have the scipt anymore. Already for accelerated mode, some settings are fed into the binary. For standalone that of course is not the case anyway.

nuitka/PythonFlavors.py
kayhayen: Check `isUninstalledPython` for the current uses.

nuitka/build/Backend.scons
kayhayen: Yeah just checked, the rpath is not set by accelerated mode scripts right now. And actually setting `PATH` on Windows to be able to load it, was like the only real reason `.CMD` files were ever created there, but the newly added shell scripts are not technically necessary with your change at all, and not even doing what this does.

kayhayen:
You didn't respond to my ideas about doing it always, and I tend to take that as agreement. I think we understand that for your Python flavor, it is needed for sure, and as such I am now going to rebase it, but making the rpath non-optional, and to only disable it where the toolchain doesn't support that feature at most.

kayhayen:
So, this rebased to 2.6.6 without any issues.

kayhayen:
And I implemented my idea, renaming the flavor, please let me know if you like it, and inviting you to test it, I didn't quite check it yet. Doing it always should not break UV-Python's use of Standalone Build Python. Adding it as a flavor will make uses outside of uv supported as well I think.

kayhayen:
It seems, somehow the rpath added survives on Linux now, where of course it should be removed for standalone to be working properly, I need to check that out.

kayhayen:
So, it was only removing the rpath for the main executable on macOS, which may have been too little anyway, need to do it for all platforms now, as it now gets added always, that should repair general Linux standalone. Glad that Github Actions Python detected that with it not using static libpython, in my setups it didn't show.

kayhayen:
Oh the joy of it, macOS (maybe Clang?) won't accept `-Wl,-rpath` while others won't accept `-rpath` it seems, making a difference there now.

kayhayen:
Rebased it to the current develop branch after the other PRs merge. It's very disturbing that this still fails, despite it claiming to set the rpath properly on `CtypesUsing.bin` to `$ORIGIN` and somehow the strace output done on the running program is not complete, since it never mentions said DLL, but the libffi one from hostedtools cache, I will try and reproduce this locally to see where it goes wrong.

kayhayen:
So, my suspect is that DT_RPATH is used by GitHub Actions Python, but no say my Debian Python, where i force it to use --static-libpython=no to reproduce it, so adding `--remove-rpath` according to patchelf man page should remove both and then setting it doesn't have us end up with both, with apparently the `DT_RPATH` then taking precedence. If that holds true, Nuitka would actually already have been bad at removing these kind of rpaths, which would be an issue by itself.

kayhayen:
So, I am being stupid, but I cannot get Github Actions Python to work. I have addressed the issue of strace output being interspersed with other Python outputs hopefully, and I tried uv-Python with the Python Standalone Build flavor, and I discovered it has in its config the `:lib_Hacl_SHA2.a` for linking, which is as practically all Python3.12 doesn't provide, but this normally only shows when statically linking, with that flavor, it's happening also when Python is dynamically linked.

I wonder if `:something` should be filtered out entirely for non-static libpython, since it's kind of inconsequential, and maybe even using it is the bug.

kayhayen:
So, this works flawlessly for me with Debian Python, "Standalone Python Build"-Python, such that I am getting really confused, also because the assertion that `$ORIGIN` is properly set also holding true (there were issues with caching on macOS). It actually should be impossible, and then also, even if interspersed, I did never see it load libpython in the strace, but maybe I will now.

kayhayen:
I am at a complete loss there, it seems the ELF output is perfect:

```
: ELF listing of main binary:

Dynamic section at offset 0x38a000 contains 30 entries:
  Tag        Type                         Name/Value
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libpython3.8.so.1.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000c (INIT)               0xb000
 0x000000000000000d (FINI)               0xb5964
 0x0000000000000019 (INIT_ARRAY)         0x383618
 0x000000000000001b (INIT_ARRAYSZ)       8 (bytes)
 0x000000000000001a (FINI_ARRAY)         0x383620
 0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0x39b760
 0x0000000000000005 (STRTAB)             0x39a230
 0x0000000000000006 (SYMTAB)             0x660
 0x000000000000000a (STRSZ)              5421 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000015 (DEBUG)              0x0
 0x0000000000000003 (PLTGOT)             0x383850
 0x0000000000000002 (PLTRELSZ)           5688 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x9600
 0x0000000000000007 (RELA)               0x3d80
 0x0000000000000008 (RELASZ)             22656 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x0000000000000018 (BIND_NOW)           
 0x000000006ffffffb (FLAGS_1)            Flags: NOW PIE
 0x000000006ffffffe (VERNEED)            0x3cc0
 0x000000006fffffff (VERNEEDNUM)         2
 0x000000006ffffff0 (VERSYM)             0x3a2e
 0x000000006ffffff9 (RELACOUNT)          863
 0x0000000000000000 (NULL)               0x0
```

Yet, the runtime traces taking complains the GitHub Action Python libpython is being loaded, for which there wouldn't be a reason why. And esp. if this wasn't sufficient, I don't see why the old code didn't show that happen, if somehow it was forced through environment or something. In the old code, it just didn't have an rpath, how would that make it check the local directory first, maybe `$ORIGIN` works differently from what I think it should.

kayhayen:
So, this comes down to `LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.8.18/x64/lib` with which the binary is getting executed, and according to my testing, this forces the run resolution to not use the DLLs at `$ORIGIN` but to prefer anything found there, this is of course nothing new. I am deeply confused what causes this to be visible now and I need to investigate it carefully, it's alsmost as if we stopped using static libpython for GitHub Actions Python somehow, or something like that.

kayhayen:
So, I think, I am making the executive decision, to minimize the differences of this PR to develop, and skip this for the next hotfix, unless I figure out what's going on through that. Failing tests on Github Actions is not acceptable. Obviously unsetting LD_LIBRARY_PATH before any execution of a standalone binary is kind of a must, for the strace to hold any water, but I would really like to understand why I haven't noticed this issue previously, since it's absolutely presenct on current hotfix branch as well, but strace doesn't load it from there. I tried manually removing the RPATH, to see if that's what causes it, but on my Debian that didn't do anything particular, so I am at a loss what is bug hiding there, or how the strace analysis is not effective previously.

kayhayen:
So, I rebased this to factory branch now, to which I picked the changes that are in here, but could be separate, flavor rename, lib_Hacl2_SHA2 issue, and the minor test improvement to display the resulting file.

kayhayen:
Now GitHub is just giving me a hard time for the merge of that other PR that now conflicts, I of course put that for the hotfix, so really, I think the best way ahead is to make the hotfix release now and rebase onto the then develop. The diff is really small now, but it seems centered around touching the linker rpath. I am contemplating to add to develop a change where for one standalone binary, it does the output of readelf anyway, so I can see what would be the change, maybe `DT_RPATH` vs. `DT_RUNPATH` or something stupid like that, although forcing to to `DT_RPATH` didn't impact my Debian now, and it would of course never be the real solution, since it wouldn't work everywhere.

kayhayen:
So, with this rebased and minimal, still fails. Maybe an explanation why I don't just fix the `LD_LIBRARY_PATH` issue, but want to know why it's not been observed previously. In ATC I have learned that the most effective way of doing things, is not to fix the symptom you are seeing, but all things that have contributed to seeing it and to not seeing it earlier. So for me, it's super important to be able to trust these standalone tests that try to assert no outside DLLs are used. Something wasn't working there right, or some assumption I am holding is wrong. Current develop does have to fail I think, but it does not.

So, what I will probably add, is like a standalone compilation with artifacts like a report, and the readelf output, and the strace captured on execution, such that I have a reference point to easily compare the situation. We are down to just adding the rpath early, during compilation, and that causing the test to fail, and it being added later, to not do that. But also, it's being removed and replaced entirely, and yet there is a difference. I would like to get to the bottom of that, and Github Actions Python is obviously a relatively difficult environment for that.

kayhayen:
So, I was distracted privately with real life Nuitka being in hospital for close to 2 weeks, but I am finally getting to this, and have updated factory with the recording of reports.

For a simple standalone this is what ldd gives without the PR code:

```
	linux-vdso.so.1 (0x00007ffe96908000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f4549719000)
	libpython3.8.so.1.0 => /home/runner/work/Nuitka/Nuitka/EmptyModuleTest.dist/libpython3.8.so.1.0 (0x00007f4549200000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f4548e00000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f4549c25000)
```

And this is what readelf says:

```
Dynamic section at offset 0x359630 contains 30 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libpython3.8.so.1.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN]
 0x000000000000000c (INIT)               0xb000
 0x000000000000000d (FINI)               0x9fe84
 0x0000000000000019 (INIT_ARRAY)         0x35a618
 0x000000000000001b (INIT_ARRAYSZ)       8 (bytes)
 0x000000000000001a (FINI_ARRAY)         0x35a620
 0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0x3b0
 0x0000000000000005 (STRTAB)             0x24e8
 0x0000000000000006 (SYMTAB)             0x658
 0x000000000000000a (STRSZ)              5316 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000015 (DEBUG)              0x0
 0x0000000000000003 (PLTGOT)             0x35a850
 0x0000000000000002 (PLTRELSZ)           5688 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x94a0
 0x0000000000000007 (RELA)               0x3cf8
 0x0000000000000008 (RELASZ)             22440 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x0000000000000018 (BIND_NOW)           
 0x000000006ffffffb (FLAGS_1)            Flags: NOW PIE
 0x000000006ffffffe (VERNEED)            0x3c38
 0x000000006fffffff (VERNEEDNUM)         2
 0x000000006ffffff0 (VERSYM)             0x39ac
 0x000000006ffffff9 (RELACOUNT)          855
 0x0000000000000000 (NULL)               0x0
```

Both are good, but supposedly the PR breaks that, and I don't know how, but lets first confirm that by me rebasing to current factory branch.

kayhayen:
So, indeed the regression is this PR branch, and not an old issue, noticed only now, the readelf has a `RUNPATH` which apparently must be weaker in terms of forcing the directory usage.

```
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libpython3.8.so.1.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
```

The ldd output then gives this:

```
	linux-vdso.so.1 (0x00007ffd71cfc000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f9dd5ddd000)
	libpython3.8.so.1.0 => /opt/hostedtoolcache/Python/3.8.18/x64/lib/libpython3.8.so.1.0 (0x00007f9dd5600000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9dd5200000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f9dd5ed2000)
```

And while this affects libpython here, I guess it could affect practially all other DLLs contained. 

I thought I had made sure it's a RPATH vs. RUNPATH, but somehow unsuccessfully maybe.

kayhayen:
Pretty sure I tried `patchelf` with `--force-rpath` in some capacity, however, on my Debian just now it switches it over to be a `RPATH` like it used to be. Previously, the `RPATH` nature wasn't enforced, but it seems it's better for isolation by far, and as such is a good finding.

Of course, "Standalone Build Python" ought to not need a dynamic libpython, but rather libpython is not off the table, but I guess that's still a very good improvement to Nuitka. I wasn't aware of there being such a difference between `RPATH` and `RUNPATH` with `LD_LIBRARY_PATH`.

kayhayen:
Rebased it back to develop and adapted the commit message, a bunch of fixes has already been evacuated from this PR before, leaving only the always-rpath change in.

kayhayen:
I merged this but will put it into 2.6.8 hotfix directly if that works.

kdeldycke:
The newly released Nuitka 2.6.8 fixed the issue for me! 🎉

See: https://github.com/Nuitka/Nuitka/issues/3325#issuecomment-2711540520

PR #3349, opened 2025-02-17
Thin flto POC

KRRT7:
this works on linux, macOS and windows as-is

### DIFF
--- a/nuitka/build/SconsCompilerSettings.py
+++ b/nuitka/build/SconsCompilerSettings.py
@@ -230,14 +230,16 @@ slower without it.
 
     if env.gcc_mode and lto_mode:
         if env.clang_mode:
-            env.Append(CCFLAGS=["-flto"])
-            env.Append(LINKFLAGS=["-flto"])
+            env.Append(CCFLAGS=["-flto=thin"])
+            env.Append(LINKFLAGS=["-flto=thin"])
         else:
-            env.Append(CCFLAGS=["-flto=%d" % job_count])
-            env.Append(LINKFLAGS=["-flto=%d" % job_count])
+            env.Append(CCFLAGS=["-flto=thin"])
+            env.Append(LINKFLAGS=["-flto=thin"])
 
             env.Append(CCFLAGS=["-fuse-linker-plugin", "-fno-fat-lto-objects"])
             env.Append(LINKFLAGS=["-fuse-linker-plugin"])
+
+            env.Append(LINKFLAGS=["-Wl,-plugin-opt,thinlto-jobs=%d" % job_count])
 
             # Need to tell the linker these things are OK.
             env.Append(LINKFLAGS=["-fpartial-inlining", "-freorder-functions"])


KRRT7:
as discussed i don't know much about compiler and their options to flesh it our more than what is currently here, but it works at least cross platform from my tests.

kayhayen:
When trying to google it, I found no evidence of thin LTO for gcc, it just errors out when I tried it on Linux:

```
Nuitka-Scons: Backend C compiler: gcc (gcc 13).
cc1: error: unrecognized argument to '-flto=' option: 'thin'
```

And

```
Nuitka-Scons: Backend C compiler: gcc (gcc 14.2.0).
cc1.exe: error: unrecognized argument to '-flto=' option: 'thin'
```

So this would be a clang only change, macOS has a clang as its gcc, so there it would probably work.

The Clang docs haven't been updated in ages it seems: https://clang.llvm.org/docs/ThinLTO.html

I somehow doubt that clang 20 hasn't yet made this the default or integrated it, when the docs speak of support in 3.8, but also people are seemingly comparing it, with no clear conclusion on performance, which however I can make if I run this with Nuitka Speedcenter and valgrind I guess. However, the gcc branch not working at all, makes me wonder what I am doing wrong there, I think gcc 15 is unreleased, so must have been dreaming of it working?!

kayhayen:
Also, this doesn't affect ClangCL, I wonder if it would work there too. And for --mingw --clang we have this

> Nuitka-Scons: LTO mode auto was resolved to mode: 'no' (known to not be supported (new MinGW64 Clang)).

If I force it with `--lto --mingw64 --clang`, I get the familar message:

> C:/Users/kayha/AppData/Local/Nuitka/Nuitka/Cache/downloads/gcc/x86_64/14.2.0posix-19.1.1-12.0.0-msvcrt-r2/mingw64/bin/ld: C:/Users/kayha/AppData/Local/Nuitka/Nuitka/Cache/downloads/gcc/x86_64/14.2.0posix-19.1.1-12.0.0-msvcrt-r2/mingw64/bin/../lib/LLVMgold.dll: error loading plugin:
> clang: error: linker command failed with exit code 1 (use -v to see invocation)

I imagine the ClangCL actually has a chance of this working, but I believe it might be ignored there, and yes, it's actively not activated there, which however may be not justified.

kayhayen:
So, this leaves us with macOS for experimenting, and maybe Clang on Linux. I would be most interested in runtime performance on a libstatic-python using setup, but those are harder to come by I fear.

polarathene:
> people are seemingly comparing it, with no clear conclusion on performance

I've seen resources online somewhere some time ago that demonstrated that ThinLTO was much faster (and less memory required) to perform than Full LTO. The performance delta was often shown to be minimal (_around 90-95% or something of full LTO IIRC_), so it was a good tradeoff IIRC. I could possibly try dig up a reference for you if you'd like?

I haven't personally investigated it with real-world projects and benching though. I looked into this at the time with Rust where it's an LTO option.

kayhayen:
Sharing hearsay is of course a very important thing to do. Anyway, as part of my scalability push, I have added for the factory branch recently changes to capture the `rusage` information during compilation. That will allow us to track kernel level accounting of CPU time for Scons compilation jobs. There are some things missing for these to show up in reports. And then I would like to become able to compare the numbers even between different compilers.

It's a bit too early for this, but I rebased it to current develop to resolve the conflicts I caused by rebasing for the release recently, and I am now going to make it an experimental flag to put the change to use or not, then merge it. First that will allow to use it with mainline Nuitka at any time before I make the determinations. Second it gets this PR closed which is a bad form of a TODO item.

Based on the reports, I hope to become able to say definitive things about various compiler performances for the actual C backend phase and LTO linking phase. I will however track that as an issue in Nuitka-Website.

kayhayen:
Still not sure about gcc there, is this AI generated crap for gcc @KRRT7 or is that for a newer gcc only?

KRRT7:
@kayhayen see for yourself: https://clang.llvm.org/docs/ThinLTO.html#controlling-backend-parallelism

kayhayen:
Why do I need to look at Clang docs for gcc thin-LTO ?

kayhayen:
I am removing the changes for gcc, that for what I had previously already researches, does not exist. You may resurrect them if you think they are real. At the time I didn't consider that elaborate options that don't exist are just a sign of AI usage and not a lack of my personal knowledge.

KRRT7:
oops, I does seem like gcc has https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-flto flto

kayhayen:
So, this goes on in https://github.com/Nuitka/Nuitka-website/issues/124 from now. Once I have the ability to identify long running C tasks not just via compile time warning, but through the actual reports, as required to check out C scalability improvements, we can solve https://github.com/Nuitka/Nuitka-website/issues/124 just fine.

Issue #3283, opened 2025-01-03
How to change PYTHONHOME compiled by accelerated mode on Windows


YAGregor:
Before submitting an Issue, please review the
[Issue Guidelines](https://github.com/Nuitka/Nuitka/blob/develop/CONTRIBUTING.md#submitting-an-issue).

- [x] Please check whether the bug was already reported or fixed.

- [x] Please check whether the issue appears when you run the code with Python, if it does, it's not a
  Nuitka issue.

- [ ] Please check out if the develop version of Nuitka works better for you.

  Download source, packages [from here](https://nuitka.net/doc/download.html) where you will also
  find instructions how to do it via pip or git.

If you want to post a problem/bug, to help us understand and resolve your issue please check that
you have provided at least the information below, and discard up to here:

- Nuitka version, full Python version, flavor, OS, etc. as output by *this exact* command.

  > python -m nuitka --version
4.5.9
Commercial: None
Python: 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13) [MSC v.1929 64 bit (AMD64)]
Flavor: Unknown
Executable: ~\AppData\Local\pypoetry\Cache\virtualenvs\*-LuQ0_9Xc-py3.10\Scripts\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 10
Version C compiler: C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.42.34433\bin\Hostx64\x64\cl.exe (cl 14.3).

- How did you install Nuitka and Python
install python with pyenv, install nuitka by poetry `poetry add nuitka --group=dev`

  > python -m pip list -v

- Also supply a Short, Self Contained, Correct, Example
minimal reproduce:
```python
if name == 'main':
    print("it works")
```
compile it in accelerate mode, move python install location, and make it still run
- Provide in your issue the Nuitka options used
`nuitka.cmd debug.py`

motivation: I have an app with a lot of pypi packages and I don't want to compile all of them (too slow and risk to compile fail), I want to deploy my app by just copy pypi packages or copy the whole python env.

tried:
1. compile app by standalone `nuitka debug.py --standalone --nofollow-import-to=pyqt5,numpy.....`, the compiled environment lacks some standard libs, eg: nuitka compiled logging, but can't import logging.config (import by third part package).
[...]

PR #3376, opened 2025-03-09
Fix, respect PYTHONHOME environment variable in accelerate mode and uninstalled python detected

YAGregor:
since for uninstalled python,script file is created to set environment variable, there is no need to set PYTHON_HOME_PATH in MainProgram.c.

# What does this PR do?

I noticed that, when `uninstalled_python` detected, scons backend generate PYTHON_HOME_PATH in `build_definition.h`, and set PYTHONHOME in MainProgram.c,which will cover PYTHONHOME environment variable. But when `uninstalled_python` detected, there also must be a script file for execution,so it's not necessary and I can custom PYTHON_HOME in accelerate mode

I want to make compiling faster by do not compile pypi packages, and just copy CPython interpreter and site_packages, I know it's not offical supported but I really need this.
https://github.com/Nuitka/Nuitka/issues/3283

# PR Checklist

## Summary by Sourcery

Respect the PYTHONHOME environment variable when running in accelerate mode with an uninstalled Python. This change removes the forced setting of PYTHONHOME in MainProgram.c, allowing users to customize it.

### DIFF
--- a/nuitka/build/Backend.scons
+++ b/nuitka/build/Backend.scons
@@ -176,9 +176,6 @@ pgo_mode = getArgumentDefaulted("pgo_mode", "no")
 # Console mode specified
 console_mode = getArgumentDefaulted("console_mode", "attach")
 
-# Windows might be running a Python whose DLL we have to use.
-uninstalled_python = getArgumentBool("uninstalled_python", False)
-
 # Unstripped mode: Do not remove debug symbols.
 unstripped_mode = getArgumentBool("unstripped_mode", False)
 
@@ -954,10 +951,6 @@ if job_count:
 
 def createBuildDefinitionsFile():
     build_definitions = {}
-
-    if uninstalled_python and not env.static_libpython:
-        # Use the non-external one, so it's not a short path.
-        build_definitions["PYTHON_HOME_PATH"] = python_prefix
 
     build_definitions["NO_PYTHON_WARNINGS"] = 1 if no_python_warnings else 0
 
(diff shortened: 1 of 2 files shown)


kayhayen:
I rebased the PR to the latest pre-release, since I do rebasing there.

kayhayen:
I am not so sure about the need of these scripts, that I added recently only, and actually might want to go the other route. I understand your motivation might be in making it possible to point it at a different path, right?

What I thought they would be for is to find the libpython of an uninstalled Python, which however with recent changes from 2.6.8 for the "Python Build Standalone" flavor that UV downloads, always adds a rpath, that could render these unnecessary unlike on Windows, where setting a PATH to find `CPythonXX.DLL` will always be necessary.

The variable PYTHONHOME may be set by a virtualenv used, but ideally the accelerated binary ignores that and isn't influenced by its PYTHONPATH either, etc. and this needs more love.

But the way I am thinking right now, this seems not the direction I want to go. I do not mean accelerated mode to become a lightweight distribution method, althought we definitely could consider making it an option, to do one or the other. I do not want to prevent such usages either. I will have to look at that, how hard it is, probably not that hard, but rpath for python DLL is kind of forced on non-Windows.

kayhayen:
I rebased the PR to 2.7.2, since I did a big history re-write that was necessary. I meant to get your intention with my last comment, do you want to share it?

YAGregor:
In fact I don't need this PR any more，now I work around software distribution by compile all code base into a pyd (and load by an exe written in Rust). It's much more safe and easy to understand, so please close this PR if it's not fit Nuitka's design.

PR #3475, opened 2025-05-19
Update standard.nuitka-package.config.yml

SirDank:
Include additional directories for `ursina` and `panda3d`

## Summary by Sourcery

Update the standard Nuitka package configuration to include missing asset directories for the panda3d and ursina modules.

Enhancements:
- Add the 'models' directory to the panda3d package data files
- Include 'audio', 'fonts', and 'textures' directories in ursina's package data files


kayhayen:
Thanks, this is the one I was looking for, merging, but later I will put it out with the next hotfix.

PR #3476, opened 2025-05-19
Add arcade gl backends to arcade package config

csd4ni3l:
This PR adds `arcade.gl.backends` to the implicit imports of arcade. This is a new submodule in the development branch which is only imported using `importlib`, so Nuitka wont include it by default.

## Summary by Sourcery

Add the new arcade.gl.backends submodule to Arcade’s implicit imports and update config checksums for several modules

Enhancements:
- Include arcade.gl.backends in the implicit imports for the arcade package config
- Update checksum entries for the arcade, panda3d, and ursina modules

### DIFF
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -202,20 +202,20 @@
     - depends:
         - 'apt_pkg'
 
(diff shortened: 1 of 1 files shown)


csd4ni3l:
Just to note, the last commit did not run autoformat and currently dev is failing on the formatting test, so thats probably why my autoformat updated the checksums

PR #3496, opened 2025-06-04
Add extra dependency group for Nuitka build-backend

b-phi:
# What does this PR do?
Add a new optional dependency group to include build dependencies when using Nuitka as a build backend. Dependencies taken from [this section](https://nuitka.net/user-documentation/use-cases.html#setuptools-wheels). Asides from being slightly cleaner for users, this would help avoid downstream breaks if the minimum version of setuptools ever needs to be bumped.

Before
```toml
[build-system]
build-backend = 'nuitka.distutils.Build'
requires = ['nuitka', 'setuptools>=42', 'toml', 'wheel']
```

After
```toml
[build-system]
build-backend = 'nuitka.distutils.Build'
requires = ['nuitka[build]']
```

# PR Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -495,6 +495,7 @@ Python compiler with full language support and CPython compatibility""",
         "console_scripts": console_scripts,
     },
     install_requires=install_requires,
+    extras_require={"build": ["setuptools>=42", "toml", "wheel"]},
     # As we do version specific hacks for installed inline copies, make the
     # wheel version and platform specific.
     distclass=BinaryDistribution,


kayhayen:
It's a shame, but I don't know how to do these things myself, outside of Nuitka I am not packing anything. I am going to upload it, I would prefer dependencies to be on `Nuitka` rather than `nuitka` since I don't want to depend on case being ignored, but it's great stuff. Also if we ever build standalone or onefile, we can add more groups, so maybe I change it to `build-wheel` for now, and add `build-onefile` one day and have `onefile` and `standalone` added now, this is actually a very important lesson I am being taught here.

kayhayen:
Stupid question, these things do work when I install from GitHub URLs like such: 

```
# With "python" being the version where you want to install Nuitka

# Standard version:
python -m pip install -U --force-reinstall "https://github.com/Nuitka/Nuitka/archive/factory.zip"
# Commercial version (subscribers only)
python -m pip install -U --force-reinstall "https://github.com/Nuitka/Nuitka-commercial/archive/staging.zip"
```

Do they?

kayhayen:
To answer my own question: python -m pip install -U --force-reinstall "https://github.com/Nuitka/Nuitka/archive/develop.zip#egg=Nuitka[build]" works for now, but it gives a warning:

DEPRECATION: https://github.com/Nuitka/Nuitka/archive/develop.zip#egg=Nuitka[build] contains an egg fragment with a non-PEP 508 name. pip 25.1 will enforce this behaviour change. A possible replacement is to use the req @ url syntax, and remove the egg fragment. Discussion can be found at https://github.com/pypa/pip/issues/13157

This works without complaints:

python -m pip install -U --force-reinstall "Nuitka[build] @ https://github.com/Nuitka/Nuitka/archive/develop.zip"

and it means we can use these everywhere. 

I have yet to rename it to `build-wheel` though, afterwards, this concrete command will start to fail of course.

Issue #2386, opened 2023-08-14
Trouble with Python 3.11 Exception Groups


yqxk:
Hello,

This looks like an amazing project, and I'm excited to try it. I've been playing around with python 3.11 exception groups, and I'm not sure if I'm doing something wrongly, but I can't seem to get a simple example to run with nuitka:

```
root@9af304f0cf5e:/tmp# cat test.py
def main():
    try:
        pass
    except* Exception:
        pass

root@9af304f0cf5e:/tmp# python3.11 -m nuitka --version
1.7.9
Commercial: None
Python: 3.11.4 (main, Jun  7 2023, 12:45:48) [GCC 11.3.0]
Flavor: Debian Python
Executable: /usr/bin/python3.11
OS: Linux
Arch: x86_64
Distribution: Ubuntu (based on Debian) 22.04.1
Version C compiler: /usr/bin/gcc (gcc).
```

```
root@9af304f0cf5e:/tmp# python3.11 -m nuitka --standalone test.py
Nuitka-Options:INFO: Used command line options: --standalone test.py
Nuitka:INFO: Starting Python compilation with Nuitka '1.7.9' on Python '3.11' commercial grade 'not installed'.
Nuitka-Optimization:WARNING: Problem at 'test.py:2' with TryStar(body=[Pass()], handlers=[ExceptHandler(type=Name(id='Exception', ctx=Load()), body=[Pass()])], orelse=[], finalbody=[]).
Nuitka-Optimization:WARNING: Problem at 'test.py:1' with FunctionDef(name='main', args=arguments(posonlyargs=[], args=[], kwonlyargs=[], kw_defaults=[], defaults=[]), body=[TryStar(body=[Pass()],
Nuitka-Optimization:WARNING: handlers=[ExceptHandler(type=Name(id='Exception', ctx=Load()), body=[Pass()])], orelse=[], finalbody=[])], decorator_list=[]).
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nuitka/__main__.py", line 155, in <module>
    main()
  File "/usr/lib/python3/dist-packages/nuitka/__main__.py", line 141, in main
    MainControl.main()
  File "/usr/lib/python3/dist-packages/nuitka/MainControl.py", line 1109, in main
    _main()
  File "/usr/lib/python3/dist-packages/nuitka/MainControl.py", line 969, in _main
    main_module = _createMainModule()
                  ^^^^^^^^^^^^^^^^^^^
[...]

PR #3509, opened 2025-06-11
Implement exception groups

ZeroIntensity:
# What does this PR do?

This implements `except*` as outlined in PEP 654.

Note that this is currently a very rough outline of how this is going to work, and definitely not finished yet. The `except*` reformulation is going to be the following:

```py
try:
    ...
except BaseException:  # Equivalent to except* exception_types where exception_types is a tuple
    _tmp_error = sys.exc_info()[1]
    if isinstance(_tmp_error, BaseExceptionGroup):
        err = _tmp_error
    else:
        err = ExceptionGroup('', (_tmp_error,))

    for _type in exception_types:
        for _val in err.exceptions:
            if isinstance(_val, _type):
                break
        else:
            raise
```

See #2386.

# PR Checklist

## Summary by Sourcery

Implement core PEP 654 exception group support by extending exception matching, AST transformations, and code generation to handle 'except*' constructs.

New Features:
- Add preliminary support for PEP 654 exception groups and 'except*' handlers in try/except statements.

Enhancements:
- Extend the C API with EXCEPTION_GROUP_MATCH_BOOL to enforce correct exception group matching semantics.
- Introduce ExpressionComparisonExceptionGroupMatch and integrate the 'exception_group_match' comparator into code generation.
- Augment try/except AST reformulation to emit specialized handling logic for starred exception handlers.


ZeroIntensity:
Tests are going to all fail right now (as expected), but if you have any advice/concerns so far @kayhayen be sure to let me know.

nuitka/build/include/nuitka/exceptions.h
```diff
@@ -1272,6 +1272,43 @@ NUITKA_MAY_BE_UNUSED static inline int EXCEPTION_MATCH_BOOL(PyThreadState *tstat
...
+        }
+    }
+    else {
+        is_subclass = PyObject_IsSubclass(exceptions_checked, PyExc_BaseExceptionGroup);
+        if (is_subclass < 0) {
+            return -1;
+        }
+    }
+    if (is_subclass) {
+        _PyErr_SetString(tstate, PyExc_TypeError,
```
kayhayen: Sometimes we have our own variants of CPython API that should be used, like this would be `SET_CURRENT_EXCEPTION_TYPE0_STR` and similarily, and e.g. `Nuitka_Object_IsSubclass` is not ready yet, so I am leaving those calls in, but do not concern yourself with this at this time.

nuitka/build/include/nuitka/exceptions.h
```diff
@@ -1272,6 +1272,43 @@ NUITKA_MAY_BE_UNUSED static inline int EXCEPTION_MATCH_BOOL(PyThreadState *tstat
     }
 }
 
+NUITKA_MAY_BE_UNUSED static inline int EXCEPTION_GROUP_MATCH_BOOL(PyThreadState *tstate, PyObject *exceptions_got,
+                                                                  PyObject *exceptions_checked) {
+    CHECK_OBJECT(exceptions_got);
+    CHECK_OBJECT(exceptions_checked);
+
+#if PYTHON_VERSION >= 0x3b0
```
kayhayen: The define should apply to the whole function. I already added a file `exception_groups.h`  that has that define itself, but is also included like this:

```
#if PYTHON_VERSION >= 0x3b0
#include "nuitka/exception_groups.h"
#endif
```

So move it to there, and maybe build off the code I already have in that file.

kayhayen:
Looking at the diff, one thing you ought to do, is to have the `misc/install-git-hooks.py` executed, so it will autoformat on commit, which will apply a combination of black, isort, clang-format to the code.

I believe the change of the published exception, the place that `sys.exc_info()` is pulling from, is not yet there.

kayhayen:
Executing `./bin/autoformat-nuitka-source` will format the code manually as needed too.

kayhayen:
One observation, if you don't develop on Windows, `_PyExc_CreateExceptionGroup` will be working on that OS, but as it's not exported, Windows and maybe macOS too, will have the issue of not linking that, so once we finish this on Linux, we will be facing issues with having to implement that ourselves too.

ZeroIntensity:
Definitely getting closer to fruition. The main part that I need to iron out is the partial matching. It's not yet clear to me how to execute the `except*` block, but then raise an arbitrary exception group object right afterwards.

nuitka/nodes/ExceptionNodes.py
```diff
@@ -334,6 +334,33 @@ def __init__(self, source_ref):
...
         return self, None, None
 
+class ExpressionCaughtExceptionGroupMatch(ExpressionBase):
+    kind = "EXPRESSION_CAUGHT_EXCEPTION_GROUP_MATCH"
+
+    __slots__ = ("caught", "catching")
+
+    def __init__(self, caught, catching, source_ref):
+        ExpressionBase.__init__(self, source_ref)
+        self.caught = caught
```
kayhayen: The way you announce children to Nuitka, is wrong here, that's probably all your remaining issues.

nuitka/tree/ReformulationTryExceptStatements.py
```diff
@@ -201,43 +201,54 @@ def makeTryExceptSingleHandlerNodeWithPublish(
         source_ref=source_ref,
     )
 
+def fix_nested_sequence(statement_sequence, source_ref):
+    # XXX: This is a hack! It seems that we can't have a StatementsSequence()
+    # nested directly under a StatementsSequence(), so we wrap it in an "if True"
```
kayhayen, line 213: There is never a need to have them, that's why they are prevented. There is `mergeStatements` if you want to pass a statements sequence, and a statement to and get a tuple which can be used to create a statements sequence.

ZeroIntensity: Well, there's a need to have them when you factor things out into smaller helpers like I've done. I'll switch to `mergeStatements`, but it's a little inconvenient; couldn't Nuitka just flatten out the statements anyway?

ZeroIntensity:
Superseded by #3743

PR #3541, opened 2025-07-07
Fix #3100 Regression

### LINKED ISSUE #3100: `pygame-ce` still broken on Nuitka 2.4.8
#2947 was supposed to fix the missing DLL problems that pygame/pygame-ce had with Nuitka, but version 2.4.8 still has those missing DLLs. I noticed that the actual DLL part of the pull request is no longer present in the config file, which is probably the cause of the regression from my pull request.

[reproducer.zip](https://github.com/user-attachments/files/16790297/reproducer.zip)
[verbose nuitka build output](https://github.com/user-attachments/files/16790298/nuitka-output.txt)

build command:
```
python -m nuitka --standalone main.py --include-data-files=./alien1.png=alien1.png --include-data-files=./alien1.jpg=alien1.jpg --include-data-files=./Roboto-Regular.ttf=Roboto-Regular.ttf --include-data-files=./boom.wav=boom.wav --include-data-files=./MIDI_sample.mid=MIDI_sample.mid
```

```
❯ python -m nuitka --version                                                                                                                                                                        
2.4.8
Commercial: None
Python: 3.12.4 (tags/v3.12.4:8e8a4ba, Jun  6 2024, 19:30:16) [MSC v.1940 64 bit (AMD64)]
Flavor: CPython Official
Executable: C:\Users\andre\Downloads\reproducer\venv\Scripts\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 11
Version C compiler: C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.40.33807\bin\Hostx64\x64\cl.exe (cl 14.3).
```

```
(venv) ~\Downloads\reproducer
❯ cd main.dist 
(venv) ~\Downloads\reproducer\main.dist
❯ .\main.exe
pygame-ce 2.5.1 (SDL 2.30.6, Python 3.12.4)
  File "C:\Users\andre\DOWNLO~1\REPROD~1\MAIN~1.DIS\main.py", line 37, in <module>
  File "C:\Users\andre\DOWNLO~1\REPROD~1\MAIN~1.DIS\main.py", line 23, in __init__
pygame.error: Failed loading libpng16-16.dll: The specified module could not be found.
```

oddbookworm:

# What does this PR do?
#3100 has regressed and both pygame and pygame-ce are broken due to missing dlls at runtime

#3100  #2947 fixed it at one point, but it appears the dll fix was removed at some point. I can assure you, the minimal set of dlls are already included in the library. We fight wheel size constantly, so if we don't absolutely need it, we don't have it. I don't know why nuitka isn't detecting runtime dll usage correctly, but we haven't been able to recommend nuitka for over a year to users because it hasn't been able to be used at all without manually copying dlls. User shouldn't have to go through that effort, especially when there are other tools that don't have that problem and it's more seamless for the user. In the case of pygame(-ce), it's just safest to include all the dlls. Yeah, in the absolute minimum examples, you'll have some bloat (when you don't load any images, you don't need libpng), but that's the literal best case scenario. Real use-cases use pretty much all of the dlls we include in our wheels.

# PR Checklist

## Summary by Sourcery

Fix the regression from #3100 by re-adding missing DLL inclusion for the pygame plugin and updating its checksum.

Bug Fixes:
- Restore missing DLL inclusion in the pygame plugin configuration to prevent runtime failures.

Enhancements:
- Update the pygame module checksum to match the revised plugin configuration.

### DIFF
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -4686,7 +4686,7 @@
     - dirs:
         - '.'
 
-- module-name: 'pygame' # checksum: 64efeec5
+- module-name: 'pygame' # checksum: db7d5ddb
   data-files:
     - patterns:
         - 'freesansbold.ttf'
@@ -4696,6 +4696,11 @@
     - patterns:
         - 'pygame_icon.bmp'
       when: 'not macos'
+  dlls:
+    - from_filenames:
+        relative_path: '.'
+        prefixes:
+          - ''
 
   anti-bloat:
     - no-auto-follow:


kayhayen:
Sorry, I messed up again. I think we need to also somehow make it clear if configurations are maintained upstream.

PR #3561, opened 2025-08-05
Fix PEP 695 generic types

ZeroIntensity:

# What does this PR do?

# PR Checklist

## Summary by Sourcery

Fix support for PEP 695 generics including TypeVar default values and variadic type variable unpacking, update generic alias handling, and adjust class node rewriting to clone and pass type parameters via temporaries.

Enhancements:
- Clone type parameter expressions instead of reusing nodes and assign them via temporary variables during class creation
- Add default_value and evaluate_default fields for TypeVar objects and switch to using types.GenericAlias for PEP 695
- Implement recursive unpacking of TypeVarTuple parameters in MAKE_TYPE_GENERIC and return the result wrapped in a tuple
- Minor lint fix in Progress download bar attribute assignment

### DIFF
--- a/nuitka/build/static_src/HelpersTypes.c
+++ b/nuitka/build/static_src/HelpersTypes.c
@@ -347,7 +347,10 @@ PyObject *MAKE_TYPE_GENERIC(PyThreadState *tstate, PyObject *params) {
 
     PyObject *result = CALL_FUNCTION_WITH_ARGS2(tstate, called, args);
     Py_DECREF(unpacked_params);
-    return result;
+    PyObject *tuple = PyTuple_New(1);
+    CHECK_OBJECT(tuple);
+    PyTuple_SET_ITEM(tuple, 1, result);
+    return tuple;
 }
 
 #endif
--- a/nuitka/build/static_src/HelpersTypes.c
+++ b/nuitka/build/static_src/HelpersTypes.c
@@ -354,11 +354,9 @@ PyObject *MAKE_TYPE_GENERIC(PyThreadState *tstate, PyObject *params) {
     PyObject *called = (PyObject *)_getTypeGenericAliasType();
 
     PyObject *result = CALL_FUNCTION_WITH_ARGS2(tstate, called, args);
-    _PyObject_Dump(result);
     Py_DECREF(unpacked_params);
     PyObject *tuple = PyTuple_New(1);
     CHECK_OBJECT(tuple);
-    //_PyObject_Dump(result);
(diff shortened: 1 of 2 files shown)


nuitka/build/static_src/HelpersTypes.c
```diff
@@ -280,36 +288,73 @@ static PyTypeObject *_getTypeGenericAliasType(void) {
 
     if (type_generic_alias_type == NULL) {
 
-        PyObject *typing_module = PyImport_ImportModule("_typing");
-        CHECK_OBJECT(typing_module);
+        PyObject *types_module = PyImport_ImportModule("types");
```
kayhayen: For most of these cases, there is something like `IMPORT_HARD_TYPES` that is cached and all, and preferable, changing that over.

nuitka/build/static_src/HelpersTypes.c
```diff
@@ -280,36 +288,73 @@ static PyTypeObject *_getTypeGenericAliasType(void) {
...
 static PyObject *_Nuitka_unpack_typevartuples(PyObject *params) {
     assert(PyTuple_Check(params));
-
-    // TODO: Not implemented yet.
-
-    return Py_NewRef(params);
+    // TypeVarTuple must be unpacked when passed to Generic, so we do that here.
+    if (_Nuitka_contains_typevartuple((PyTupleObject *)params)) {
+        Py_ssize_t n = PyTuple_GET_SIZE(params);
+        PyObject *new_params = PyTuple_New(n);
```
kayhayen: We have a function that is not API and creates a tuple with a size, using that instead. and also passing tstate in, as we have it on the call side, and getting it can only be expensive.

nuitka/build/static_src/HelpersTypes.c
```diff
@@ -280,36 +288,73 @@ static PyTypeObject *_getTypeGenericAliasType(void) {
...
+    assert(PyTuple_CheckExact(unpacked_params));
 
     PyObject *args[2] = {(PyObject *)tstate->interp->cached_objects.generic_type, unpacked_params};
 
     PyObject *called = (PyObject *)_getTypeGenericAliasType();
 
     PyObject *result = CALL_FUNCTION_WITH_ARGS2(tstate, called, args);
     Py_DECREF(unpacked_params);
-    return result;
+    PyObject *tuple = PyTuple_New(1);
```
kayhayen: We also have a function for create a tuple with a single value, one for with and one for without reference count.

PR #3565, opened 2025-08-10
Update `tkinterweb` package and dependencies to API version 4.x.x

AppSolves:
# What does this PR do?

The implementation of [`tkinterweb`](https://github.com/Andereoo/TkinterWeb) in the [standard package config](https://github.com/Nuitka/Nuitka/blob/develop/nuitka/plugins/standard/standard.nuitka-package.config.yml) was deprecated, having last been modified in 2023. Since then, the tkinterweb API has been [updated to version 4.x.x](https://tkinterweb.readthedocs.io/en/latest/upgrading.html), and the original package was split into two separate packages. This PR updates the configuration accordingly to reflect these changes and ensure compatibility with the latest API.

The Nuitka-generated standalone executable failed at runtime due to a missing path: `<standalone_root>/tkinterweb_tkhtml/tkhtml`. This indicated that the existing plugin configuration was outdated and incompatible with the new package structure, necessitating this fix.

# PR Checklist

## Summary by Sourcery

Adapt the Nuitka standard plugin configuration to support the updated tkinterweb API v4.x.x by reflecting its new split package structure and resource layout.

Bug Fixes:
- Fix missing 'tkhtml' path by adding implicit import of the new tkinterweb_tkhtml package

Enhancements:
- Add separate packaging rules for tkinterweb.utilities with updated ROOT_DIR resolution
- Add new module configuration for tkinterweb_tkhtml including data-files, unified DLL loading on Windows, and anti-bloat replacements

### DIFF
--- a/nuitka/plugins/standard/standard.nuitka-package.config.yml
+++ b/nuitka/plugins/standard/standard.nuitka-package.config.yml
@@ -7972,14 +7972,37 @@
           - ''
       when: 'use_tkinter and macos and arch_amd64 and version("tkinterdnd2") >= (0,4)'
 
-- module-name: 'tkinterweb' # checksum: 2aa01e7a
+- module-name: 'tkinterweb' # checksum: 3f91ac54
   data-files:
     - dirs:
+        - 'tkhtml'
         - 'resources'
 
+  dlls:
+    - from_filenames:
+        relative_path: 'tkhtml/Windows/32-bit'
+        prefixes:
+          - 'Tkhtml'
+      when: 'win32 and arch_x86'
+    - from_filenames:
+        relative_path: 'tkhtml/Windows/64-bit'
+        prefixes:
+          - 'Tkhtml'
+      when: 'win32 and arch_amd64'
+
+  anti-bloat:
+    - description: 'remove "sys.path" hack'
+      replacements_plain:
+        'sys.path.append': ''
   implicit-imports:
     - depends:
         - 'tkinterweb_tkhtml'
+
+  import-hacks:
+    - global-sys-path:
+        # This package forces itself into "sys.path" and expects absolute
+        # imports to be available.
+        - ''
 
 - module-name: 'tkinterweb.utilities' # checksum: 575a76dc
   anti-bloat:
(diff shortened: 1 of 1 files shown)


kayhayen (changes requested):
When new versions appear, config needs to do both, depending on version checks in a when clause maybe, but I can take it from here and do that myself

kayhayen:
I tried to merge old and new codes, can you confirm or deny if this works for you @AppSolves and sorry for the delay, my health has reduced my time available recently

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -7975,12 +7976,42 @@
...
+    - dirs:
+        - 'tkhtml'
+  dlls:
+    - from_filenames:
+        relative_path: 'tkhtml'
+        prefixes:
+          - 'lib'
+        suffixes:
+          - 'dll'
+      when: 'win32 and (arch_x86 or arch_amd64)'
```
kayhayen: Do we have to limit arches here really?

nuitka/plugins/standard/standard.nuitka-package.config.yml
```diff
@@ -7975,12 +7976,42 @@
...
+  data-files:
+    - dirs:
+        - 'tkhtml'
+  dlls:
+    - from_filenames:
+        relative_path: 'tkhtml'
+        prefixes:
+          - 'lib'
+        suffixes:
+          - 'dll'
```
kayhayen: The suffix "dll" would be default on Windows anyway, and need not be specified

AppSolves:
Thanks for the review @kayhayen ! I'm currently on vacation, so I apologize for being a bit inactive lately. Would you mind if I take a look at the changes this evening and get back to you then?

kayhayen:
I'll just take the risk and release it like that then, looks good to me, if it fails to work, we can hotfix it into 2.8 still

nuitka/plugins/standard/standard.nuitka-package.config.yml
kayhayen: Removed that myself

AppSolves:
@kayhayen Alright, so I took a quick look at it and it seems to be working fine 👍 Thank you very much!

Issue #3583, opened 2025-08-25
Missing msvcp140.dll when using packages depending on VC++ runtime (e.g. ujson, pyclipper) in standalone mode


timminator:
## 🐛 Bug Description

When compiling with --standalone, Nuitka does not include msvcp140.dll in the output folder if an extension module depends on it.
On a clean Windows install (without VC++ Redistributables), this causes the executable to fail with the error reported further down.
I also loaded the ujson.pyd and the _pyclipper.pyd file into dependencies - msvcp140.dll is listed as a dependency for both of them but is not included in the dist folder.

### 1. Nuitka Version, Python Version, OS, and Platform

```
2.7.12
Commercial: None
Python: 3.12.8 (tags/v3.12.8:2dc476b, Dec  3 2024, 19:30:04) [MSC v.1942 64 bit (AMD64)]
Flavor: CPython Official
Executable: ~\Desktop\New folder (23)\venv-3.12\Scripts\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 11
Version C compiler: C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.42.34433\bin\Hostx64\x64\cl.exe (cl 14.3).
```

### 2. How Nuitka and Python were Installed

Nuitka was installed via pip, python via the standard windows installer from python.org

### 3. Relevant PyPI Packages and Versions

Via pip:
```
pyclipper   1.3.0.post6 C:\Users\GPUVM\Desktop\New folder (23)\venv-3.12\Lib\site-packages pip
ujson       5.11.0      C:\Users\GPUVM\Desktop\New folder (23)\venv-3.12\Lib\site-packages pip
```

Via nuitka:
```
pyclipper 1.3.0.post6 pip
ujson 5.11.0 pip
```

### 2. Short, Self-Contained, Correct, Eligible (SSCCE) Example

```python
# nuitka-project: --standalone

import ujson
print(ujson.dumps({"hello": "world"}))
```

## 📉 Expected Behavior

Application should work on a different windows system without VC++ Redistributables in standalone mode.

## 📄 Actual Behavior & Output

Application crashes with the error:
```
PS C:\Users\GPUVM\Downloads\test_ujson.dist> .\test_ujson.exe
Traceback (most recent call last):
  File "C:\Users\GPUVM\DOWNLO~1\TEST_U~1.DIS\test_ujson.py", line 3, in <module>
[...]

PR #3587, opened 2025-08-27
Windows: fix DLL search path selection in _attemptToFindNotFoundDLL

timminator:
# What does this PR do?

Corrects the directory choice for DLL lookup on Windows: System32 for x86_64, SysWOW64 for x86.

While investigating #3583 I noticed that this was inverted.

# PR Checklist

## Summary by Sourcery

Bug Fixes:
- Use System32 for x86_64 and SysWOW64 for x86 in _attemptToFindNotFoundDLL on Windows instead of the reverse mapping

### DIFF
--- a/nuitka/freezer/DependsExe.py
+++ b/nuitka/freezer/DependsExe.py
@@ -72,7 +72,7 @@ def _attemptToFindNotFoundDLL(dll_filename):
     # Lets try the Windows system, spell-checker: ignore systemroot
     dll_filename = os.path.join(
         os.environ["SYSTEMROOT"],
-        "SysWOW64" if getArchitecture() == "x86_64" else "System32",
+        "System32" if getArchitecture() == "x86_64" else "SysWOW64",
         dll_filename,
     )
     dll_filename = os.path.normcase(dll_filename)


kayhayen:
I am confused and propably expose a knowledge gap here, can you explain some more about this. Maybe we need to check two directories on x64, but this would have been wrong for a long time indeed. Also it doesn't change anything about the issue referenced, since that DLL still gets ignored, right?

Good on you for sharing the observation of yours.

kayhayen:
After investigating, `System32` would be different depending on the architecture in the correct way already, so I can simplify your code further, to just use that.

kayhayen (approved):
Good catch

timminator:
> I am confused and propably expose a knowledge gap here, can you explain some more about this. Maybe we need to check two directories on x64, but this would have been wrong for a long time indeed. Also it doesn't change anything about the issue referenced, since that DLL still gets ignored, right?
> 
> Good on you for sharing the observation of yours.


This being wrong for a long time seemed to have caused no issues because this function is only run when dependencies of the pythonxx.dll can not be resolved. But on Windows for newer python versions as far as i could see all dependencies can be resolved so this code was never triggered. :-)
This fix is indeed not causing the dll to be included, I just noticed it while I was tinkering with the code. Nuitka is running dependencyWalker without path lookups, so msvcp140.dll or vcomp140.dll cannot be resolved by it and are marked with a question mark, so I added them like this:
```python
        if "?" in line[: line.find("]")]:
            if (
                dll_filename.startswith("python") 
                or dll_filename.startswith("msvcp") 
                or dll_filename.startswith("vcomp")
            ) and dll_filename.endswith(".dll"):
                print("test")
                print(dll_filename)
                dll_filename = _attemptToFindNotFoundDLL(dll_filename)
                print(dll_filename)
                if dll_filename is None:
                    continue
            else:
                continue
```

and in the standalone.py script I made a simple exception:
```python
    for dll_filename in used_dll_paths:
        dll_basename = os.path.basename(dll_filename)
        print(dll_basename)
        if decideInside(dll_filename) or dll_basename == "msvcp140.dll":
            kept_used_dll_paths.add(dll_filename)
```

and the dll was now included in the final dist folder. But then I noticed that my test script still did not work because nuitka copied the 32bit version of the msvcp140.dll from SysWOW64 even though I have python in 64 bit mode installed.

That's how I noticed this mistake and how this PR was made. :-)

That's of course not a valid solution as these dlls are not allowed to be copied. They would need to be copied from the MSVC redist folder if Microsoft Visual Studio is installed. I was just tinkering around. :-)

Thanks for merging.

PR #3590, opened 2025-08-28
Support arbitrary debuggers through `--debugger`

ZeroIntensity:
# What does this PR do?

The `--debugger` flag can now be used with something like `--debugger=...` to tell Nuitka to choose an arbitrary executable to run (for example, `--debugger=valgrind` or `--debugger="rr record"`). The existing behavior with `--debugger` without an `=` is retained; GDB is still automatically chosen.

It was difficult to use valgrind when debugging Nuitka. This will improve my productivity when debugging crashes.

# PR Checklist

## Summary by Sourcery

Enable support for arbitrary debugger commands via the existing --debugger flag

New Features:
- Allow specifying any debugger executable and arguments using --debugger=<cmd>

Enhancements:
- Preserve implicit "gdb" default when --debugger is used without a value
- Enhance wrapCommandForDebuggerForExec to resolve and invoke custom debugger paths and error if not found
- Refactor option parsing to support implicit default values and add getDebuggerName to reflect new debugger semantics

### DIFF
--- a/nuitka/MainControl.py
+++ b/nuitka/MainControl.py
@@ -792,7 +792,9 @@ def _executeMain(binary_filename):
     if Options.shallRunInDebugger() and not Options.shallCreateScriptFileForExecution():
         # The path needs to be absolute for some debuggers to work e.g. valgrind
         absolute_binary_filename = os.path.abspath(binary_filename)
-        args = wrapCommandForDebuggerForExec(command=(absolute_binary_filename,), debugger=Options.getDebuggerName())
+        args = wrapCommandForDebuggerForExec(
+            command=(absolute_binary_filename,), debugger=Options.getDebuggerName()
+        )
     else:
         args = (binary_filename, binary_filename)
 
@@ -843,7 +845,7 @@ import sys; sys.path.insert(0, %(output_dir)r)
     if Options.shallRunInDebugger():
         args = wrapCommandForDebuggerForExec(
             command=(sys.executable, "-c", python_command),
-            debugger=Options.getDebuggerName()
+            debugger=Options.getDebuggerName(),
         )
     else:
         args = (sys.executable, "python", "-c", python_command)
--- a/nuitka/MainControl.py
+++ b/nuitka/MainControl.py
@@ -806,10 +806,8 @@ def callExecPython(args, add_path, uac):
 def _executeMain(binary_filename):
     # Wrap in debugger, unless the CMD file contains that call already.
     if Options.shallRunInDebugger() and not Options.shallCreateScriptFileForExecution():
-        # The path needs to be absolute for some debuggers to work e.g. valgrind
-        absolute_binary_filename = os.path.abspath(binary_filename)
         args = wrapCommandForDebuggerForExec(
-            command=(absolute_binary_filename,), debugger=Options.getDebuggerName()
+            command=(binary_filename,), debugger=Options.getDebuggerName()
         )
     else:
         args = (binary_filename, binary_filename)
--- a/nuitka/MainControl.py
+++ b/nuitka/MainControl.py
@@ -807,7 +807,8 @@ def _executeMain(binary_filename):
     # Wrap in debugger, unless the CMD file contains that call already.
     if Options.shallRunInDebugger() and not Options.shallCreateScriptFileForExecution():
         args = wrapCommandForDebuggerForExec(
-            command=(binary_filename,), debugger=Options.getDebuggerName()
+            command=(binary_filename,),
+            debugger=Options.getDebuggerName(),
         )
     else:
         args = (binary_filename, binary_filename)
--- a/nuitka/MainControl.py
+++ b/nuitka/MainControl.py
@@ -790,7 +790,9 @@ def callExecPython(args, add_path, uac):
 def _executeMain(binary_filename):
     # Wrap in debugger, unless the CMD file contains that call already.
     if Options.shallRunInDebugger() and not Options.shallCreateScriptFileForExecution():
-        args = wrapCommandForDebuggerForExec(command=(binary_filename,))
+        # The path needs to be absolute for some debuggers to work e.g. valgrind
+        absolute_binary_filename = os.path.abspath(binary_filename)
+        args = wrapCommandForDebuggerForExec(command=(absolute_binary_filename,), debugger=Options.getDebuggerName())
     else:
         args = (binary_filename, binary_filename)
 
@@ -840,7 +842,8 @@ import sys; sys.path.insert(0, %(output_dir)r)
 
     if Options.shallRunInDebugger():
         args = wrapCommandForDebuggerForExec(
-            command=(sys.executable, "-c", python_command)
+            command=(sys.executable, "-c", python_command),
+            debugger=Options.getDebuggerName()
         )
     else:
         args = (sys.executable, "python", "-c", python_command)
--- a/nuitka/MainControl.py
+++ b/nuitka/MainControl.py
@@ -332,9 +332,18 @@ def pickSourceFilenames(source_dir, modules):
         # or not, but that's probably not worth the effort. False positives do
         # no harm at all. We cannot use normcase, as macOS is not using one that
         # will tell us the truth.
-        collision_filename = base_filename.lower()
+        collision_filename = os.path.join(
+            source_dir, "module." + module.getFullName().asString().lower()
+        )
 
-        return base_filename, collision_filename
+        # When the filename becomes to long to add ".const", we use a hash name
+        # instead.
+        hash_filename = os.path.join(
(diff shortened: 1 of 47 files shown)


nuitka/Options.py
```diff
@@ -1467,6 +1467,11 @@ def shallExecuteImmediately():
 
 def shallRunInDebugger():
     """:returns: bool derived from ``--debug``"""
+    return options.debugger is not None
+
```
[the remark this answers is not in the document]
ZeroIntensity: Silly robot, the point is that it supports *arbitrary* debuggers, not ones from a preset list. We can always add special-casing to improve support for our favorite debuggers later, but generally we just want a way to use any executable.

nuitka/MainControl.py
```diff
@@ -840,7 +844,8 @@ def _executeModule(tree):
 
     if Options.shallRunInDebugger():
         args = wrapCommandForDebuggerForExec(
-            command=(sys.executable, "-c", python_command)
+            command=(sys.executable, "-c", python_command),
+            debugger=Options.getDebuggerName(),
```
kayhayen, line 863: That was apparently missing, good find

nuitka/MainControl.py
```diff
@@ -790,7 +790,11 @@ def callExecPython(args, add_path, uac):
 def _executeMain(binary_filename):
     # Wrap in debugger, unless the CMD file contains that call already.
     if Options.shallRunInDebugger() and not Options.shallCreateScriptFileForExecution():
-        args = wrapCommandForDebuggerForExec(command=(binary_filename,))
+        # The path needs to be absolute for some debuggers to work e.g. valgrind
```
kayhayen: If that's really necessary, that's something `wrapCommandForDebuggerForExec` ought to be doing, clearly doesn't belong into main control.

nuitka/OptionParsing.py
```diff
@@ -667,14 +667,14 @@ def _handleHelpModes():
...
+    "--gdb",  # Remove this now that this supports any arbitrary debugger?
     dest="debugger",
-    default=False,
+    default=None,
     help="""\
 Execute inside a debugger, e.g. "gdb" or "lldb" to automatically get a stack trace. The
 debugger is automatically chosen unless specified by name with the NUITKA_DEBUGGER_CHOICE
-environment variable. Defaults to off.""",
+environment variable or this flag. Defaults to off.""",
+    implicit_default="gdb",
```
kayhayen: So what you did here was to make `--debugger` have an ability to accept arguments optionally? I am not so sure I like that idea, and it's not documented here how to do it. The argparse also doesn't have that, right? I think that makes descriptions very convoluted. And other formats like toml, yaml, etc. which feed Nuitka options loose the ability to confirm to any schema, which might be a deal breaker (although on gitHub that option --debugger wouldn't be present, and it also not make sense in pyproject.toml (much like --run), but I don't like this feature for the reason that it will break that.

And I couldn't say `--debugger=lldb` without repeating Nuitka code to make the automatic traceback, right? I added the option to be able to forget those.

I am not so sure if I wouldn't always prefer a separate options to provide those things, like `--debugger-args` and want sane defaults to be commony used, even and esp. for valgrind memcheck. And complement that with `--debugger-choice` which replaces `NUITKA_DEBUGGER_CHOICE` where then valigrind is allowed as well.

nuitka/build/SconsUtils.py
```diff
@@ -21,8 +21,8 @@
 from nuitka.utils.Execution import executeProcess
 from nuitka.utils.FileOperations import (
     changeFilenameExtension,
-    getExternalUsePath,
     deleteFile,
+    getExternalUsePath,
```
kayhayen: You should rebase to current develop, this shouldn't be there unless I did something stupid to it.

kayhayen:
I don't even care about `__main__.py: error: --version option does not take a value` errors by CI, please remove those parts that require a changed command line parsing unless you have a very convincing case with many other (good) programs doing similar.

ZeroIntensity:
> The implicit default thing is something we should have discussed before impelementing it, as it's pretty invasive to not know if an option has an argument or not.

I thought I mentioned that this was my approach, sorry!

I'm fine with changing this, I won't bother arguing.

nuitka/OptionParsing.py
ZeroIntensity: Yeah, that all makes sense. I was trying to keep things consolidated, but I can just move the logic to a `--debugger-choice` flag.

Are you okay with the design of supporting arbitrary debuggers, or would you like to add special-casing for valgrind in `--debugger`?

nuitka/PostProcessing.py
```diff
@@ -226,7 +226,7 @@ def createScriptFileForExecution(result_filename):
     python_path = os.pathsep.join(makeFilesystemEncodable(e) for e in sys.path)
 
     debugger_call = (
-        (" ".join(wrapCommandForDebuggerForExec(command=())) + " ")
+        (" ".join(wrapCommandForDebuggerForExec(command=())[1:]) + " ")
```
kayhayen: I believe that's an infererior version of `wrapCommandForDebuggerForSubprocess`  as the first one is the binary, the second one is the process name, and that's what shell and subprocess don't have.

nuitka/utils/Execution.py
```diff
@@ -275,10 +277,19 @@ def wrapCommandForDebuggerForExec(command, debugger=None):
 
     # Default from environment variable.
     if debugger is None:
-        debugger = os.getenv("NUITKA_DEBUGGER_CHOICE")
+        debugger = Options.getDebuggerName()
 
     if debugger not in ("gdb", "lldb", None):
-        general.sysexit("Error, the selected debugger name '%s' is not supported.")
+        # We don't know how to do anything special for this debugger -- just
```
kayhayen, line 282: I still dislike that we don't have a `valgrind-memcheck` or so, where the typical options to make it launch and quit are not auto-provided. Can you add that, and still leave the other thing in there, for flexility. But oommonly used debugger tools ought to be trivially recognizable and just to be provided by name, so I don't have to look up how to do it.

nuitka/PostProcessing.py
ZeroIntensity: Yeah, but interestingly, GDB seems to be fine with running `gdb gdb whatever.bin` in the terminal. Valgrind isn't as relaxed, so it gives an error.

ZeroIntensity: Oh, I may have misunderstood your original comment. Should I switch to `wrapCommandForDebuggerForSubprocess` here?

nuitka/utils/Execution.py
ZeroIntensity: Done.

nuitka/PostProcessing.py
kayhayen: Yeah, it gives you a better form, just checkout the code for `wrapCommandForDebuggerForSubprocess` to understand what it's doing better.

nuitka/MainControl.py
kayhayen: Can you still move the code to the wrapping function? That's the other thing holding me back from merging it, as I do not want to pollute MainControl and think it may only be beneficial to get that right in more code paths.

kayhayen:
Also somehow the debugger choice default seems to be bad, or why else why are th checks running lldb

nuitka/MainControl.py
ZeroIntensity: Sorry, forgot to do this one. I moved the logic.

ZeroIntensity:
Hm, I think I fixed it, but I'm not sure why this wasn't being picked up before. As far as I can tell, lldb was already the default.

ZeroIntensity:
Nevermind, that didn't fix it.

kayhayen:
I fixed the issue that was causing lldb to be used. I usually like to use `None` to distinguish not given from `False` which would be user provided just like `True` would be. The `shallRunInDebugger` is comparing to `None` for the result, so setting it to false wouldn't work either. Your reverting of the unwanted option changes probably was incomplete. That said, I will probably change said function to also accept `False`, to allow a --no-debugger option to eventually work easily as well.

ZeroIntensity:
Ok, that makes sense, thanks.

kayhayen:
So, that's merged and hopefully useful in the future. If you find yourself using a lot of the same --debugger stuff added, lets see how to expand it. I feel like the control flow has become pretty bad there, and we probably need to refactor it if we add another one surely.

ZeroIntensity:
Yeah, I thought about refactoring when adding Valgrind support, but I held back. The arbitrary `--debugger-choice` should be pretty helpful for unsupported debuggers, though.

kayhayen:
I fully agree with that. I just don't want it to be the norm.

PR #3591, opened 2025-09-01
Add support for including vc runtime dlls

### LINKED ISSUE #3583: Missing msvcp140.dll when using packages depending on VC++ runtime (e.g. ujson, pyclipper) in standalone mode

timminator:
# What does this PR do?

This PR allows Nuitka to also include dlls from VC++ runtime redistributables in the final standalone folder that are not guaranteed to be installed on every Windows System.

I am interested in your feedback.

Fixes #3583 

# PR Checklist

## Summary by Sourcery

Include missing Visual C++ runtime redistributable DLLs in standalone builds by discovering the MSVC Redist installation via vswhere and searching for known redistributable DLLs.

New Features:
- Discover the VC++ Redistributable installation path on Windows using vswhere
- Automatically search for and include known VCRedist DLLs (e.g., vcruntime140.dll) in the standalone output
- Retain redistributable DLLs in standalone builds even if they lie outside the Python search path

Enhancements:
- Add a helper function to recursively search for DLLs in the redistributable directory
- Emit a warning when a required VCRedist DLL cannot be found due to missing Visual Studio installation

### DIFF
--- a/nuitka/utils/Utils.py
+++ b/nuitka/utils/Utils.py
@@ -512,6 +512,8 @@ _msvc_redist_path = None
 
 
 def _getMSVCRedistPath(logger):
+    # The detection is return driven with many cases to look at
+    # pylint: disable=too-many-return-statements
 
     # TODO: We could try and export the vswhere information from what Scons
     # found out, to avoid the (then duplicated) vswhere call entirely.
@@ -597,7 +599,7 @@ def _getMSVCRedistPath(logger):
 
 
 def getMSVCRedistPath(logger):
-    global _msvc_redist_path
+    global _msvc_redist_path  # singleton, pylint: disable=global-statement
 
     if _msvc_redist_path is None:
         if isWin32Windows():


nuitka/freezer/DependsExe.py
```diff
@@ -59,26 +59,40 @@ def getDependsExePath():
...
+        dll_filename = os.path.normcase(dll_filename)
+
+        if os.path.exists(dll_filename):
+            return dll_filename
 
-    if os.path.exists(dll_filename):
-        return dll_filename
+    if is_vcredist_dll and VCREDIST_PATH:
+        # Lets attempt to find it in the MSVC Redistributable folder
+        return _find_dll_in_redist(VCREDIST_PATH, dll_filename)
```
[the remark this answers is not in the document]
timminator: It's just the basename already...

nuitka/freezer/DependsExe.py
```diff
@@ -118,14 +132,44 @@ def _parseDependsExeOutput2(lines):
...
+            "vcruntime140_threads.dll",
+            "vcamp140.dll",
+            "vcomp140.dll",
+        }
+
         if "?" in line[: line.find("]")]:
-            # One exception are "PythonXY.DLL", we try to find them from Windows folder.
-            if dll_filename.startswith("python") and dll_filename.endswith(".dll"):
-                dll_filename = _attemptToFindNotFoundDLL(dll_filename)
+            is_vcredist_dll = os.path.basename(dll_filename) in VCREDIST_DLLS
```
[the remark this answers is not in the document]
timminator: all VC++ Runtime redistributables are lowercase and will match the list exactly...

nuitka/utils/Utils.py
```diff
@@ -508,6 +509,81 @@ def isCoffUsingPlatform():
...
+        if not all_folders:
+            return None
+
+        version_folders = [v for v in all_folders if v[0].isdigit()]
+
+        if not version_folders:
+            return None
+
+        latest_version = sorted(
+            version_folders, key=lambda v: list(map(int, v.split(".")))
```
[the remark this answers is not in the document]
timminator: As far as im aware, this folder only contains folders with the Redistributables version name or folders with the name of the MSVC compiler version (e.g. v143) which is checked already above these lines of code.

nuitka/freezer/Standalone.py
```diff
@@ -343,7 +344,13 @@ def decideInside(dll_filename):
...
 
     for dll_filename in used_dll_paths:
-        if decideInside(dll_filename):
+        is_from_vc_redist = False
+        if VCREDIST_PATH:
+            is_from_vc_redist = isFilenameBelowPath(
+                path=VCREDIST_PATH, filename=dll_filename
+            )
+
+        if decideInside(dll_filename) or is_from_vc_redist:
```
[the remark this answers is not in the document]
timminator: They cannot be in the Python path...

nuitka/utils/Utils.py
```diff
@@ -508,6 +509,81 @@ def isCoffUsingPlatform():
...
+        command = [
+            vswhere_path,
+            "-latest",
+            "-property",
+            "installationPath",
+            "-products",
+            "*",
+            "-prerelease",
+        ]
+        vs_path = subprocess.check_output(command).decode("utf8").strip()
```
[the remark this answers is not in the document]
timminator: The command is not built on user input it is hardcoded right above it. We are also not using shell=True.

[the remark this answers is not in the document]
timminator: The command is not built on user input it is hardcoded right above it. We are also not using shell=True.

timminator:
@sourcery-ai review

timminator:
I expected a rereview to consider the comments on the resolved issues instead of reraising all of them again... Not sure how to stop it from requesting changes then.
@sourcery-ai dismiss

nuitka/freezer/DependsExe.py
```diff
@@ -118,14 +132,44 @@ def _parseDependsExeOutput2(lines):
...
+            is_python_dll = dll_filename.startswith("python") and dll_filename.endswith(
```
kayhayen: Just noticing that "python" is not treated the same, i.e. no basename taken. Apparently as a DLL it's usually found either way.

nuitka/freezer/DependsExe.py
```diff
@@ -118,14 +132,44 @@ def _parseDependsExeOutput2(lines):
...
+                    dll_filename, is_python_dll, is_vcredist_dll
+                )
+
+                if found_path:
+                    dll_filename = found_path
+                else:
+                    if is_vcredist_dll:
+                        inclusion_logger.warning(
+                            """\
+                            Required Visual C++ Redistributable DLL '%s' could not be included. \
```
kayhayen: This warning ought to happen only once if at all. If it's not found, we just don't include it, or our code is incorrect. Arguably we are lacking as `warning_once` to achieve that, but I expect to not see these anyway, so what gives.

nuitka/freezer/Standalone.py
```diff
@@ -50,6 +50,7 @@
 from nuitka.utils.Signing import addMacOSCodeSignature
 from nuitka.utils.Timing import TimerReport
 from nuitka.utils.Utils import (
+    VCREDIST_PATH,
```
kayhayen: I would prefer this to be an accessor function and not a global variable.

nuitka/utils/Utils.py
```diff
@@ -508,6 +509,81 @@ def isCoffUsingPlatform():
...
+
+    try:
+        command = [
+            vswhere_path,
+            "-latest",
+            "-property",
+            "installationPath",
+            "-products",
+            "*",
+            "-prerelease",
```
kayhayen: That appears to be not specifying the product, and as such could get multiple paths, I guess. There has to be a proper single path to look for.

nuitka/freezer/DependsExe.py
timminator: Thanks for taking a look, the code `dll_filename.startswith("python") and dll_filename.endswith(` existed already beforehand and was not touched by me. And yes, as mentioned by me in the other PR, this code branch was never taken for the pythonxy.dll in my testing. Maybe this happened in some very old python versions?
The os.path.basename(dll_filename) I added is also not really necessary for the vcredist dlls as dll_filename extracted out of the dependency walker log is always just the basename from my testing. Maybe its a bit overcautious but it cannot hurt I thought.
But in the little helper function _find_dll_in_redist I did not add os.path.basename in front of the dll_filename - so this is maybe a bit inconsistent from my side.

kayhayen: The question that bugs me is why the dependency walker doesn't find it and through what method they enter the picture on the live system.

It would have to suffice to add it as a scan directory I bet.

Dependency Walker would find absolute paths. These DLLs are not. But they ought to be and then ignored later on. Maybe something I did when I years ago made it fast noticing it was scanning many folders, I might have removed those.

nuitka/utils/Utils.py
```diff
@@ -508,6 +509,81 @@ def isCoffUsingPlatform():
     return isAIX()
 
 
+def findVCRedistPath():
+    vswhere_path = os.path.join(
```
kayhayen: There may be no installer on many systems used to compile. I would expect e.g. containers, potentially even GitHub to remove them, so I don't know if vswhere is usable.

Also scons uses it optionally and does this:

```
    # For bug 3333 - support default location of vswhere for both 64 and 32 bit windows
    # installs.
    for pf in ['Program Files (x86)', 'Program Files']:
        vswhere_path = os.path.join(
            'C:\\',
            pf,
            'Microsoft Visual Studio',
            'Installer',
            'vswhere.exe'
        )
        if os.path.exists(vswhere_path):
            # If we found vswhere, then use it.
            break
````

File: nuitka/utils/Utils.py (hunk 9 above)
Comment (timminator) [reply nuitka/utils/Utils.py]: No, this will only return one path. The `latest` flag I added causes it to detect all installed versions and return the one with the highest version number. Tested it also on my system where I have VS 2019 and VS 2022 installed.

File: nuitka/freezer/DependsExe.py (hunk 6 above)
Comment (timminator) [reply nuitka/freezer/DependsExe.py]: DependencyWalker is not finding them because because we are wiping the path and only allowing it to search for dlls in custom directories we add via the scan_dirs variable. System32 is not one of them. So they cannot be resolved. If we run DependencyWalker without these restrictions it can resolve them just fine, so exactly as you thought with your second comment. So you think we should add the MSVC redist part if it exists directly to the scan dirs of DependencyWalker?

File: nuitka/utils/Utils.py (hunk 10 above)
Comment (timminator) [reply nuitka/utils/Utils.py]: The vswhere.exe is always in the Program Files (x86) directory because the installer is only 32bit. So we do not need to check both. Otherwise the scons code seems similar to what I have done.
I understand your concern that vswhere could not be available, but I'm not sure what else we could do. We could check the hardcoded paths where VS Studio is normally installed but that could get messy quickly (e.g. several system drives etc.)...

Edit: Maybe also some registry checks for the install location could be possible, not sure what you would think of that...

File: nuitka/freezer/Standalone.py (hunk 8 above)
Comment (timminator) [reply nuitka/freezer/Standalone.py]: Fixed this in my new commit.

File: nuitka/freezer/DependsExe.py (hunk 7 above)
Comment (timminator) [reply nuitka/freezer/DependsExe.py]: Fixed this in my new commit. The warning is now printed once in the end.

**Comment by timminator** [comment]
I made a new commit now. The MSVC Redist path is added now to the scan directories dependency walker is using, so the dlls are now found right away. In the dependsexe.py script we keep track now of missing VC++ Runtime Redistributables and print one Nuitka inlcusion warning in the end.
This should address all your comments. Only the usage of vswhere is still open. I made a comment about it above. I'm interested in your thoughts on this.

**Comment by kayhayen** [comment]
Yes, I think it's a scan dir or a set of. We also wouldn't have to care about arch, and can let it handle that.

Am I getting you right there, that the vc redist lives on `%PATH%` or inside of there? I guess, I need to actually run your code or the vswhere command. Nuitka has had issues with unwanted DLLs from Tencent software in %PATH%, and as such I added restrictions to not scan outside of Python at all, unless system DLLs are allowed for. Maybe then your path just needs to be exception for that.

I may not find the time to look into this further today as I got a sponsor issue at hand, but it seems to me, that got the building blocks for a solution at least.

File: nuitka/utils/Utils.py (hunk 10 above)
Comment (kayhayen) [reply nuitka/utils/Utils.py]: Pretty sure, registry ought to help us out there. From what I took with reading the scons code only briefly is that vswhere is only an option to them, and that they have over methods. The location code, kind of can steal their ideas freely.

But now that I think of it, maybe when we scan %PATH% and it's there, maybe there is a way to uniquely identify it while doing that, so we don't need to know it ahead of time at all. Obviously the DLLs are going to be present, but maybe also other files that could serve as a hallmark.

Usually I like the best method for detecting it, like registry more. But we enter the business of deciding which directory to go to, etc. and dependency walker does that. I must however not forget about pefile, where we do get to use exact directories for ARM Windows, and that wouldn't work well then, so your architecture selection code better be there I guess, but maybe that's already what is in %PATH%.

File: nuitka/utils/Utils.py (hunk 9 above)
Comment (kayhayen) [reply nuitka/utils/Utils.py]: But what about other products, and I thought we are looking at the installed VC redisributables path, and not Visual Stuiod.

**Comment by kayhayen** [comment]
So, it seems the `msvcp140_atomic_wait.dll` is only in my `C:}Windows}System32` folder, and that's also in my `%PATH%` and that's it. I did find it also from a bunch other folders. So maybe all we need to do is to declare that folder a scan_dir overriding use_path.

```
def _getPathContribution(use_path):
    """Contributions from PATH environment variable.

    These are ignored if use_path is False, but for
    Anaconda we need to keep those elements pointing
    to inside of it.
    """
```

There is an exception for Anaconda paths, and one more can be made for Windows and that system folder (still need to look it up, Nuitka does that somewhere already). 

And when then the result is produced (by any tool), so after this we just need the `_reduceToPythonPath` to allow it on Win32 for the filenames you know, and we would be good?

This of course all hinges on the fact that a VC redist installer has no other installation location than System32 folder.

And one thing not to forget, for the caching of the DLL dependencies we better make sure to consider the option you added.

**Comment by kayhayen** [comment]
I just checked, there is no such option yet, but I think `--windows-include-redist-dlls` could be such an option. I am still pretty sure it's not legal to distribute them. But a lot of people do that, LLVM, Firefox, Android Studio, all big people, and I am not a lawyer.

**Comment by timminator** [comment]
I'm not sure if I can follow you correctly, but we are not allowed to copy them out of the System32 that is not allowed. So we cannot add that as a scan dir. That's why I added the Microsoft Visual Studio detection and added these paths to the scan_dirs. As stated in one of the posts in the mentioned issue we are allowed to use them from there:
https://learn.microsoft.com/en-us/visualstudio/releases/2022/redistribution#visual-c-runtime-files

> Subject to the License Terms for the software, you may copy and distribute with your program any of the files within the following folder and its subfolders except as noted below. You may not modify these files.

> [VisualStudioFolder]\VC\redist

So what you want to do is way easier but definitely not allowed. So I came up with the solution implemented in this PR.

File: nuitka/utils/Utils.py (hunk 10 above)
Comment (timminator) [reply nuitka/utils/Utils.py]: I also took a look at the scons vc.py file. It is quite a big one. From what I see for VS 2017+ they also only use the vswhere way of detecting MSVC. From what I found this also always shipped - it is also present if just the build tools were installed for example.
For older possible installations it falls back to checking the registry. For VS 2017+ checking the registry does no longer work as far as I can see as Microsoft is no longer writing install locations inside the registry.
On both my test system Visual Studio is also not adding any entries to the %PATH% variable so this will also not help us.
I think supporting VS 2017+ should be enough? VSWhere is then our best option.

**Comment by kayhayen** [comment]
You have at least communicated a lot better to me why this is this way. Yeah, letting dependency walker scan system32 cannot really be the solution, it will just cause a lot of noise with DLLs using DLLs there which get resolved for no good reason. But if it was allowed to take them from there, I would go that route. And I sort of was under the belief that redist packages cannot be distributed file wise. But only taking it from there would be doing that.

But your argument there now explains it really well. We are allowed indeed to take it from SDK installations from mine and your reading apparently. So we cannot deploy the used DLLs, but these instead, and then we need to find them. And if we don't, but we use them, then we can inform the user about that.

The question is then, why don't we just use Scons for help. At least with MSVC build, by the time we need to look up the DLLs we have built the binary and have Windows SDK information available, and know the folder to take it from.

**Comment by kayhayen** [comment]
Let me rebase this PR shortly, make sure `git pull --rebase`

**Comment by kayhayen** [comment]
I rebased it to current develop, which should give a better view on the quality checks for which develop had been broken for a while, but now ought to be good.

**Comment by kayhayen** [comment]
I wonder if the Win32 SDK directory wouldn't be available by a call similar to this: ```getSconsReportValue(source_dir=source_dir, key="CCACHE_LOGFILE")```` which is how we lookup values from the scons environment at C compile time.

**Comment by kayhayen** [comment]
I might have added that at some point myself:

```
cat AssertsTest.build/scons-report.txt | grep -i SDK
WindowsSDKVersion=10.0.26100.0
```

The folder this lives in, is not recorded, but maybe I can add that.

**Comment by kayhayen** [comment]
There is at least t his here:

```
VSWHERE=C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe
```

That would let us re-use Scons cleverness at locating a usable binary.

**Comment by kayhayen** [comment]
But I wonder, if we can just work from this:

```
VCINSTALLDIR ['C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\']
```

When MSVC is detected, that value would be available. Right now there is this code in the 4.x inline copy of Scons, that we could change:

```
    for k, v in d.items():
        # Nuitka: Make the Windows SDK version visible in environment.
        if k == "WindowsSDKVersion":
            # Always just a single version if any.
            if len(v) == 1:
                env["WindowsSDKVersion"] = v[0].rstrip('\\')
            elif len(v) == 0:
                env["WindowsSDKVersion"] = None
            else:
                assert False, v

            continue

        print(k,v) # Added that to see what we could publish too
        assert "SDK" not in k, (k,v)

        env.PrependENVPath(k, v, delete_existing=True)
        debug("env['ENV']['%s'] = %s" % (k, env['ENV'][k]))
```

**Comment by kayhayen** [comment]
For MinGW64 and Python3.12, there wouldn't be those things, so we need to keep your code definitely. But we might want to indeed go the registry route.

**Comment by kayhayen** [comment]
It would appear, we have these keys:

For 64-bit systems: HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\Microsoft\Microsoft SDKs\Windows
For 32-bit systems: HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Microsoft SDKs\Windows

Checking below there is a going to give the list of installed SDKs, only one for me, which is a bit surprising, in Visual Studio, I may have more, but "InstallationFolder" will do. For Windows 10, there is a sub-folder, olders ones maybe don't have that. So adding that should be cheap and not require vswhere at all.

**Comment by kayhayen** [comment]
The scons inline copy has code dealing with it:

```
class WindowsSDK(SDKDefinition):
    """
    A subclass for trying to find installed Windows SDK directories.
    """
    HKEY_FMT = r'Software\Microsoft\Microsoft SDKs\Windows\v%s\InstallationFolder'
    def __init__(self, *args, **kw):
        SDKDefinition.__init__(self, *args, **kw)
        self.hkey_data = self.version
```

But it comes down to mapping arch to folders which your code already does, and then it's good. We just need to use registry methods as per `getInstalledPythonRegistryPaths` which makes sure to use stdlib rather than pywin32. It seems easy to do this.

**Comment by kayhayen** [comment]
Let me know if you are willing to do away with vswhere and implement this kind of detection code. It seems we need nothing from scons, but only the folder from registry added to scan_dirs, because then it will find the DLLs, and we can still decide to include it or not when we do the path reduction. Adding it to the PATH would mean, we don't even have to know what files we are talking of. I guess the folder needs to be rather at the end of scan dirs, so it has lower priority that PyPI included runtime libraries, or maybe rather not, maybe higher, to it repairs breaking of sometimes an extension module having it, and another not, making it order dependend if it works from PyPI packages.

File: nuitka/freezer/DependsExe.py (hunk 6 above)
Comment (kayhayen) [reply nuitka/freezer/DependsExe.py]: Yes, see further down, I think we add the proper path from the SDK though, not the redist. Your license comments are fully spot on. Technically I would be correct, legally I almost certainly know the redist is not usable to us.

File: nuitka/utils/Utils.py (hunk 10 above)
Comment (kayhayen) [reply nuitka/utils/Utils.py]: I think we can do with just manual registry search, bonus is that a MinGW64 based compilation works equally well, and we effectively only have to maximize versions over the registry folder, which will be less effort. I can see how it's not correct for MSVC compilation, and it wouldn't be wrong if we used its folders instead, but how much more correct can it be. Lets just depend on an installed Windows SDK, and I kind of assume that my Windows SDK comes from the MSVC installation anyway.

If we don't find them, we can prompt MinGW64 users still to install a Windows SDK, which will be a lot less involved than installing Visual Studio.

**Comment by kayhayen** [comment]
Apart from how the folder is located (subprocess vs winreg), I think this PR is probably good to go.

**Comment by timminator** [comment]
I've checked out the installationfolder linked in the registry on both my test systems. In both cases none of the redist dlls were present in these folders. I searched for example for msvcp140.dll and vcomp140.dll without any results. So I think these SDK's are not sufficient at all... Were the dlls present in these folders on your system?

**Comment by kayhayen** [comment]
So, `10.0.26100.0/ucrt/DLLs/x64` contains the UCRT runtime files, but not the ones we are looking for. Still those are interesting too I guess, and sometimes picked up from `PATH` and should be taken from there I guess if at all. But Python on Windows is married to the other runtime it seems, so even some times these are used as well, the files you are looking for are mandatory.

Going with `vswhere` is therefore probably indispensable indeed and it probably contains those UCRT ones as well. So we can only shortcut by re-using what Scons found, but we always have to call it. And if you don't have Visual Studio at least installed, you cannot do it. Right now for Python3.13 with Nuitka it's mandatory anyway, but I of course want to change that. But what can we do other than allowing people to specify the folder also via an option (where `auto` does use `vswhere`) and then if people want to, they can have those files from where-ever.

File: nuitka/utils/Utils.py (hunk 9 above)
Comment (kayhayen) [reply nuitka/utils/Utils.py]: Scons also uses `*` for product, so I think it's probably the right thing to do.

**Comment by kayhayen** [comment]
Seems I am coming around to merging this as it is now, it seems reasonable to do it like this.

I wonder if for older Python, we would have to use older MSVC installations potentially, but I am choosing to not consider that to matter much now, mostly we need not bother with this. I also wonder about the Windows 7 impact.

**Comment by timminator** [comment]
From what I found out Visual Studio 2017 and 2019 can also be installed on Windows 7. So this PR should also work with Windows 7 in its current state.

File: nuitka/freezer/DllDependenciesWin32.py [hunk 11]
```diff
@@ -209,6 +209,17 @@ def _getScanDirectories(package_name, original_dir, use_path):
 
     scan_dirs = [os.path.dirname(sys.executable), getSystemPrefixPath()]
 
+    # Add the VCRedist path to the list of directories to search if it exists
+    vc_redist_path = getVCRedistPath()
+    if vc_redist_path:
```
kayhayen: We got a function that gets the list of sub directories and handles this. I think it doesn't ignore OSError, but hopefully that would be unnecessary.

nuitka/freezer/Standalone.py
```diff
@@ -51,6 +51,7 @@
 from nuitka.utils.Timing import TimerReport
 from nuitka.utils.Utils import (
     getOS,
+    getVCRedistPath,
```
kayhayen: Just a pet peeve of mine, I like to call is MSVC and will rename it before merge myself.

nuitka/utils/Utils.py
```diff
@@ -508,6 +509,91 @@ def isCoffUsingPlatform():
...
+
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return None
+
+    redist_base_path = os.path.join(vs_path, "VC", "Redist", "MSVC")
+
+    if not os.path.exists(redist_base_path):
+        return None
+
+    try:
```
kayhayen, line 566: This seems repeated code, and we might actually always want that full thing, so lets move it to the utils function.

nuitka/utils/Utils.py
```diff
@@ -508,6 +509,91 @@ def isCoffUsingPlatform():
...
+
+    return None
+
+
+def getVCRedistPath():
+    global _vcredist_path, _vcredist_path_searched
+
+    if not _vcredist_path_searched:
+        if isWin32Windows():
+            _vcredist_path = findVCRedistPath()
```
kayhayen: Hm, my last comment might be misguided, as this does what I think it ought to.

nuitka/freezer/DllDependenciesWin32.py
kayhayen: I used getSubDirectoriesWithDlls which limits it nicely to only directories with DLLs inside.

nuitka/utils/Utils.py
kayhayen: That was already the function I was looking for

nuitka/freezer/Standalone.py
kayhayen: I renamed a bunch of things accordingly

kayhayen:
Please check after my changes, I might have broken stuff and I am not too aware of what test case to use. With --report you should be able to confirm where DLLs are coming from. Some runtime DLLs to be taken from Python installation like `vcruntime140.dll` and from PyPI are OK in my book, but I think this change overrides them I guess.

Also I will probably add an option to disable this, such that if it causes a regression for some users, we can handle that easily.

kayhayen:
But other than that, this is ready for merge now.

timminator:
I've tested the PR in it's current state with your changes. Everything works correctly on two test cases where these dlls were previously missing. The report also shows that they were correctly included from the msvc redist folder.
Just for your info - the vcruntime140.dll for example was still taken from the python installation as before, so this PR did not change that behaviour.
I cannot see how this could cause a regression, but an option to disable it cannot hurt I guess.
Thanks for approving!

kayhayen:
For people wanting to reduce size, including runtime DLLs has no point, so I would then also use that to block their inclusion, although I think some packages kind of insist on manually loading what they brought along, which in turn can also break stuff. I guess, I will forego the option, as I won't have the capacity to make it work well at this time, and we can still tell people to manually exclude such DLLs if they care, but I somehow get the feeling this will come up later.

kayhayen:
Thanks @timminator would you also be willing to follow https://nuitka.net/pages/website-manual.html and update this page https://nuitka.net/user-documentation/tips.html#redistributing-windows-standalone-programs which will become outdated. I think we should inform people also about this issue now. The UCRT side of things doesn't seem to cover it.

I need to check winlibs, if it does have those DLLs that we were speaking of, but probably not.

Either way, definite thanks for bearing with me. This is something I wouldn't have figured out on my own, esp. the legal side of things seemed inhibiting to me. Of course we both are not giving legal advice, but I feel that the default behaviour might be a lot more legal and usable now. Commonly people run things in empty VMs and then encounter issues, this will be gone now for MSVC users at least.

timminator:
In the next few days I won't have time for this, but I will take a look at the docs when I find some time.
Thanks for merging.

PR #3603, opened 2025-09-14
Fix PEP 695 type alias syntax

ZeroIntensity:
# What does this PR do?

Fixes use of type variables in 3.12+ `type` assignments, e.g., `type A[T] = T | int`.

# PR Checklist

## Summary by Sourcery

Implement full PEP 695 type alias syntax support by extending the AST node model, refining the assignment transformation to handle type parameters in an outline helper, and updating code generation to emit correct alias code.

New Features:
- Add support for PEP 695 style type alias assignments with type parameters (e.g., type A[T] = ...).

Enhancements:
- Extend AST by introducing a mixin that carries the alias name, type parameters, and value and update ExpressionTypeAlias to use it.
- Refactor buildTypeAliasNode to generate an outline function that binds type variables before returning the alias.
- Adjust code generation logic in TypeAliasCodes to derive the alias name from the AST and correctly manage temporary variables.

Chores:
- Remove stale license headers from generated node files.

### DIFF
--- a/nuitka/code_generation/TypeAliasCodes.py
+++ b/nuitka/code_generation/TypeAliasCodes.py
@@ -47,7 +47,7 @@ def generateTypeAliasCode(to_name, expression, emit, context):
 
         getErrorExitCode(
             check_name=value_name,
-            release_names=(type_alias_name, compute_value_name),
+            release_names=(type_alias_name, compute_value_name, type_params_name),
             emit=emit,
             context=context,
             needs_check=expression.mayRaiseExceptionOperation(),
--- a/nuitka/code_generation/TypeAliasCodes.py
+++ b/nuitka/code_generation/TypeAliasCodes.py
@@ -26,10 +26,7 @@ def generateTypeAliasCode(to_name, expression, emit, context):
         expression=expression.subnode_value, emit=emit, context=context
     )
 
-    assert (
-        expression.getParent().isStatementAssignmentVariable()
-    ), expression.getParent()
-    type_alias_name = expression.getParent().getVariableName()
+    type_alias_name = expression.subnode_name.getVariableName()
 
     with withObjectCodeTemporaryAssignment(
         to_name, "type_alias_value", expression, emit, context
--- a/nuitka/nodes/ChildrenHavingMixins.py
+++ b/nuitka/nodes/ChildrenHavingMixins.py
@@ -1,3 +1,6 @@
+#     Copyright 2025, Kay Hayen, mailto:kay.hayen@gmail.com find license text at end of file
+
+
 # We are not avoiding these in generated code at all
 # pylint: disable=I0021,line-too-long,too-many-instance-attributes,too-many-lines
 # pylint: disable=I0021,too-many-arguments,too-many-return-statements,too-many-statements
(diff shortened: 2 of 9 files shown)


ZeroIntensity:
@kayhayen This should be ready. I'm not sure why the `generate-specialized-python-code` script deleted some license comments, but I'll assume that's intentional.

nuitka/nodes/ChildrenHavingMixins.py
```diff
@@ -12586,6 +12583,161 @@ def collectVariableAccesses(self, emit_read, emit_write):
...
+    def finalize(self):
+        del self.parent
+
+        self.subnode_name.finalize()
+        del self.subnode_name
+        for c in self.subnode_type_params:
+            c.finalize()
+        del self.subnode_type_params
+        self.subnode_value.finalize()
+        del self.subnode_value
```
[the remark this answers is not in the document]
ZeroIntensity: This is generated code.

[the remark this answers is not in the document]
kayhayen: Also finalize crashing tell us that a finalized object is used, and a guard would have to do the same. We need to break cyclic references there or else memory doesn't get freed.

kayhayen:
> @kayhayen This should be ready. I'm not sure why the `generate-specialized-python-code` script deleted some license comments, but I'll assume that's intentional.

Yeah, it's a problem with me not having copyright headers in what I work with internally and using the tool mostly there, it of course ought to generate them properly, which is a kind of non-important bug to fix though, but also one that prevents a test for that to work properly. I want copyright stuff to be generated so it's consistent. But of course that could also be achieved by some sort of test.

kayhayen:
You also left behind a bunch of unused imports

nuitka/code_generation/TypeAliasCodes.py
```diff
@@ -26,10 +26,7 @@ def generateTypeAliasCode(to_name, expression, emit, context):
         expression=expression.subnode_value, emit=emit, context=context
     )
 
-    assert (
```
kayhayen, line 29: Very good finding of yours.

nuitka/code_generation/TypeAliasCodes.py
```diff
@@ -47,7 +44,7 @@ def generateTypeAliasCode(to_name, expression, emit, context):
 
         getErrorExitCode(
             check_name=value_name,
-            release_names=(type_alias_name, compute_value_name),
+            release_names=(type_alias_name, compute_value_name, type_params_name),
```
kayhayen, line 47: That looks like another good catch, this must have been memory leaking before.

nuitka/tree/ReformulationAssignmentStatements.py
```diff
@@ -1228,18 +1235,50 @@ def buildTypeVarNode(node, source_ref):
 def buildTypeAliasNode(provider, node, source_ref):
     """Python3.12 or higher, type alias statements."""
 
-    assert not node.type_params, node.type_params
-    type_alias_node = ExpressionTypeAlias(
-        type_params=buildNodeTuple(provider, node.type_params, source_ref),
-        value=buildNode(provider, node.value, source_ref),
-        source_ref=source_ref,
-    )
+    type_alias_name = mangleName(node.name.id, provider)
```
kayhayen, line 1234: Good for realizing the need for mangling

nuitka/tree/ReformulationAssignmentStatements.py
```diff
@@ -1228,18 +1235,50 @@ def buildTypeVarNode(node, source_ref):
...
+
+    def createTypeExpression():
```
kayhayen, line 1236: What I don't like as much is that creates an outline function even if it becomes a simple return. Optimization will remove that, but it creates a bit of churn in verbose output I bet and produce a "change" which will though only make a difference if there is ever not more than one micro pass, but still. I probably want to make that outlike optional and just create the alias directly if there are no generic types.

kayhayen (approved):
I am taking it from here and will apply a few changes before merging it.

kayhayen:
This ought to be ready for merge. I had issues rebasing it on top of my t-string changes, the generated code is really ugly for merges I guess.

kayhayen:
We already know that deferred evaluation is not working the way it is expected, but that's pretty independent of the generic issue and exists for the whole thing. We will deal with that separately. I however wish when we add that for there to be optimization for said functions that will inline them if possible, and possible there means module variables are trusted values or assigned before usage of the type, etc. and that needs to be dealt with separately.

Also we know that star syntax also doesn't work yet for these.

Generally we should try to name PRs for PEPs and rather for the actual thing they do. In this case adding generic type alias support.

Issue #3457, opened 2025-05-05
TypeVarTuple unpacking syntax in function arguments causes an assertion error during compilation


Wizzerinus:
Attempting to compile files using one specific syntax from PEP 646 causes a Nuitka error. Other syntax features (i.e. `Generic[*Ts]`) work fine.

Nuitka version (installed from pip):

```
2.7
Commercial: None
Python: 3.13.3 (main, Apr  9 2025, 07:44:25) [GCC 14.2.1 20250207]
Flavor: Arch Python
GIL: yes
Executable: ~/.local/share/virtualenvs/nuitka/bin/python
OS: Linux
Arch: x86_64
Distribution: Arch None
Version C compiler: /usr/bin/gcc (gcc 15.1.1).
```

The file that crashes:

```python
from typing import TypeVarTuple

Ts = TypeVarTuple("Ts")
def a(*args: *Ts): pass
```

The traceback:
```
  File "/home/wizzerinus/.local/share/virtualenvs/nuitka/lib/python3.13/site-packages/nuitka/tree/ReformulationFunctionStatements.py", line 343, in buildFunctionNode
    annotations = buildParameterAnnotations(provider, node, source_ref)
  File "/home/wizzerinus/.local/share/virtualenvs/nuitka/lib/python3.13/site-packages/nuitka/tree/ReformulationFunctionStatements.py", line 635, in buildParameterAnnotations
    extractArgAnnotation(node.args.vararg)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/home/wizzerinus/.local/share/virtualenvs/nuitka/lib/python3.13/site-packages/nuitka/tree/ReformulationFunctionStatements.py", line 603, in extractArgAnnotation
    value=buildAnnotationNode(provider, arg.annotation, source_ref),
          ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wizzerinus/.local/share/virtualenvs/nuitka/lib/python3.13/site-packages/nuitka/tree/TreeHelpers.py", line 427, in buildAnnotationNode
    return buildNode(provider, node, source_ref)
  File "/home/wizzerinus/.local/share/virtualenvs/nuitka/lib/python3.13/site-packages/nuitka/tree/TreeHelpers.py", line 319, in buildNode
    assert False, ast.dump(node)
           ^^^^^
[...]

Issue #2930, opened 2024-06-17
PEP 695 support


trebinor:
Python 3.12 added support for [PEP 695: Type Parameter Syntax](https://docs.python.org/3/whatsnew/3.12.html#pep-695-type-parameter-syntax).  How much of this is intended to be supported?

Here is an example of the upper bound specification working in CPython, but not when built with Nuitka.

```
# main.py
def foo[T: str](b: T) -> None:
    print(b)

if __name__ == "__main__":
    foo("bar")
```
```
# pyproject.toml
[project]
name = "pep695"
version = "0.1.0"
requires-python = ">= 3.12"
dependencies = [ "nuitka" ]
```
```
# Running CPython version in .venv (works)
$ python3 main.py
bar
```
```
# Running Nuitka version (error)
$ python3 -m nuitka main.py && ./main.bin
Nuitka-Options: Used command line options: main.py
Nuitka-Options:WARNING: You did not specify to follow or include anything but main program. Check options and make sure that is intended.
Nuitka: Starting Python compilation with Nuitka '2.3.4' on Python '3.12' commercial grade 'not installed'.
Nuitka: Completed Python level compilation and optimization.
Nuitka: Generating source code for C backend compiler.
Nuitka: Running data composer tool for optimal constant value handling.
Nuitka: Running C compilation via Scons.
Nuitka-Scons: Backend C compiler: gcc (gcc 14.1.1).
Nuitka-Scons: Backend linking program with 6 files (no progress information available for this stage).
Nuitka-Scons: Compiled 6 C files using ccache.
Nuitka-Scons: Cached C files (using ccache) with result 'cache hit': 6
Nuitka: Keeping build directory 'main.build'.
Nuitka: Successfully created 'main.bin'.
Traceback (most recent call last):
  File "/home/trebinor/ws/raw/pep695/main.py", line 1, in <module>
    def foo[T: str](b: T) -> None:

NameError: name 'T' is not defined
```
Here is environment information required by the issue submission template.
```
$ python -m nuitka --version
2.3.4
Commercial: None
Python: 3.12.3 (main, Apr 23 2024, 09:16:07) [GCC 13.2.1 20240417]
Flavor: Arch Python
Executable: /home/trebinor/ws/raw/pep695/.venv/bin/python
OS: Linux
Arch: x86_64
Distribution: Arch None
Version C compiler: /usr/bin/gcc (gcc 14.1.1).
```

kayhayen:
[...]

PR #3648, opened 2025-10-20
Implement function type syntax

ZeroIntensity:
# What does this PR do?

Implement function type syntax, so this now works correctly:

```py
from typing import Callable

def foo[T, **P](function: Callable[P, T]) -> Callable[P, T]:
    print(type(T), type(P))
    return function

def a():
    pass

foo(a)
```

- #2930
- #3457

# PR Checklist

## Summary by Sourcery

Implement function type syntax with support for TypeVarTuple and ParamSpec

New Features:
- Add C runtime structs and allocator functions for TypeVarTuple and ParamSpec
- Introduce MAKE_TYPE_VAR_TUPLE and MAKE_PARAM_SPEC helpers and integrate them into type alias code generation
- Extend ExpressionTypeVariable to include a variant field and dispatch between TypeVar, TypeVarTuple, and ParamSpec
- Wrap functions with type parameter annotations via an AST transformation when generating Python bytecode

### DIFF
--- a/nuitka/tree/ReformulationFunctionStatements.py
+++ b/nuitka/tree/ReformulationFunctionStatements.py
@@ -247,7 +247,7 @@ def wrapWithTypeAnnotations(provider, type_params, body, source_ref):
 
     outline_body.setChildBody(
         makeStatementsSequenceFromStatements(
-            *assignments, StatementReturn(expression=body, source_ref=source_ref)
+            assignments, StatementReturn(expression=body, source_ref=source_ref)
         )
     )
 


nuitka/tree/ReformulationAssignmentStatements.py
```diff
@@ -1225,7 +1225,15 @@ def buildNamedExprNode(provider, node, source_ref):
...
-    return ExpressionTypeVariable(node.name, source_ref=source_ref)
+    node_type = type(node).__name__
+    if node_type == "TypeVar":
+        variant = 0
+    elif node_type == "TypeVarTuple":
+        variant = 1
+    else:
+        assert node_type == "ParamSpec", node_type
+        variant = 2
+    return ExpressionTypeVariable(node.name, variant, source_ref=source_ref)
```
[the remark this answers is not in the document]
ZeroIntensity: We can't use `isinstance` because the AST classes won't exist on some versions.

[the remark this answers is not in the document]
kayhayen: I have done it like this in such cases for example:

```
        elif python_version >= 0x370 and node_class is ast.comprehension:
```

I don't think `isinstance` makes any sense. I don't even want to know their inhertitence. Going via `is` is good enough for me there and definitely only faster.

nuitka/tree/ReformulationAssignmentStatements.py
```diff
@@ -1225,7 +1225,15 @@ def buildNamedExprNode(provider, node, source_ref):
 
 
 def buildTypeVarNode(node, source_ref):
-    return ExpressionTypeVariable(node.name, source_ref=source_ref)
+    node_type = type(node).__name__
+    if node_type == "TypeVar":
```
kayhayen: If anything, it feels like variant should be a string and not a number that conveys no information, and I don't see why it wouldn't be different node types at this point.

nuitka/tree/ReformulationAssignmentStatements.py
```diff
@@ -1225,7 +1225,15 @@ def buildNamedExprNode(provider, node, source_ref):
 
 
 def buildTypeVarNode(node, source_ref):
-    return ExpressionTypeVariable(node.name, source_ref=source_ref)
+    node_type = type(node).__name__
```
kayhayen: I would prefer using ast.TypeVar and the like, guarded with version checks for Python if necessary, but working of these strings I feel is ugly and demotes expressiveness of code.

nuitka/tree/ReformulationFunctionStatements.py
```diff
@@ -224,6 +226,33 @@ def _buildBytecodeOrSourceFunction(provider, node, compilation_mode, source_ref)
...
+        )
+        assignments.append(assign)
+
+    outline_body.setChildBody(
+        makeStatementsSequenceFromStatements(
+            *assignments, StatementReturn(expression=body, source_ref=source_ref)
+        )
+    )
+
+    return outline_body
```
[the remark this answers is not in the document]
kayhayen: I don't even get what it's saying here, but if it prefers list contractions, I do not. They just make life harder when attempting to debug things.

nuitka/nodes/TypeNodes.py
```diff
@@ -355,7 +355,7 @@ class ExpressionTypeVariable(ExpressionTypeVariableBase, ExpressionBase):
     kind = "EXPRESSION_TYPE_VARIABLE"
 
     auto_compute_handling = "final,no_raise"
-    node_attributes = ("name",)
+    node_attributes = ("name", "variant")
```
kayhayen: I don't know how interesting it is ever going to be for optimization, but this variant ought to be derived classes of a common base class, no?

nuitka/tree/ReformulationAssignmentStatements.py
ZeroIntensity: A string makes sense to me. I could switch it to dedicated node types, but it would just be more complicated. Which would you prefer?

nuitka/tree/ReformulationAssignmentStatements.py
ZeroIntensity: Ok, I'll do it that way.

nuitka/tree/ReformulationAssignmentStatements.py
kayhayen: I don't know how different they are, they look like very different things to me, being different types and all being constructed very differently. Looking at the code generation for the "variants", do  they have anything in common really, except being in a similar spot in a syntax tree? Therefore I thought they better be 3 different ndoes. I am very open to being proven wrong, if these are really all the same, but I just don't see it now.

ZeroIntensity: They're not similar in terms of code generation, they're just functionally similar at runtime. I'll give each a dedicated node, I guess.

ZeroIntensity: Done.

nuitka/tree/ReformulationAssignmentStatements.py
ZeroIntensity: Done.

nuitka/nodes/TypeNodes.py
ZeroIntensity: Done.

ZeroIntensity:
How can I shut up the 3.9 linter?

kayhayen:
Search for the mnemonic, I allow using pylint on older versions as well, wrong OS, etc. so this comes a lot. Typically along with a disable of false positives.

These are two examples:

```
    if str is bytes:
        import _winreg as winreg  # pylint: disable=I0021,import-error,no-name-in-module
    else:
        import winreg  # pylint: disable=I0021,import-error,no-name-in-module
```

kayhayen:
When that's used in more places, I have at times moved things to `__past__` and provided a fallback, so I can write code that just uses the name, but it might be `None`.

```
try:
    ExceptionGroup = ExceptionGroup  # pylint: disable=I0021,redefined-builtin
except NameError:
    ExceptionGroup = None

try:
    BaseExceptionGroup = BaseExceptionGroup  # pylint: disable=I0021,redefined-builtin
except NameError:
    BaseExceptionGroup = None
````

In Nuitka we never want backports and the like to be effective, or else we loose the ability to compile time compute the program environment from our own.

I kind of like to have pylint lingering a bit in the past, as it's really good at pointing out that something doesn't exist yet. One day I am going to make the linting be a separate stage more clearly. Not sure how I feel it preventing runs if it fails, that feels annoying, but right now it's sort of in the middle. Maybe that's something that AI could help me come up with. But if linting fails, maybe we don't need all Python versions to be tested, and just one, as long as we guantuee an overall build failure. Saving cycles for Actions is not unimportant actually.

File: nuitka/code_generation/TypeAliasCodes.py [hunk 6]
```diff
@@ -57,10 +57,22 @@ def generateTypeVarCode(to_name, expression, emit, context):
     with withObjectCodeTemporaryAssignment(
         to_name, "type_var_value", expression, emit, context
     ) as value_name:
+        if expression.kind == "EXPRESSION_TYPE_VARIABLE":
```
kayhayen: Expressions has test functions like `isExpressionTypeVariable()` automatically.

kayhayen: Checkout `generateCAPIObjectCode` for a function that given the helper name, expects it to be a callable that returns a reference (there is also generateCAPIObjectCode0 that does not) and that does the call code and error checking much more elegant, if you ask the expression about it. It seems only the API name changes.

Instead of "function" which can be so many things, I would have named it "helper_function". Avoid generic names.

kayhayen (changes requested):
Good on the nodes side. One thing I was contemplating why we don't use `ExpressionTypeShapeExactMixin`  but it seems it's unused and for good reason, as that would mean it's exactly `type` and not a type derived one.

Knowing that shape could at times avoid certain checks, when used for other things, like say use in base classes, that kind of thing, but I think we cannot go there as part of this at all, esp. as the impact would be limited, but correctness could easily suffer and we aim for this to be hotfix material.

For the C side of code generation, it was maybe missing error handling. If that's because it cannot raise, the nodes need to say so, and the function suggested will then add an assertion instead of an error exit goto with cleanup.

nuitka/code_generation/TypeAliasCodes.py
ZeroIntensity: Sorry for the delay on my part. How do I use `generateCAPIObjectCode` without a subnode to pass as the name?

nuitka/tree/ReformulationAssignmentStatements.py
```diff
@@ -1225,7 +1230,16 @@ def buildNamedExprNode(provider, node, source_ref):
 
 
 def buildTypeVarNode(node, source_ref):
-    return ExpressionTypeVariable(node.name, source_ref=source_ref)
+    assert python_version >= 0x3C0
```
kayhayen: The is entirely unnecessary to do your own dispatching as buildnode already goes by the name and `setBuildingDispatchers` should do it and then instead of calling "buildTypeVarNode" you ought to call "buildNode" instead. Just register these names as well and have dedicated build functions for them

PR #3662, opened 2025-11-06
Python 3.14 support

ZeroIntensity:
# What does this PR do?

Adds support for Python 3.14 by running the test suite until it all passes.

# PR Checklist

## Summary by Sourcery

Add compatibility for Python 3.14 by updating generated C runtime integration, annotation handling in compiled functions, build scripts, and tests.

Enhancements:
- Expand compiled generator integration and opcode deoptimization mapping to support Python 3.14's new interpreter frame API
- Introduce deferred annotation evaluation via a new __annotate__ mechanism and update compiled function annotation getters/setters
- Refine dict unhashable key error formatting and reverse context manager __exit__/__enter__ lookup order for Python 3.14

Build:
- Adjust SCons build script to skip hacl library linkage on Python 3.14

Tests:
- Update async test to use asyncio.new_event_loop() for compatibility with Python 3.14

### DIFF
--- a/nuitka/tree/ReformulationFunctionStatements.py
+++ b/nuitka/tree/ReformulationFunctionStatements.py
@@ -68,7 +68,6 @@ from nuitka.PythonVersions import python_version
 from nuitka.specs.ParameterSpecs import ParameterSpec
 
 from .FutureSpecState import getFutureSpec
-from .InternalModule import makeInternalHelperFunctionBody
 from .ReformulationExecStatements import wrapEvalGlobalsAndLocals
 from .ReformulationTryFinallyStatements import (
     makeTryFinallyReleaseStatement,


nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -458,8 +458,17 @@ static PyObject *Nuitka_Function_get_annotations(PyObject *self, void *data) {
...
     if (function->m_annotations == NULL) {
         NUITKA_MAY_BE_UNUSED PyThreadState *tstate = PyThreadState_GET();
-
+#if PYTHON_VERSION < 0x3e0
         function->m_annotations = MAKE_DICT_EMPTY(tstate);
+#else
+        if (function->m_annotate == NULL) {
+            return MAKE_DICT_EMPTY(tstate);
+        }
+        function->m_annotations = CALL_FUNCTION_WITH_SINGLE_ARG(tstate, function->m_annotate, _PyLong_GetOne());
```
kayhayen: We got `const_int_pos_1` as a global value for that.

nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -458,8 +458,17 @@ static PyObject *Nuitka_Function_get_annotations(PyObject *self, void *data) {
...
+        if (function->m_annotations == NULL) {
+            return NULL;
```
kayhayen, line 469: Will that function be executed each time when it throws an error over and over? Is that compatible?

nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -492,6 +501,46 @@ static int Nuitka_Function_set_annotations(PyObject *self, PyObject *value, void
...
+        Py_INCREF_IMMORTAL(Py_None);
+        return Py_None;
+    }
+    CHECK_OBJECT(function->m_annotate);
+
+    Py_INCREF(function->m_annotate);
+    return function->m_annotate;
+}
+
+static int Nuitka_Function_set_annotate(PyObject *self, PyObject *value, void *data) {
```
kayhayen, line 521: Can one really set that attribute in Python?

nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -698,14 +750,7 @@ static PyObject *Nuitka_Function_clone(struct Nuitka_FunctionObject *function, P
...
 
-    PyObject *annotations = function->m_annotations;
-    if (annotations != NULL) {
-        if (DICT_SIZE(annotations) != 0) {
-            annotations = DICT_COPY(tstate, annotations);
-        } else {
-            annotations = NULL;
-        }
-    }
+    PyObject *annotations = function->m_annotate;
```
kayhayen, line 754: That change not having versions guard seems wrong?

nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -1429,9 +1478,17 @@ struct Nuitka_FunctionObject *Nuitka_Function_New(function_impl_code c_code, PyO
 #if PYTHON_VERSION >= 0x300
     assert(kw_defaults == NULL || (PyDict_Check(kw_defaults) && DICT_SIZE(kw_defaults) > 0));
     result->m_kwdefaults = kw_defaults;
+#endif
 
+#if (PYTHON_VERSION >= 0x300) && (PYTHON_VERSION < 0x3e0)
     assert(annotations == NULL || (PyDict_Check(annotations) && DICT_SIZE(annotations) > 0));
     result->m_annotations = annotations;
+#elif PYTHON_VERSION >= 0x3e0
```
kayhayen: Putting  that branch first would be a lot easier.

nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -1429,9 +1478,17 @@ struct Nuitka_FunctionObject *Nuitka_Function_New(function_impl_code c_code, PyO
...
+    // For simplicity's sake, the annotations parameter doubles as the __annotate__
+    // parameter on 3.14+
+    assert(annotations == NULL || PyCallable_Check(annotations));
+    result->m_annotations = NULL;
```
kayhayen, line 1498: However, `m_annotations` seems now unused with 3.14?

nuitka/build/static_src/CompiledGeneratorTypeUncompiledIntegration.c
```diff
@@ -501,7 +501,264 @@ static int Nuitka_PyGen_gen_close_iter(PyThreadState *tstate, PyObject *yf);
 
 // Private opcode mapping, that we need too
 const uint8_t Nuitka_PyOpcode_Deopt[256] = {
-#if PYTHON_VERSION >= 0x3d0
+#if PYTHON_VERSION >= 0x3e0
+    [121] = 121,
```
kayhayen, line 505: Curious, why are there plain numbers used in CPython now?

nuitka/tree/ReformulationFunctionStatements.py
```diff
@@ -643,9 +735,15 @@ def extractArgAnnotation(arg):
         )
 
     if keys:
-        return makeDictCreationOrConstant2(
-            keys=keys, values=values, source_ref=source_ref
-        )
+        # On 3.14+, annotations are deferred by default.
+        if python_version >= 0x3E0:
+            return makeDeferredAnnotateFunction(
```
kayhayen: Are you saying, that all this is doing, is to create a function that returns the dictionary, just later? I thought the actual code of the annotations is to be executed delayed.

```
def f():
   x : something()

f()
f.__annotations__
```

I was assuming this will not crash when calling f().

kayhayen: Actually, that's ignored in all versions, but this is showing it:

```
class C():
   x : something()

print(C.__annotations__)
```

With 3.14, this crashes when `.__annotations__` is looked up, and with 3.13, it does not.

kayhayen: With a function, it shows like this:

```
def f(a : something()):
   pass

print(f.__annotations__)
```

The re-formulation needs to use the nodes and create the updates to the `__annotations__` as a function result. You need to check all uses of `__annotations__` in the building code.

kayhayen: On second thought, that might be already correct for functions if "keys" and "values" are the ones from the function definition, not sure now, but it definitely doesn't do classes yet. And for empty annotations, I think it ought not to create a function, but instead end up passing a value. Or we check if that functions becomes a constant returner, in which case we ought to just pass its value or something, but that is optimization, which should for now maybe only result in TODOs. But a new function object per function, we don't want to hurt performance that much if we can help it in any way.

nuitka/build/static_src/CompiledFunctionType.c
kayhayen: I just checked, you can indeed, which makes it kind of a requirement to have a function object indeed.

kayhayen:
The current build is also failing for pre-3.14, I wouldn't add 3.14 to the matrix yet, but in the end when it all works, otherwise it's hiding those kinds of errors.

nuitka/build/static_src/CompiledFunctionType.c
ZeroIntensity: No, it's used as a cache. Once `__annotate__` is called, the result is stored in `__annotations__`, so the annotations don't have to be re-evaluated every time.

nuitka/build/static_src/CompiledGeneratorTypeUncompiledIntegration.c
ZeroIntensity: Unknown opcodes need to be marked as de-opting to themselves to support some third-party JIT compilers. See https://github.com/python/cpython/issues/128045. It's probably fine to omit it for Nuitka, but I don't know what will happen if we do.

nuitka/build/static_src/CompiledFunctionType.c
ZeroIntensity: You're right, fixed.

nuitka/build/static_src/CompiledFunctionType.c
ZeroIntensity: It seems so, you can try this script:

```py
def test(x):
    pass

def evil(format):
    raise RuntimeError("evil")

test.__annotate__ = evil

for _ in range(3):
    try:
        print(test.__annotations__)
    except RuntimeError as error:
        print(error)
```

nuitka/build/include/nuitka/prelude.h
```diff
@@ -392,8 +392,15 @@ typedef long Py_hash_t;
  * The Python 3.7.0 release on at Linux doesn't work this way either, was
  * a bad CPython release apparently and between 3.7.3 and 3.7.4 these have
  * become runtime incompatible.
+ *
+ * On Python 3.14.0, the introduction of _Py_TriggerGC() also broke this.
```
kayhayen: Please elaborae some more, not having these is very problematic for performance I believe

tests/basics/ReferencingTest35.py
```diff
@@ -180,7 +180,7 @@ async def run():
 
 
 def simpleFunction10():
-    asyncio.get_event_loop().run_until_complete(run())
+    asyncio.new_event_loop().run_until_complete(run())
```
kayhayen: What's being tried there is to do minimal work under test, creating a new event loop (and releasing it?) is by definition not that. If we were to create our own on the outside and use that, no problem though.

nuitka/nodes/DictionaryNodes.py
```diff
@@ -232,21 +233,42 @@ def computeExpression(self, trace_collection):
...
-                    ),
-                    source_ref=key.source_ref,
-                )
-                result = wrapExpressionWithSideEffects(
-                    side_effects=side_effects,
-                    old_node=key,
-                    new_node=result,
-                )
+                    )
+                    # For some reason, calling wrapExpressionWithSideEffects
```
kayhayen: Not keeping them (the side effects) is not acceptable though, I would need to see what you actually tried there to tell, but side effects need to be of course that of the dictionaries key/values, not the dictionary building itself.

nuitka/tree/ReformulationFunctionStatements.py
```diff
@@ -607,6 +618,87 @@ def buildParameterKwDefaults(provider, node, function_body, source_ref):
...
+            source_ref=source_ref,
+        ),
+        source_ref=source_ref,
+    )
+
+    outer_body.setChildBody(body)
+    return outer_body
+
+
+def makeDeferredAnnotateFunction(provider, keys, values, source_ref):
```
kayhayen: I believe this may actually be correct, and execute indeed delayed, and as such then be correct. 

Does this pass any kinds of tests for it?

tests/basics/ReferencingTest35.py
ZeroIntensity: Unfortunately, this raises an exception now, since `get_event_loop` no longer creates an event loop if one isn't running. It might work once I fix the exceptions in `makeDiffable`.

nuitka/build/include/nuitka/prelude.h
ZeroIntensity: In 3.14, `_PyObject_GC_TRACK` now has a call to `_Py_TriggerGC`, which is not exported.

nuitka/tree/ReformulationFunctionStatements.py
ZeroIntensity: Yeah, this passes the tests with annotations in `tests/basics`.

tests/basics/ReferencingTest35.py
kayhayen: Yeah, but we actually mean to succeed in using it.

kayhayen:
So, my goal is to get the merge-ready bits sorted out after I make a new pre-release out of factory later today.

kayhayen:
Seems the event loop taking is already leaking on 3.13, which is maybe kind of expected, I will try and move that to the outside the function, so it happens only once, max. I think we want to cover "run()" as a coroutine and not that.

kayhayen:
Since I previously cherry-picked content and not commits, and even then squashed, I will have to squash this branch, to rebase it and do another round of cherry-picks. I recommend you set a tag in case you want to refer to this state in case I made mistakes.

kayhayen:
So, this had become empty and can be closed. Rebasing gave me issues, but I merged it all, with the caveat of deferred annotation functions not being used yet.

Issue #3634, opened 2025-10-10
Feature: Add --macos-console-mode=force to auto-open Terminal for console apps launched from DMG/GUI on macOS


KRRT7:
# Feature: Add `--macos-console-mode=force` to auto-open Terminal for console apps launched from DMG/GUI on macOS

## 🎯 Problem / Use Case

When building Textual-based or other console applications with Nuitka on macOS using `--macos-create-app-bundle`, applications are packaged into `.app` bundles that can be distributed via DMG files. However, when users double-click these applications from Finder (or launch them from a DMG), they execute in a GUI context without a TTY/terminal connection.

For console applications (especially TUI apps built with libraries like Textual), this results in:
- Users see nothing happen when they launch the app
- The application may crash or behave unexpectedly without proper terminal I/O
- Poor user experience requiring users to manually open Terminal and navigate to the app

Currently, this forces developers to create custom launcher solutions or educate users on manual terminal launching.

## 💡 Proposed Solution

Add a new macOS-specific compiler option `--macos-console-mode=force` (name is bikesheddable) that:

1. **Runtime Detection**: When the compiled application starts, detects if it's running in a GUI context (launched from Finder/DMG) without an associated terminal
2. **Automatic Terminal Spawning**: If the condition is met, automatically re-launches itself inside the user's default Terminal application
3. **Fork-bomb Protection**: Includes safeguards to prevent infinite re-launching loops

This would provide a native Nuitka solution similar to what PyInstaller users currently implement with custom launchers.

### Integration Point
Hook into `nuitka/build/static_src/MainProgram.c` near the existing `_NUITKA_MACOS_BUNDLE_MODE` logic (around lines 43-45).

### Implementation Approach
- **Detection**: Check if `isatty(STDIN_FILENO)` returns false and no controlling terminal exists
- **Re-launch Method**: Use macOS's `open -a Terminal` command or AppleScript approach to spawn the application inside Terminal
- **Protection**: Set environment variable `NUITKA_TERMINAL_RELAUNCH=1` before re-exec to prevent recursive launches

### Option Integration
Add to the existing macOS options group in `nuitka/OptionParsing.py` alongside other `--macos-*` flags:

```python
macos_group.add_option(
    "--macos-console-mode",
    action="store",
    dest="macos_console_mode",
    choices=("force", "detect", "disable"),
    metavar="CONSOLE_MODE",
    default="detect",
    help="""
    Select console mode for macOS app bundles. 'force' always opens Terminal
[...]

PR #3675, opened 2025-11-26
implement #3634 --macos-console-mode=force

KRRT7:

# What does this PR do?

# PR Checklist

## Summary by Sourcery

Add configurable macOS console launching behavior for app bundles and propagate the setting through option parsing and the build system.

New Features:
- Introduce a --macos-console-mode option to control whether macOS app bundles relaunch in a Terminal window.
- Implement macOS runtime logic to relaunch the executable in Terminal based on the selected console mode when not already in a terminal.

Enhancements:
- Pass the selected macOS console mode from option parsing into Scons and define corresponding compile-time flags for bundle builds.

### DIFF
--- a/nuitka/tree/ReformulationClasses3.py
+++ b/nuitka/tree/ReformulationClasses3.py
@@ -201,6 +201,30 @@ def buildClassNode3(provider, node, source_ref):
 
     class_locals_scope.registerProvidedVariable(class_variable)
 
+    if python_version >= 0x3C0:
+        type_param_nodes = node.type_params
+    else:
+        type_param_nodes = None
+
+    if type_param_nodes is not None:
+        type_params_expressions = buildNodeTuple(
+            provider=outline_body, nodes=type_param_nodes, source_ref=source_ref
+        )
+    else:
+        type_params_expressions = ()
+
+    type_variables = []
+
+    for type_params_expression in type_params_expressions:
+        type_variable = class_locals_scope.getLocalVariable(
+            owner=class_creation_function,
+            variable_name=type_params_expression.name,
+        )
+
+        class_locals_scope.registerProvidedVariable(type_variable)
+
+        type_variables.append(type_variable)
+
     class_variable_ref = ExpressionVariableRef(
         variable=class_variable, source_ref=source_ref
     )
@@ -257,6 +281,17 @@ def buildClassNode3(provider, node, source_ref):
         ),
     ]
 
+    for type_variable, type_params_expression in zip(
+        type_variables, type_params_expressions
+    ):
+        statements.append(
+            makeStatementAssignmentVariable(
+                variable=type_variable,
+                source=type_params_expression.makeClone(),
+                source_ref=source_ref,
+            )
+        )
+
     if class_doc is not None:
         statements.append(
             StatementAssignmentVariableName(
@@ -285,27 +320,6 @@ def buildClassNode3(provider, node, source_ref):
 
     if python_version >= 0x300:
         qualname_assign = statements[-1]
-
-    if python_version >= 0x3C0:
-        type_param_nodes = node.type_params
-    else:
-        type_param_nodes = None
-
-    if type_param_nodes is not None:
-        type_params_expressions = buildNodeTuple(
-            provider=outline_body, nodes=type_param_nodes, source_ref=source_ref
-        )
-    else:
-        type_params_expressions = ()
-
-    for type_params_expression in type_params_expressions:
-        assign = StatementAssignmentVariableName(
-            provider=class_creation_function,
-            variable_name=type_params_expression.name,
-            source=type_params_expression.makeClone(),
-            source_ref=source_ref,
-        )
-        statements.append(assign)
 
     if python_version >= 0x3D0:
         statements.append(
--- a/nuitka/tree/ReformulationFunctionStatements.py
+++ b/nuitka/tree/ReformulationFunctionStatements.py
@@ -237,7 +237,7 @@ def _wrapWithTypeAnnotations(provider, type_params, body, source_ref):
         type_var = buildNode(provider=provider, node=type_param, source_ref=source_ref)
 
         assign = StatementAssignmentVariableName(
-            provider=outline_body,
+            provider=provider,
             variable_name=type_param.name,
             source=type_var,
             source_ref=source_ref,


nuitka/Options.py
```diff
@@ -2580,6 +2580,11 @@ def getMacOSAppVersion():
     return options.macos_app_version
 
 
+def getMacOSConsoleMode():
+    """*str* console mode for macOS app bundles ("force", "detect", "disable")"""
```
kayhayen: What's normally does is to give the option name.

nuitka/Options.py
```diff
@@ -2580,6 +2580,11 @@ def getMacOSAppVersion():
...
+    return options.macos_console_mode or "detect"
```
kayhayen: The value cannot be False, default is "detect" in the option definition.

nuitka/build/SconsInterface.py
```diff
@@ -673,6 +673,10 @@ def getCommonSconsOptions():
     if shallCreateAppBundle():
         scons_options["macos_bundle_mode"] = asBoolStr(True)
 
+        # Pass macOS console mode setting
+        from nuitka.Options import getMacOSConsoleMode
```
kayhayen: That import is misplaced and should not be local

nuitka/build/SconsUtils.py
```diff
@@ -389,6 +389,14 @@ def createEnvironment(
         env.Append(CPPDEFINES=["_NUITKA_MACOS_BUNDLE_MODE"])
         env.Append(LINKFLAGS=["-framework", "CoreFoundation"])
 
+        # MacOS console mode: Control terminal launching behavior
+        macos_console_mode = getArgumentDefaulted("macos_console_mode", "detect")
```
kayhayen: The value is passed only for application bundles, the option never makes it it clear it's for app bundles only. It's name should be `--macos-app-console-mode` and the default here, probably should not be `detect` but something else, as that default will only be ehere for non-macOS and non-app ?

nuitka/build/static_src/MainProgram.c
```diff
@@ -1427,6 +1428,94 @@ static int Nuitka_Main(int argc, native_command_line_argument_t **argv) {
...
+                snprintf(applescript_cmd, sizeof(applescript_cmd),
+                    "osascript -e 'tell application \"Terminal\"' "
+                    "-e 'activate' "
+                    "-e 'do script \"%s\"' "
+                    "-e 'end tell'",
+                    full_command);
+
+                // Execute the command - system() is used here as we're just launching Terminal
+                // and then we'll exit. The actual program will run via exec in the Terminal.
+                int result = system(applescript_cmd);
```
kayhayen: Why is this not using `exec`, keeping the old process around seems wasteful and system takes a shell command, which suffers from all kinds of issues that Sourcery also mentioned. Very bad.

kayhayen (changes requested):
A good start, but needs to become much more robust.

nuitka/OptionParsing.py
```diff
@@ -1818,6 +1818,23 @@ def _handleHelpModes():
...
 
+macos_group.add_option(
+    "--macos-app-console-mode",
+    action="store",
+    dest="macos_app_console_mode",
+    choices=("force", "detect", "disable"),
+    metavar="MODE",
+    default="detect",
+    help="""\
+Select console mode for macOS app bundles. 'force' always opens Terminal
```
kayhayen, line 1961: The punctions are weird, quoting things like Finder or Terminal would be a must.

kayhayen:
This has stalled, correct?

nuitka/Options.py
KRRT7: Fixed - docstring now references the option name: ``--macos-app-console-mode``

nuitka/Options.py
KRRT7: Fixed - removed the `or "detect"` fallback since the default is already set in the option definition.

nuitka/build/SconsInterface.py
KRRT7: Fixed - import moved to module level.

nuitka/build/SconsUtils.py
KRRT7: Fixed - option renamed to `--macos-app-console-mode` and default in SconsUtils is now `"disable"` for non-app-bundle cases.

nuitka/build/static_src/MainProgram.c
KRRT7: Fixed - now using `execve` to directly replace the process with osascript instead of spawn+wait+exit. Also removed the now-unused `<spawn.h>` and `<sys/wait.h>` includes.

kayhayen:
This has hopefully been made obsolete by a simpler solution using a wrapper script with the following advantages

- No osascript used, which would require permissions to be allowed that.
- Argument passing would work, although I have no idea where they would come from.
- No change of process, the launched process does exec to the Terminal which launches the script again which then execs into the main binary. Clean process hierachy. I hope for permissions that means they are retained. When I was still doing onefile in apps, there were issues, but that was probably due to not signed binaries, they now are.

The approach I did may still have issues with the script being not part of the signing, so maybe they are. Anyway, this feels obsolete now.

nuitka/freezer/MacOSApp.py
```diff
@@ -147,6 +160,45 @@ def createPlistInfoFile(logger):
...
+        target_arch: Target architecture (arm64, x86_64, or None for native).
+
+    Returns:
+        True if successful, False otherwise.
+    """
+    stub_source = _getTerminalLauncherStubSourcePath()
+    stub_output = os.path.join(macos_dir, binary_name)
+
+    # Build clang command
+    clang_cmd = ["clang", "-O2", "-o", stub_output, stub_source]
```
kayhayen, line 182: Seriously invoking out own C compilations is not where I want to go.

nuitka/freezer/MacOSApp.py
```diff
@@ -147,6 +160,45 @@ def createPlistInfoFile(logger):
...
+
+    # Add architecture flag if specified
+    if target_arch is not None:
```
kayhayen, line 185: Precisely because of these things.

nuitka/freezer/MacOSApp.py
```diff
@@ -147,6 +160,45 @@ def createPlistInfoFile(logger):
...
+        clang_cmd.extend(["-arch", target_arch])
+
+    # Add macOS deployment target for compatibility
+    clang_cmd.extend(["-mmacosx-version-min=10.13"])
```
kayhayen, line 189: Or these, all just bad duplications. This would have to be done in the Scons backend file if at all, but I don't see the need to use C yet.

nuitka/freezer/MacOSApp.py
```diff
@@ -147,6 +160,45 @@ def createPlistInfoFile(logger):
...
+
+    try:
+        subprocess.check_call(
+            clang_cmd,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
+        )
+        return True
+    except subprocess.CalledProcessError:
```
kayhayen, line 198: AI Error handling at its best.

nuitka/MainControl.py
```diff
@@ -1200,6 +1204,14 @@ def _main():
                     copy_standalone_entry_points=copy_standalone_entry_points,
                 )
 
+                # Create terminal launcher script for macOS app bundles (must be
+                # after DLL processing and signing to avoid install_name_tool errors)
+                if shallCreateAppBundle():
+                    binary_filename = getNormalizedPathJoin(
```
kayhayen, line 1210: Not using functions intended for it also bad

KRRT7:
closing this for now, I think I've identified a better solution

PR #3681, opened 2025-11-28
implement #3634 --linux-console-mode=force

Abdur-rahmaanJ:

# What does this PR do?

# PR Checklist

## Summary by Sourcery

Add configurable Linux app bundle console mode and relaunch compiled binaries in a terminal when appropriate.

New Features:
- Introduce the --linux-app-console-mode option to control terminal behavior for Linux app bundles with modes force, detect, and disable.
- Add a Linux-only startup path that relaunches the binary inside a terminal emulator when no TTY is attached, based on the selected console mode.

Enhancements:
- Plumb the linux_app_console_mode option through Scons to define compile-time flags used by the runtime startup logic.

### DIFF
--- a/nuitka/build/static_src/MainProgram.c
+++ b/nuitka/build/static_src/MainProgram.c
@@ -1393,7 +1393,7 @@ void setLANGSystemLocaleMacOS(void) {
 
 static int Nuitka_Main(int argc, native_command_line_argument_t **argv) {
 
-#if(defined(_NUITKA_LINUX_CONSOLE_MODE_FORCE) || defined(_NUITKA_LINUX_CONSOLE_MODE_DETECT)) && !defined(_WIN32) && !defined(__APPLE__)
+#if(defined(_NUITKA_LINUX_CONSOLE_MODE_FORCE) || defined(_NUITKA_LINUX_CONSOLE_MODE_DETECT)) && defined(__linux__)
 
 if (!isatty(STDIN_FILENO)) {
 
--- a/nuitka/build/static_src/MainProgram.c
+++ b/nuitka/build/static_src/MainProgram.c
@@ -20,7 +20,7 @@
 #include <unistd.h>
 #endif
 
- #if !defined(_WIN32) && !defined(__APPLE__)
+ #if defined(__linux__)
  #include <errno.h>
  #include <limits.h>
  #include <spawn.h>
--- a/nuitka/build/static_src/MainProgram.c
+++ b/nuitka/build/static_src/MainProgram.c
@@ -20,17 +20,17 @@
 #include <unistd.h>
 #endif
 
- #if defined(__linux__)
- #include <errno.h>
- #include <limits.h>
- #include <spawn.h>
- #include <stdio.h>
- #include <stdlib.h>
- #include <string.h>
- #include <sys/types.h>
- #include <sys/wait.h>
- #include <unistd.h>
- #endif
+#if defined(__linux__)
+#include <errno.h>
+#include <limits.h>
+#include <spawn.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+#endif
 
 #include "nuitka/prelude.h"
 
--- a/nuitka/build/static_src/MainProgram.c
+++ b/nuitka/build/static_src/MainProgram.c
@@ -20,6 +20,18 @@
 #include <unistd.h>
 #endif
 
+ #if !defined(_WIN32) && !defined(__APPLE__)
+ #include <errno.h>
+ #include <limits.h>
+ #include <spawn.h>
+ #include <stdio.h>
+ #include <stdlib.h>
+ #include <string.h>
+ #include <sys/types.h>
+ #include <sys/wait.h>
+ #include <unistd.h>
+ #endif
+
 #include "nuitka/prelude.h"
 
 #ifndef __IDE_ONLY__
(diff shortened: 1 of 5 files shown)


Abdur-rahmaanJ:
Need more time to work on this

nuitka/build/static_src/MainProgram.c
```diff
@@ -1367,6 +1379,81 @@ void setLANGSystemLocaleMacOS(void) {
 #endif
 
 static int Nuitka_Main(int argc, native_command_line_argument_t **argv) {
+
+#if(defined(_NUITKA_LINUX_CONSOLE_MODE_FORCE) || defined(_NUITKA_LINUX_CONSOLE_MODE_DETECT)) && !defined(_WIN32) && !defined(__APPLE__)
+
+if (!isatty(STDIN_FILENO)) {
+
+    const * already_relaunched = getenv("NUITKA_TERMINAL_RELAUNCH");
```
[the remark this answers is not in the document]
kayhayen: The AI is correct here, always go with `const` for all code if you.

nuitka/build/static_src/MainProgram.c
```diff
@@ -1367,6 +1379,81 @@ void setLANGSystemLocaleMacOS(void) {
...
+        if (spawn_argv == NULL) {
+            fprintf(stderr, "Error: Memory allocation failed for xdg-terminal-exec argv.\n");
+            exit(1); // Exit indicating a failure
+        }
+
+
+        spawn_argv[0] = "xdg-terminal-exec";
+        spawn_argv[1] = (char * ) binary_path;
+        for (int i = 1; i < argc; i++) {
+            spawn_argv[i + 1] = argv[i];
```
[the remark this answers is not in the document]
kayhayen: The `native_command_line_argument_t` is not a concern, the code could use it, but the idea for the type is more for cross platform code.

nuitka/build/static_src/MainProgram.c
```diff
@@ -1367,6 +1379,81 @@ void setLANGSystemLocaleMacOS(void) {
 #endif
 
 static int Nuitka_Main(int argc, native_command_line_argument_t **argv) {
+
+#if(defined(_NUITKA_LINUX_CONSOLE_MODE_FORCE) || defined(_NUITKA_LINUX_CONSOLE_MODE_DETECT)) && !defined(_WIN32) && !defined(__APPLE__)
```
kayhayen: You can check for __linux__ no problem here, if we want to add others, like the ones for FreeBSD, we can sure, but right now we seem to limit the OS anyway to Linux, and I am fine for that.

nuitka/build/static_src/MainProgram.c
```diff
@@ -1367,6 +1379,81 @@ void setLANGSystemLocaleMacOS(void) {
...
+        const * binary_path = getBinaryFilenameHostEncoded(true);
+
+        char ** spawn_argv = (char ** ) malloc(sizeof(char * ) * (argc + 2));
+        if (spawn_argv == NULL) {
+            fprintf(stderr, "Error: Memory allocation failed for xdg-terminal-exec argv.\n");
+            exit(1); // Exit indicating a failure
+        }
+
+
+        spawn_argv[0] = "xdg-terminal-exec";
```
kayhayen: The xdg-terminal-exec should be preferred if it exists, but if it does not, and it may well not do so, we should fall back to invoking something else that mimics it, like `x-terminal-emulator` and such, trying to mimic what `xdg-terminal-exec` does.

kayhayen: I am OK with not doing that right away until we are happy with how it works if xdg-terminal-exec exists.

nuitka/build/static_src/MainProgram.c
```diff
@@ -1367,6 +1379,81 @@ void setLANGSystemLocaleMacOS(void) {
...
+        spawn_argv[1] = (char * ) binary_path;
+        for (int i = 1; i < argc; i++) {
+            spawn_argv[i + 1] = argv[i];
+        }
+        spawn_argv[argc + 1] = NULL;
+
+        pid_t pid;
+        extern char ** environ;
+        int spawn_result = posix_spawnp( & pid, "xdg-terminal-exec", NULL, NULL, spawn_argv, environ);
```
kayhayen: Is it not possible to use an exec function like `execv` (or other variant), because why keep this around waiting? The new process can do all the error exiting. Of course if the command is not found and not working, it will be ugly, but mind you, we got no away of talking to the user anyway with no terminal.

kayhayen (commented):
This is pretty good already, but I believe may also be unnecessarily complicated when it comes to forking the process as a subprocess instead of one that replaces the existing one, since then no waiting and checking of exit code is needed.

Abdur-rahmaanJ:
@kayhayen Why are tests failing?

kayhayen:
> @kayhayen Why are tests failing?

Good question. Only 3.13 Linux. Maybe it hates on some of the includes you added causing pythons header files to be confused.

nuitka/OptionParsing.py
```diff
@@ -1842,7 +1842,7 @@ def _handleHelpModes():
     metavar="MODE",
     default="detect",
     help="""\
-Select console mode for Linux app bundles. 'force' always opens Terminal
+Select console mode for Linux. 'force' always opens Terminal
```
kayhayen, line 1990: App bundle is a macOS only term and indeed that's more correct. Also I believe the change is not really related to modes on Linux, since all modes that yield executables can have this code path active.

Abdur-rahmaanJ: I copied Kevin's description, the downsides of copy pasting ...

Abdur-rahmaanJ:
@kayhayen How should i handle all terminals failing in the ci? remove ```exit(1);```?

kayhayen:
First make sure the feature is only active when asked for. Changing the behavior of test cases that don't use the option indicates a lack of proper #ifdef or defines passed. Just check out the command from the actions run that failed.

kayhayen:
This i surprisingly good work, sorry for me getting to this only so late. 

I would want to see 3 things.

a) Pass the command line arguments via an environment variable. If it's present (ideally checking the parent pid if it's relevant) take argv from there in MainProgram.c

b) Sign the CLA please.

kayhayen:
Seems to also need an autoformat invocation before any real tests will be executed.

Abdur-rahmaanJ:
@kayhayen Ok i see only a) and b) for the 3 things. What is the 3rd one?

kayhayen:
I gave up on the third one while writing it, wasn't a concern for Linux after all.

Issue #3701, opened 2025-12-23
Scoping Behavior is not considered in generic functions


typenoob:
# Bug Report

## 🐛 Bug Description

#3648 implemented generic functions python3.12 brings. Howerver it failed to consider the scoping behavior which is very subtle from the original PEP.

Therefore, lots of runtime deviations between Nuitka and Python are exhibited.

One can refer https://jellezijlstra.github.io/pep695#scoping for details.

### 1. Nuitka Version, Python Version, OS, and Platform

```sh
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.12`.
4.0rc2
Commercial: None
Python: 3.12.4 (main, Jul 25 2024, 21:57:00) [MSC v.1929 64 bit (AMD64)]
Flavor: Unknown
Executable: ~\Downloads\nuitka-develop\.venv\Scripts\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 10
Nuitka-Scons:WARNING: Windows SDK must be installed in Visual Studio for it to be usable with
Nuitka-Scons:WARNING: Nuitka. Use the Visual Studio installer for adding it.
Version C compiler: ~\AppData\Local\Nuitka\Nuitka\Cache\DOWNLO~1\gcc\x86_64\14.2.0posix-19.1.1-12.0.0-msvcrt-r2\mingw64\bin\gcc.exe (gcc 14.2.0).
```

### 2. How Nuitka and Python were Installed

NA

### 3. Relevant PyPI Packages and Versions

NA

## 🛠️ To Reproduce

```python
# test.py
def X[T]():
    ...
print(T)
```

## 📉 Expected Behavior

Same as python output:
```cmd
python test.py      
Traceback (most recent call last):
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 112, in _get_module_details
  File "C:\Users\46805\Downloads\nuitka-develop\test.py", line 3, in <module>
    print(T)
          ^
NameError: name 'T' is not defined
```

## 📄 Actual Behavior & Output

```cmd
nuitka test.py
.\test.cmd
T
```

kayhayen:
That's probably using Outline vs. OutlineFunction. We need to use something that has "local variables" for T, which would be removed after they end, and we might be allocating the variable names from the provider instead. Just my guess @ZeroIntensity

ZeroIntensity:
[...]

PR #3704, opened 2025-12-23
Fix generic type scoping

ZeroIntensity:
# What does this PR do?

Fixes the scope of generic type variables in functions.

- #3701

# PR Checklist

## Summary by Sourcery

Bug Fixes:
- Fix incorrect provider used when assigning generic type variables so they are scoped to the function outline body rather than the outer provider.


nuitka/tree/ReformulationFunctionStatements.py
```diff
@@ -238,7 +238,7 @@ def _wrapWithTypeAnnotations(provider, type_params, body, source_ref):
         type_var = buildNode(provider=provider, node=type_param, source_ref=source_ref)
```
typenoob, line 238: Seems TypeVar doesn't take the provider param, do it still has a default provider? cc @kayhayen

kayhayen:
Part of the 4.1 release I just made.

PR #3707, opened 2026-01-01
Fix function type scoping

ZeroIntensity:
# What does this PR do?

Fixes the scope of generic type variables defined for functions. This time it comes with a test so we're more confident about the fix.

- #3701

# PR Checklist

## Summary by Sourcery

Fix handling of generic function type parameters so their type variables are scoped inside the function body rather than at creation time.

Bug Fixes:
- Ensure generic function type variables are bound within the function body scope instead of leaking or being evaluated at function creation.

Tests:
- Add a regression test verifying that a generic function’s type variable is not available at module scope after the function is called.

### DIFF
--- a/tests/basics/Generics312.py
+++ b/tests/basics/Generics312.py
@@ -1,9 +1,10 @@
-def has_generic[T]():
+def someGenericFunction[T]():
     print("hello", T)
 
 
-has_generic()
+someGenericFunction()
 
+# Verify the name didn't leak.
 try:
     print(T)
 except NameError:
@@ -12,9 +13,9 @@ except NameError:
 T = 1
 
 
-def overwrites_generic[T]():
+def someGenericFunctionShadowingGlobal[T]():
     print(T)
 
 
-overwrites_generic()
+someGenericFunctionShadowingGlobal()
 print(T)


nuitka/tree/ReformulationFunctionStatements.py
```diff
@@ -366,6 +355,14 @@ def buildFunctionNode(provider, node, source_ref):
             function_body=code_body, function_statements_body=function_statements_body
         )
 
+    if node.type_params and python_version >= 0x3c0:
```
kayhayen: This has the order of checks reversed, but old Python will crash therefore.

tests/basics/FunctionsTest32.py
```diff
@@ -180,6 +180,24 @@ def g():
 
 print("Using deleted non-local variable:", deletingClosureVariable())
 
+def has_generic[T]():
```
kayhayen: Move this to `Generics312.py` and then it won't crash for SyntaxError for all earlier versions.

kayhayen (changes requested):
Only if you autoformat the code, it will run tests. Run "python misc/install-git-hooks.py" to prevent pushing unformatted code, or run "./bin/autoformat-nuitka-source --un-pushed" manually.

ZeroIntensity:
> Run "python misc/install-git-hooks.py" to prevent pushing unformatted code

Thanks, I missed that.

kayhayen (approved):
I just removed one assertion, which is pointless, as Python is really good at saying "None" had not the attribute

kayhayen:
Also I made the test function names more speaking like I normally do, expressing what's special about them:

```
def someGenericFunction[T]():
    print("hello", T)


someGenericFunction()

# Verify the name didn't leak.
try:
    print(T)
except NameError:
    print("not found")

T = 1


def someGenericFunctionShadowingGlobal[T]():
    print(T)


someGenericFunctionShadowingGlobal()
print(T)
```

typenoob:
This pr broke the original implementation for function type (https://github.com/Nuitka/Nuitka/pull/3648) @ZeroIntensity 

How about to add this case also in tests?
```python
from typing import Callable

def foo[T, **P](function: Callable[P, T]) -> Callable[P, T]:
    print(type(T), type(P))
    return function

def a():
    pass

foo(a)
```

kayhayen:
@ZeroIntensity can we make a new issue out of it, if this holds true, just so we cannot forget

ZeroIntensity:
See #3744.

PR #3831, opened 2026-03-23
Fix and optimize mergeMultipleBranches, add dict index for getVariableDeclarationTop

KRRT7:
## Summary

Optimizes several hot paths in Nuitka's compilation pipeline, measured on `AttributeNodesGenerated.py` (10.7K lines, `--module --generate-c-only`).

### `mergeMultipleBranches` — 4.20s → 1.99s (53% faster)

This function was ~28% of total compile time. Two changes:

1. **Early-exit `for/else` loop.** The original code builds a `set()` of versions across all branches for *every* variable via a generator expression, then checks `len(versions) == 1`. In practice, ~99% of variables have identical versions across all branches (no divergence). The new code iterates `collections[1:]` with a `for/else` — if all match, it assigns directly and `continue`s. Only on mismatch does it fall back to building a set of unique versions.

2. **Bug fix: `escaped.append(escaped)` → `escaped.add(escaped_trace)`.** The original code appends the `escaped` *list to itself* instead of the `escaped_trace` variable. This also changes `escaped` from a list to a set, giving O(1) membership checks in `if trace not in escaped`.

### `getVariableDeclarationTop` — 1.57s → 0.01s (99% faster)

Called 30,669 times. The original linearly scans `variable_declarations_main` and `variable_declarations_heap` to find a declaration by `code_name`. This adds a `variable_declarations_top` dict that's populated incrementally in `addVariableDeclarationFunction` and `addVariableDeclarationTop`, making each lookup O(1) via `dict.get()`.

### `SourceCodeCollector` — minor

- Add `__slots__` to avoid per-instance `__dict__` allocation.
- Replace `__call__` → `emit` indirection with `emit = __call__`.
- Use `append` instead of `extend(code.split("\n"))` — the split is no longer needed.

## Verification

Generated C code is identical before and after (diffed output of `--generate-c-only`).

## Benchmark results

Compiling `nuitka/nodes/AttributeNodesGenerated.py` (10.7K lines, `--module --generate-c-only`):

| Function | Before | After | Improvement |
|---|---|---|---|
| `mergeMultipleBranches` | 4.203s (28.3%) | 1.988s (20.9%) | **-2.22s (53% faster)** |
| `onControlFlowEscape` | 1.245s (8.4%) | 1.122s (11.8%) | -0.12s |
| `getVariableDeclarationTop` | 1.566s (10.5%) | 0.015s (0.2%) | **-1.55s (99% faster)** |
| `mergeBranches` | 0.451s (3.0%) | 0.362s (3.8%) | -0.09s |
| **Total compilation** | **14.84s** | **9.52s** | **-5.32s (36% faster)** |

<details>
<summary>Benchmark script (click to expand)</summary>

```python
#!/usr/bin/env python
"""Benchmark for trace collection hot paths in Nuitka.

Compiles a large file and reports time spent in key functions.
Run on two commits to compare.

Usage:
    PYTHONHASHSEED=0 python -S -X frozen_modules=off bench.py [target_file]

Default target: nuitka/nodes/AttributeNodesGenerated.py (10.7K lines)
"""
import sys
import os
import time
import warnings
import functools

# ── Setup ────────────────────────────────────────────────────────────────

os.environ["PYTHONHASHSEED"] = "0"
sys.path.insert(0, os.getcwd())

target = sys.argv[1] if len(sys.argv) > 1 else "nuitka/nodes/AttributeNodesGenerated.py"
sys.argv = [
    "nuitka", "--module", "--generate-c-only",
    "--output-dir=/tmp/nuitka_bench", target,
]
warnings.simplefilter("ignore", DeprecationWarning)

# Nuitka expects this function injected by __main__.py
_cached = {}
def _getLaunch(name):
    if name not in _cached:
        _cached[name] = os.getenv(name)
        if _cached[name] is not None:
            del os.environ[name]
    v = _cached[name]
    if v is not None and ":" not in v:
        v = None
    if v is not None:
        pid, v = v.split(":", 1)
        try:
            pid = int(pid)
        except ValueError:
            v = None
        else:
            if os.name != "nt" and pid != os.getpid():
                v = None
    return v

import nuitka
nuitka.getLaunchingNuitkaProcessEnvironmentValue = _getLaunch

from nuitka.plugins.Plugins import setupHooks, activatePlugins
from nuitka.options import Options

setupHooks()
Options.parseArgs()
Options.commentArgs()
activatePlugins()


# ── Instrumentation ──────────────────────────────────────────────────────

class FunctionTimer:
    """Wraps a method to accumulate wall-clock time and call count."""
    def __init__(self, name):
        self.name = name
        self.total_time = 0.0
        self.call_count = 0

    def wrap(self, cls, method_name):
        original = getattr(cls, method_name)
        timer = self

        @functools.wraps(original)
        def timed(self_inner, *args, **kwargs):
            t0 = time.perf_counter()
            result = original(self_inner, *args, **kwargs)
            timer.total_time += time.perf_counter() - t0
            timer.call_count += 1
            return result

        setattr(cls, method_name, timed)
        return original


from nuitka.optimizations.TraceCollections import TraceCollectionBase
from nuitka.code_generation.VariableDeclarations import VariableStorage

timers = {}
for method in ("mergeMultipleBranches", "mergeBranches", "onControlFlowEscape"):
    timers[method] = FunctionTimer(method)
    timers[method].wrap(TraceCollectionBase, method)

timers["getVariableDeclarationTop"] = FunctionTimer("getVariableDeclarationTop")
timers["getVariableDeclarationTop"].wrap(VariableStorage, "getVariableDeclarationTop")


# ── Run compilation ──────────────────────────────────────────────────────

from nuitka.MainControl import main as nuitka_main

t_start = time.perf_counter()
try:
    nuitka_main()
except SystemExit:
    pass
t_total = time.perf_counter() - t_start


# ── Report ───────────────────────────────────────────────────────────────

print()
print("=" * 65)
print(f"  Nuitka Trace Collection Benchmark")
print(f"  Target: {target}")
print(f"  Total compilation time: {t_total:.3f}s")
print("=" * 65)
print(f"  {'Function':<30s} {'Calls':>8s} {'Time':>8s} {'% Total':>8s}")
print("-" * 65)

accounted = 0.0
for method, timer in timers.items():
    pct = timer.total_time / t_total * 100 if t_total > 0 else 0
    accounted += timer.total_time
    print(f"  {timer.name:<30s} {timer.call_count:>8d} {timer.total_time:>7.3f}s {pct:>7.1f}%")

print("-" * 65)
other = t_total - accounted
print(f"  {'(other)':<30s} {'':>8s} {other:>7.3f}s {other / t_total * 100:>7.1f}%")
print("=" * 65)
```

</details>

## Test plan

- [x] Verify generated C code is byte-identical before/after
- [ ] Run existing test suite

### DIFF
--- a/nuitka/code_generation/VariableDeclarations.py
+++ b/nuitka/code_generation/VariableDeclarations.py
@@ -136,7 +136,7 @@ class VariableStorage(object):
         "variable_declarations_closure",
         "variable_declarations_locals",
         "exception_variable_name",
-        "_top_declarations_index",
+        "variable_declarations_top",
     )
 
     def __init__(self, heap_name):
@@ -150,8 +150,7 @@ class VariableStorage(object):
 
         self.exception_variable_name = None
 
-        # Dict index for O(1) lookup by code_name in getVariableDeclarationTop.
-        self._top_declarations_index = {}
+        self.variable_declarations_top = {}
 
     @contextmanager
     def withLocalStorage(self):
@@ -169,7 +168,7 @@ class VariableStorage(object):
         self.variable_declarations_locals.pop()
 
     def getVariableDeclarationTop(self, code_name):
-        return self._top_declarations_index.get(code_name)
+        return self.variable_declarations_top.get(code_name)
 
     def getVariableDeclarationClosure(self, closure_index):
         return self.variable_declarations_closure[closure_index]
@@ -225,7 +224,7 @@ class VariableStorage(object):
         result = VariableDeclaration(c_type, code_name, init_value, None)
 
         self.variable_declarations_main.append(result)
-        self._top_declarations_index[code_name] = result
+        self.variable_declarations_top[code_name] = result
 
         return result
 
@@ -237,7 +236,7 @@ class VariableStorage(object):
         else:
             self.variable_declarations_main.append(result)
 
-        self._top_declarations_index[code_name] = result
+        self.variable_declarations_top[code_name] = result
 
         return result
 
--- a/nuitka/code_generation/VariableDeclarations.py
+++ b/nuitka/code_generation/VariableDeclarations.py
@@ -136,6 +136,7 @@ class VariableStorage(object):
         "variable_declarations_closure",
         "variable_declarations_locals",
         "exception_variable_name",
+        "_top_declarations_index",
     )
 
     def __init__(self, heap_name):
@@ -148,6 +149,9 @@ class VariableStorage(object):
         self.variable_declarations_locals = []
 
         self.exception_variable_name = None
+
+        # Dict index for O(1) lookup by code_name in getVariableDeclarationTop.
+        self._top_declarations_index = {}
 
     @contextmanager
     def withLocalStorage(self):
@@ -165,15 +169,7 @@ class VariableStorage(object):
         self.variable_declarations_locals.pop()
 
     def getVariableDeclarationTop(self, code_name):
-        for variable_declaration in self.variable_declarations_main:
-            if variable_declaration.code_name == code_name:
-                return variable_declaration
-
-        for variable_declaration in self.variable_declarations_heap:
-            if variable_declaration.code_name == code_name:
-                return variable_declaration
-
-        return None
+        return self._top_declarations_index.get(code_name)
 
     def getVariableDeclarationClosure(self, closure_index):
         return self.variable_declarations_closure[closure_index]
@@ -229,6 +225,7 @@ class VariableStorage(object):
         result = VariableDeclaration(c_type, code_name, init_value, None)
 
         self.variable_declarations_main.append(result)
+        self._top_declarations_index[code_name] = result
 
         return result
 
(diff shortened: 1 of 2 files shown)


KRRT7:
Thanks for the feedback — pushed fixes in b85958a14:

1. **Reverted `onControlFlowEscape` `% 3 != 2` fast path** — this was incorrect. `LocalVariable.onControlFlowEscape` can call `markActiveVariableAsEscaped` or `markClosureVariableAsUnknown` (not just `markActiveVariableAsUnknown`), so skipping based on `version % 3 == 2` would miss those paths. Only `markActiveVariableAsUnknown` has an internal `% 3 != 2` guard already.

2. **Restored `iterItems()`** — reverted the `.items()` change back to `iterItems()` per project convention.

Also reverted all attribute hoisting (`variable_traces_all = self.variable_traces`, etc.) in the previous commit.

kayhayen:
I rebased the PR, taking the partial merge of Emission with additional improvements in, looking at the other code generation change now. I am a bit concerned that the cache is hiding a preference of one value over another, so an assertion that the cache doesn't get values overwritten will be added for the certainty this doesn't regress. After a release cycle we can remove that maybe.

kayhayen:
I manually merged these ones into commits attributed to you, right now on factory only.

PR #3854, opened 2026-04-08
Python3.12+: Align BUILTIN_SUM1 float summation with CPython compensated sum

### LINKED ISSUE #3843: Nuitka’s Built‑in sum() Produces Incorrect Float Results (Python 3.12+ Regression)
# Bug Report

## 🐛 Bug Description

When compiling Python code with Nuitka **4.1rc7** on **Python 3.13.12**, the built-in `sum()` for floats produces results inconsistent with CPython.

Starting in **Python 3.12**, CPython upgraded `sum()` for float inputs to use an **improved Neumaier-style compensated summation** (see [Issue #100425](https://github.com/python/cpython/issues/100425)).  
Nuitka, however, still lowers `sum()` to its old naive left-fold implementation inside `BUILTIN_SUM1()`, causing **loss of precision and incorrect numerical results**.

This discrepancy is visible even in the simplest example.

---

### Nuitka Version, Python Version, OS, and Platform

```text
$ python -m nuitka --version
4.1rc7
Commercial: None
Python: 3.13.12 (main, Feb  4 2026, 15:06:39) [GCC 15.2.0]
Flavor: Debian Python
GIL: yes
Executable: /usr/bin/python
OS: Linux
Arch: x86_64
Distribution: Debian sid
Version C compiler: gcc (gcc 15).
```

### How Nuitka and Python Were Installed

- Debian system Python (sid)
- Nuitka installed via `pip`

### Relevant PyPI Packages

None required — bug reproduces using only Python built-ins.

---

## 🛠️ To Reproduce

SSCCE

```python
x = [0.1] * 10
print(repr(sum(x)))
```

Running under CPython

```
$ python t.py
1.0
```

Running the same code compiled by Nuitka

```
$ nuitka --module t.py
$ python3 -c 'import t'
0.9999999999999999
```

## 📉 Expected Behavior

Nuitka should match CPython 3.12+ behavior: 1.0

CPython uses the improved Neumaier compensated summation algorithm (implemented via PR #100426 linked from Issue #100425).

## 📄 Actual Behavior & Output

Nuitka emits the older, naive summation: 0.9999999999999999

This is due to Nuitka performing builtin-call optimization and replacing `sum(x)` with an internal naïve left fold (`BUILTIN_SUM1()`), which does not implement the compensation step used in CPython 3.12+.

As a result, floating-point drift appears immediately and accumulates across larger sums.

sealpp:
# Description

## ❓ What does this PR do?

Align `BUILTIN_SUM1()` with CPython 3.12+ float summation semantics by switching the Python 3.12+ exact-float path to compensated summation after the integer fast path has already produced a float result.

This keeps the existing integer/bool fast path unchanged, leaves `BUILTIN_SUM2()` untouched, and adds regression coverage in `tests/basics/BuiltinsTest.py` for float, generator, mixed int/float, mixed bool/float, `inf`, and overflow cases.

Fixes https://github.com/Nuitka/Nuitka/issues/3843.

On Python 3.12+, CPython changed builtin `sum()` for exact-float inputs to use compensated summation (`python/cpython` issue `#100425`, PR `#100426`). Nuitka still used the older naive accumulation in `BUILTIN_SUM1()`, which caused compiled results such as `sum([0.1] * 10)` to differ from CPython.

## 🧱 Type of Change

- [ ] 🏗️ Build / CI System
- [ ] 📚 Documentation Update

## ✅ PR Checklist

- [ ] **Formatting**: Enabled commit hook or executed `./bin/autoformat-nuitka-source`.
- [ ] **Tests**: All tests still pass. (See [Running the Tests](https://nuitka.net/doc/developer-manual.html#running-the-tests)).

## 🤖 AI Generated Code Policy

  - [ ] **Issue First**: I have created an issue explaining the problem *before* using AI to generate a fix, ensuring the direction is correct.
    > **Note**: AI generated code is welcome, but it must be peer-reviewed and understood by the submitter. Blindly copying AI output without understanding is not acceptable.

## Summary by Sourcery

Align Python 3.12+ builtin sum() float behavior with CPython by updating BUILTIN_SUM1 and extending regression coverage.

Bug Fixes:
- Match Python 3.12+ CPython compensated float summation semantics in BUILTIN_SUM1 to avoid numerical discrepancies for float-heavy sums.

Tests:
- Add regression tests for float and generator sums, mixed int/float and bool/float inputs, infinities, and large float overflow scenarios to verify sum() behavior.

### DIFF
--- a/nuitka/build/static_src/CompiledCodeHelpers.c
+++ b/nuitka/build/static_src/CompiledCodeHelpers.c
@@ -1470,6 +1470,84 @@ PyObject *BUILTIN_SUM1(PyThreadState *tstate, PyObject *sequence) {
         return NULL;
     }
 
+#if PYTHON_VERSION >= 0x3c0
+    if (PyFloat_CheckExact(result)) {
+        double f_result = PyFloat_AS_DOUBLE(result);
+        double c = 0.0;
+        Py_SETREF(result, NULL);
+
+        for (;;) {
+            bool finished;
+            item = QUICK_ITERATOR_NEXT(tstate, &qiter, &finished);
+
+            if (finished) {
+                if (c && Py_IS_FINITE(c)) {
+                    f_result += c;
+                }
+
+                return PyFloat_FromDouble(f_result);
+            } else if (item == NULL) {
+                return NULL;
+            }
+
+            CHECK_OBJECT(item);
+
+            if (PyFloat_CheckExact(item)) {
+                /* Improved Kahan-Babuska algorithm by Arnold Neumaier. */
+                double x = PyFloat_AS_DOUBLE(item);
+                double t = f_result + x;
+
+                if (fabs(f_result) >= fabs(x)) {
+                    c += (f_result - t) + x;
+                } else {
+                    c += (x - t) + f_result;
+                }
+
+                f_result = t;
+                _Py_DECREF_SPECIALIZED(item, _PyFloat_ExactDealloc);
+
+                continue;
+            }
+
+            if (PyLong_Check(item)) {
+                long value;
+                int overflow;
+
+                value = PyLong_AsLongAndOverflow(item, &overflow);
+
+                if (!overflow) {
+                    f_result += (double)value;
+                    Py_DECREF(item);
+
+                    continue;
+                }
+            }
+
+            if (c && Py_IS_FINITE(c)) {
+                f_result += c;
+            }
+
+            result = PyFloat_FromDouble(f_result);
+
+            if (result == NULL) {
+                Py_DECREF(item);
+                return NULL;
+            }
+
+            PyObject *temp2 = PyNumber_Add(result, item);
+            Py_DECREF(result);
+            Py_DECREF(item);
+            result = temp2;
+
+            if (unlikely(result == NULL)) {
+                return NULL;
+            }
+
+            break;
+        }
+    }
+#endif
+
     for (;;) {
         CHECK_OBJECT(result);
 
--- a/nuitka/build/static_src/CompiledCodeHelpers.c
+++ b/nuitka/build/static_src/CompiledCodeHelpers.c
@@ -1504,7 +1504,7 @@ PyObject *BUILTIN_SUM1(PyThreadState *tstate, PyObject *sequence) {
                 }
 
                 f_result = t;
-                _Py_DECREF_SPECIALIZED(item, _PyFloat_ExactDealloc);
+                Py_DECREF(item);
 
                 continue;
             }
--- a/nuitka/build/static_src/CompiledCodeHelpers.c
+++ b/nuitka/build/static_src/CompiledCodeHelpers.c
@@ -1419,7 +1419,7 @@ PyObject *BUILTIN_SUM1(PyThreadState *tstate, PyObject *sequence) {
 #if PYTHON_VERSION >= 0x270
         if (PyLong_CheckExact(item)) {
             int overflow;
-            long b = PyLong_AsLongAndOverflow(item, &overflow);
+            long b = Nuitka_PyLong_AsLongAndOverflow(item, &overflow);
 
             if (overflow) {
                 break;
@@ -1485,7 +1485,7 @@ PyObject *BUILTIN_SUM1(PyThreadState *tstate, PyObject *sequence) {
                     f_result += c;
                 }
 
-                return PyFloat_FromDouble(f_result);
+                return MAKE_FLOAT_FROM_DOUBLE(f_result);
             } else if (item == NULL) {
                 return NULL;
             }
@@ -1510,13 +1510,24 @@ PyObject *BUILTIN_SUM1(PyThreadState *tstate, PyObject *sequence) {
             }
 
             if (PyLong_Check(item)) {
-                long value;
                 int overflow;
-
-                value = PyLong_AsLongAndOverflow(item, &overflow);
+                long value = Nuitka_PyLong_AsLongAndOverflow(item, &overflow);
 
                 if (!overflow) {
+#if PYTHON_VERSION >= 0x3e0
+                    double x = (double)value;
+                    double t = f_result + x;
+
+                    if (fabs(f_result) >= fabs(x)) {
+                        c += (f_result - t) + x;
+                    } else {
+                        c += (x - t) + f_result;
+                    }
+
+                    f_result = t;
+#else
                     f_result += (double)value;
+#endif
                     Py_DECREF(item);
 
                     continue;
@@ -1527,7 +1538,7 @@ PyObject *BUILTIN_SUM1(PyThreadState *tstate, PyObject *sequence) {
                 f_result += c;
             }
 
-            result = PyFloat_FromDouble(f_result);
+            result = MAKE_FLOAT_FROM_DOUBLE(f_result);
 
             if (result == NULL) {
                 Py_DECREF(item);
--- a/nuitka/build/include/nuitka/helper/ints.h
+++ b/nuitka/build/include/nuitka/helper/ints.h
@@ -5,6 +5,7 @@
 
 // Our "PyLong_FromLong" replacement.
 extern PyObject *Nuitka_PyLong_FromLong(long ival);
+extern long Nuitka_PyLong_AsLongAndOverflow(PyObject *value, int *overflow);
 
 // Our "PyInt_FromLong" replacement, not done (yet?).
 #if PYTHON_VERSION >= 0x300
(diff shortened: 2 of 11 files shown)


nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -1470,6 +1470,84 @@ PyObject *BUILTIN_SUM1(PyThreadState *tstate, PyObject *sequence) {
...
+                double t = f_result + x;
+
+                if (fabs(f_result) >= fabs(x)) {
+                    c += (f_result - t) + x;
+                } else {
+                    c += (x - t) + f_result;
+                }
+
+                f_result = t;
+                _Py_DECREF_SPECIALIZED(item, _PyFloat_ExactDealloc);
```
kayhayen: For Nuitka, this may actually be slower on some platforms, e.g. Windows could break down making API calls suddenly.

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -1470,6 +1470,84 @@ PyObject *BUILTIN_SUM1(PyThreadState *tstate, PyObject *sequence) {
         return NULL;
     }
 
+#if PYTHON_VERSION >= 0x3c0
```
kayhayen, line 1473: I thought we could actually not limit this to 3.12, but do it for all versions, as it's just an optimization, or at least that's what I am thinking now.

nuitka/build/static_src/CompiledCodeHelpers.c
```diff
@@ -1470,6 +1470,84 @@ PyObject *BUILTIN_SUM1(PyThreadState *tstate, PyObject *sequence) {
...
+
+                    continue;
+                }
+            }
+
+            if (c && Py_IS_FINITE(c)) {
+                f_result += c;
+            }
+
+            result = PyFloat_FromDouble(f_result);
```
kayhayen, line 1530: We got a better method for that I believe, or at least we ought to.

kayhayen (changes requested):
Looks pretty good, I still need to review it in its full context and I am not so sure that BUILTIN_SUM2 would not be affected as well.

kayhayen:
>     * The `#if PYTHON_VERSION >= 0x3c0` guard looks wrong for targeting Python 3.12+ (should be `0x30c` for 3.12), so this new compensated-sum path will never be compiled in for current Python versions.

Very wrong of course, the check is correct as it is.

kayhayen:
I need to confirm with looking at the CPython 3.12 code, why builtin sum with 2 arguments is not done. Probably because it's more oriented to containers, they left that out. But not sure if there is not a common implementation in CPython right now.

nuitka/build/static_src/CompiledCodeHelpers.c
sealpp: I took the same concern from that comment and reduced the follow-up to the lowest-risk change. The only new low-level dependency introduced by this PR was the private float dealloc path, so I removed that and switched it to plain `Py_DECREF(item)` instead. That keeps the 3.12+ compensated summation behavior while avoiding extra platform risk from CPython-private deallocation details.

nuitka/build/static_src/CompiledCodeHelpers.c
sealpp: I am keeping this PR limited to Python 3.12+ on purpose. The goal here is semantic alignment with released CPython 3.12+ behavior, not a broader optimization change for older versions. I would prefer to keep any "do this for all versions" idea as a separate discussion, and I rechecked Python 3.10 locally to make sure the current gate does not affect pre-3.12 behavior.

nuitka/build/static_src/CompiledCodeHelpers.c
sealpp: For this follow-up I kept the implementation intentionally conservative. I did not introduce any additional private headers or lower-level CPython coupling, and I only removed the private dealloc dependency that was causing risk. If we want a broader or more optimized approach later, I think that should be evaluated separately from this compatibility fix.

kayhayen: As a comment it's potentially often a reminder to myself, to do that at merge time. The function is called `MAKE_FLOAT_FROM_DOUBLE`` and will be way faster on >=3.10 as it avoids making an API call and quickly just goes to the float object cache added in that version, and if not, allocates it manually.

For the performance side of things, I keep a tap on this, for compatability of course, these things are only risks, and I accept these to not happen.

In code this looks like this:

```
// Replacement for PyFloat_FromDouble that is faster
extern PyObject *MAKE_FLOAT_FROM_DOUBLE(double value);
#else
#define NUITKA_FLOAT_HAS_FREELIST 0
#define MAKE_FLOAT_FROM_DOUBLE(value) PyFloat_FromDouble(value)
#endif
```

One weakness of these replacements is that they got inconsistent naming (`Nuitka_FloatFromDouble` would be better) and often have no comment linking it, this one does, and such. One day I need to clean that up and make it a coding rule to check this, such that an AI can do these automatically for me. And then maybe over time expand the set of mappings.

nuitka/build/static_src/CompiledCodeHelpers.c
kayhayen: Actually, I believe we ought to have Py_DECREF_FLOAT and the like, because on some platforms, of course, this could be crazy faster. I need to look at `_Py_DECREF_SPECIALIZED` how it works., and maybe we can even have Nuitka_Float_ExactDealloc and be good with it.

But for now, I would much rather merge it like you did that, and only work on that in this PR if it turns out super easy. At least for 3.14, it looks as if it would be:

```
void
_PyFloat_ExactDealloc(PyObject *obj)
{
    assert(PyFloat_CheckExact(obj));
    _Py_FREELIST_FREE(floats, obj, PyObject_Free);
}
```

I do like it also when code annotates the type, and frequently generated code of Nuitka could avoid a little bit of pointer indirectly and release tuples and list typed values directly. For now it complicates no-GIL as there it needs to be done differently, so I am probably staying away for it right away. And also the Win32 difference of macro vs. function is there again, but we might not have to touch `_Py_DECREF_SPECIALIZED`.

kayhayen (approved):

kayhayen:
I will take it from there, apply the function I meantion, with TODOs added for the things I commented on regarding the specialization or just try my hand on those right away.

kayhayen:
One thing noteworthy is that CPython currently barely uses Py_DECREF_SPECIALIZED, while for Nuitka it could in theory be very important. If anything generated code will become smaller by using it.

The macros make this most obvious:

```
#undef Py_DECREF
#define Py_DECREF(arg) \
    do { \
        PyObject *op = _PyObject_CAST(arg); \
        if (_Py_IsImmortal(op)) { \
            _Py_DECREF_IMMORTAL_STAT_INC(); \
            break; \
        } \
        _Py_DECREF_STAT_INC(); \
        if (--op->ob_refcnt == 0) { \
            _PyReftracerTrack(op, PyRefTracer_DESTROY); \
            destructor dealloc = Py_TYPE(op)->tp_dealloc; \
            (*dealloc)(op); \
        } \
    } while (0)

#undef _Py_DECREF_SPECIALIZED
#define _Py_DECREF_SPECIALIZED(arg, dealloc) \
    do { \
        PyObject *op = _PyObject_CAST(arg); \
        if (_Py_IsImmortal(op)) { \
            _Py_DECREF_IMMORTAL_STAT_INC(); \
            break; \
        } \
        _Py_DECREF_STAT_INC(); \
        if (--op->ob_refcnt == 0) { \
            _PyReftracerTrack(op, PyRefTracer_DESTROY); \
            destructor d = (destructor)(dealloc); \
            d(op); \
        } \
    } while (0)
````

It will be saving going through the type object to find the function, but rather call it directly. For the C compiler this means, it needs to decide if it wants to inline it, which for generated code could become a hard issue with LTO, which is why they make it macros for MSVC to begin with, I believe. But if that holds true, we could still not doing it for the very bad C compiler there, and gain elsewhere or just suffer from the slow down.

kayhayen:
I will merge that for the next hotfix, 4.0.9 only, since I don't want to take the risk before running my CPython test suites internally. In terms of risk it ought to be small, but you never know before you tried.

sealpp:
Thanks, that sounds good to me.

kayhayen:
I rebased it onto current develop branch.

kayhayen:
The Python 3.14 behaves different as well, I added support for that as well, and it uses more of our own API, and `Nuitka_PyLong_AsLongAndOverflow` to replace `PyLong_AsLongAndOverflow` to avoid even more API calls.

kayhayen:
Merged for the 4.1 release to come.

Issue #3856, opened 2026-04-15
Accessing `__type_params__` results in a segfault


ZeroIntensity:
This will be part of #3814.

Here's a simple repro:

```py
class TestClass[T, *Ts, **P]:
    pass

print(TestClass.__type_params__)  # Crash!
```

I'll be working on this.

kayhayen:
This is solved with the PR merge and in develop, will be in 4.1 release.

kayhayen:
Part of the 4.1 release I just made.

PR #3866, opened 2026-04-24
Fix the `__type_params__` attribute for functions

ZeroIntensity:
# Description

## ❓ What does this PR do?

Adds support for the `__type_params__` attribute on compiled function objects. Previously, it would just return an empty tuple.

- #3856

## 🧱 Type of Change

  expected)
- [ ] 🏗️ Build / CI System
- [ ] 📚 Documentation Update

## ✅ PR Checklist

## 🤖 AI Generated Code Policy

## Summary by Sourcery

Propagate function type parameter information into compiled function objects so the __type_params__ attribute reflects actual type parameters instead of always being empty.

Bug Fixes:
- Fix compiled function __type_params__ to return the function's real type parameters rather than an empty tuple.
- Ensure compiled functions correctly manage lifetime of stored type parameter tuples to avoid reference count issues.

Enhancements:
- Extend function creation nodes, mixins, and code generation paths to carry optional type parameter information from parsing through to compiled function instances.


kayhayen:
It seems coroutines, etc. are not done yet according to checks. Lets add that and see. To me this looks good. I am a bit unsure if that change should not be guarded with a version guard, like 3.12 or something in that area, since it doesn't exist for earlier versions.

kayhayen:
Also sourcery is pretty correct in its findings, esp. the cloning must not loose information.

ZeroIntensity:
Fixed the lint issues. I still need to figure out how to get this working for coroutines.

> I am a bit unsure if that change should not be guarded with a version guard, like 3.12 or something in that area, since it doesn't exist for earlier versions.

Yeah, I omitted the version guard on parameters for simplicity's sake. I think it's fine as long as the details don't leak into other versions.

kayhayen:
Let me know when you consider it ready, not to rush you, just so I don't miss if that's already the case.

ZeroIntensity:
This should be ready now @kayhayen

nuitka/build/include/nuitka/compiled_function.h
```diff
@@ -102,7 +102,7 @@ extern struct Nuitka_FunctionObject *Nuitka_Function_New(function_impl_code c_co
                                                          PyCodeObject *code_object, PyObject *defaults,
                                                          PyObject *kw_defaults, PyObject *annotations, PyObject *module,
                                                          PyObject *doc, struct Nuitka_CellObject **closure,
-                                                         Py_ssize_t closure_given);
+                                                         Py_ssize_t closure_given, PyObject *type_params);
```
kayhayen, line 105: Ideally it's only present for the Python version that needs it, not for older versions.

nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -678,9 +678,13 @@ static PyObject *Nuitka_Function_get_type_params(PyObject *self, void *data) {
     assert(Nuitka_Function_Check(self));
     assert(_PyObject_GC_IS_TRACKED(self));
 
-    // TODO: Probably not needed anymore?
-    Py_INCREF(const_tuple_empty);
-    return const_tuple_empty;
+    struct Nuitka_FunctionObject *function = (struct Nuitka_FunctionObject *)self;
+    if (function->m_type_params == NULL) {
+        assert(_Py_IsImmortal(const_tuple_empty));
```
kayhayen: Py_INCREF_IMMORTAL does that check too, not sure if we need to keep in mind, that this is 3.12 only code, or just do it like that. I would vote for that, since it's the pattern to use for common code.

nuitka/build/static_src/CompiledFunctionType.c
```diff
@@ -857,7 +861,7 @@ static PyObject *Nuitka_Function_clone(struct Nuitka_FunctionObject *function, P
 #if PYTHON_VERSION >= 0x300
                             kwdefaults, annotations,
 #endif
-                            function->m_module, function->m_doc, function->m_closure, function->m_closure_given);
+                            function->m_module, function->m_doc, function->m_closure, function->m_closure_given, NULL);
```
kayhayen: The cloning needs to pass the type params too, does it not?

nuitka/code_generation/FunctionCodes.py
```diff
@@ -295,11 +321,13 @@ def handleDefaults():
         kw_defaults_name=kw_defaults_name,
         annotations_name=annotations_name,
         closure_variables=expression.getClosureVariableVersions(),
+        type_params_name=type_params_name,
         emit=emit,
         context=context,
     )
 
     getReleaseCode(release_name=annotations_name, emit=emit, context=context)
+    getReleaseCode(release_name=type_params_name, emit=emit, context=context)
```
kayhayen: getReleaseCode accepts also release_names

nuitka/nodes/FunctionNodes.py
```diff
@@ -41,7 +41,7 @@
 from .ChildrenHavingMixins import (
     ChildHavingBodyOptionalMixin,
     ChildrenHavingDefaultsTupleFunctionRefMixin,
-    ChildrenHavingDefaultsTupleKwDefaultsOptionalAnnotationsOptionalFunctionRefMixin,
+    ChildrenHavingDefaultsTupleKwDefaultsOptionalAnnotationsOptionalFunctionRefTypeParamsOptionalMixin,
```
kayhayen: This has a nicer name that does express its unique content, ChildrenExpressionFunctionCreationMixin

nuitka/nodes/FunctionNodes.py
```diff
@@ -885,7 +885,7 @@ def getFunctionInlineCost(self, values):
 
 
 def makeExpressionFunctionCreation(
-    function_ref, defaults, kw_defaults, annotations, source_ref
+    function_ref, defaults, kw_defaults, annotations, source_ref, type_params=None
```
kayhayen: Default argument feels like bug hiding

kayhayen (changes requested):

nuitka/build/include/nuitka/compiled_function.h
ZeroIntensity: Changing it does add some needless complexity to this PR though -- is it important that we don't do this for older versions?

kayhayen: Yes, and optional arguments are not actually more complexity. They help answering what the argument value should be for versions that don't have something.

nuitka/nodes/FunctionNodes.py
kayhayen: I feel like that's also a cleanup waiting to happen, seeing the other imports now. But lets not make it worse.

PR #3869, opened 2026-05-03
Add [project] table to pyproject.toml for uv/PEP 621 support

KRRT7:
## Summary

- Populate `pyproject.toml` with a full `[project]` table (PEP 621) ported from `setup.py`: metadata, dependencies, scripts, entry-points, optional-dependencies, and URLs
- Add a `[dependency-groups]` `dev` group consolidating `requirements-devel.txt` and `requirements-private.txt` for local development
- Set `build-backend = "setuptools.build_meta:__legacy__"` so the existing `setup.py` continues to handle package-data, inline copies, and `BinaryDistribution`
- Add `uv.lock` for reproducible installs

This enables `uv sync --group dev` for local development. The existing `setup.py` build logic is unchanged.

## Test plan

- [ ] `uv sync --group dev` resolves and installs successfully
- [ ] `uv run nuitka --version` prints correct version
- [ ] `pip install -e .` still works via legacy setuptools path
- [ ] Existing CI/CD pipelines pass without changes

## Summary by Sourcery

Add PEP 621 project metadata and development dependency groups to pyproject.toml and introduce a lockfile for reproducible installs.

New Features:
- Define the PEP 621-compliant [project] table in pyproject.toml including metadata, runtime dependencies, console scripts, entry points, optional dependencies, and URLs.
- Introduce a dev dependency group in pyproject.toml for local development tooling and test dependencies.

Build:
- Set the build backend to the legacy setuptools.build_meta:__legacy__ backend to preserve the existing setup.py-based build behavior.
- Add a uv.lock file to support reproducible dependency installation with uv.

### DIFF
--- a/.github/workflows/testing-debug.yml
+++ b/.github/workflows/testing-debug.yml
@@ -4,6 +4,13 @@ on:
   pull_request:
     branches:
       - develop
+    paths-ignore:
+      - "doc/**"
+      - "*.md"
+      - "*.rst"
+      - "LICENSE"
+      - ".github/ISSUE_TEMPLATE/**"
+      - ".github/FUNDING.yml"
   push:
     branches:
       - main
@@ -11,6 +18,13 @@ on:
       - factory
       - release/**
       - hotfix/**
+    paths-ignore:
+      - "doc/**"
+      - "*.md"
+      - "*.rst"
+      - "LICENSE"
+      - ".github/ISSUE_TEMPLATE/**"
+      - ".github/FUNDING.yml"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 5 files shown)


kayhayen (commented):
I am not against it, but so far I thought I cannot have both things, setup.py and "pyproject.toml" with actual content. 

This comes across as a terrible duplication of the code we have. I hope we can use some sort of generators to avoid it.

kayhayen:
No sure about the "dev" thing, I am trying hard to get away from polluted python installations from Nuitka dependencies. Also I thought "onefile" was a dependency group, why is "dev" not among it, and is that a group that should be present or not. I personally like the private pip space for a tool like Nuitka. Installing "imageio" or "zstandard" globally is a mistake that I mean to rectify through it.

kayhayen:
This is an example of a PR not following the AI guide lines.

I want an issue with a discussion first, so I even know why I should review this. As it stands, this is useless without my knowing what problem it solves. Please create issues before creating PRs with solutions that don't fit the project goals.

PR #3882, opened 2026-05-13
Comparisons: Reduce generated helpers via INT/CLONG shortcuts and safe inversion

KRRT7:
# Description

## ❓ What does this PR do?

Enable the codegen to emit fewer comparison helper functions for `INT`/`CLONG` pairs by using the existing shortcut mechanism. Previously `INT`/`CLONG` generated all six comparison variants (`LT`, `LE`, `EQ`, `NE`, `GT`, `GE`); with `shortcut=True` only the three-element subset `{LT, LE, EQ}` is generated, and `GT`/`GE`/`NE` are handled at codegen time by inverting to the matching subset helper and negating the result.

The inversion logic in `_handleArgumentSwapAndInversion` was previously too broad. It would also invert float and container comparisons, which is wrong because `not (a > b) != a <= b` when NaN is involved. A new guard `_canInvertComparisonToHelperSubset` restricts inversion to type families with a guaranteed total order (integer and string shapes), making the optimization safe to enable for `INT`/`CLONG` without regressing float/container correctness.

#### `ComparisonHelperDefinitions.py` - enable shortcut for INT/CLONG

```python
# before
_makeFriendOps("INT", "CLONG", may_raise=False, shortcut=False)
# after
_makeFriendOps("INT", "CLONG", may_raise=False, shortcut=True)
```

`shortcut=True` tells the helper set to include only `{LT, LE, EQ}` for this pair. `GT`/`GE`/`NE` are no longer in `specialized_cmp_helpers_set`, so their implementations are deleted (~188 lines across `.c`/`.h`).

`LONG`/`INT` mixed pairs remain `shortcut=False` because enabling shortcuts there would also drop the internal `LONG_CLONG` GT/GE/NE helpers that the NILONG dual-type helpers still require.

#### `ComparisonCodes.py` - restrict inversion to safe shapes

`_canInvertComparisonToHelperSubset` returns `True` only when inversion is known-safe:

- `PyObject * / CTypeCLong` mixed pair - the new INT/CLONG case
- Same `PyObject *` type where both shapes are in a totally-ordered family: `(tshape_int, tshape_long)`, `(tshape_str, tshape_unicode)`, `(tshape_bytes,)`

Float, unknown, container, and all other shapes return `False`. The codegen no longer inverts those comparators, so NaN cannot produce wrong results.

#### Dead code removed (`.c` / `.h`)

`RICH_COMPARE_{GE,GT,NE}_{OBJECT,CBOOL}_INT_CLONG` and their extern declarations are deleted. They are now unreachable since the helper set no longer includes GT/GE/NE for INT/CLONG.

#### `BuiltinsTest.py` - runtime coverage

- Ordered NaN comparisons in object-result and condition-result contexts, both operand orders, and `not (nan <= x)` / `not (x <= nan)` inversion forms.
- `int` object vs `nan` comparisons.
- Python 2-only mixed `int`/`long` coverage, guarded by `if str is bytes:` to avoid `long` literal syntax errors on Python 3.

This was initiated to reduce generated comparison helpers while keeping comparison inversion safe for values such as NaN and containers where the logical inversion is not equivalent.

## 🧱 Type of Change

- [ ] 🏗️ Build / CI System
- [ ] 📚 Documentation Update

## ✅ PR Checklist

- [x] **Tests**: All tests still pass. (See [Running the Tests](https://nuitka.net/doc/developer-manual.html#running-the-tests)).
  - See validation commands below.

## Validation

Property-based exploration with `crosshair-tool` and `hypothesis` in `tests/scratch/` confirmed which inversions are unsafe.

CrossHair found NaN counterexamples for scalar float and tuple comparisons:

```bash
uv run crosshair diffbehavior --per_condition_timeout=10 tests.scratch.crosshair_compare.gt_original tests.scratch.crosshair_compare.gt_inverted
uv run crosshair diffbehavior --per_condition_timeout=10 tests.scratch.crosshair_compare.ge_original tests.scratch.crosshair_compare.ge_inverted
uv run crosshair diffbehavior --per_condition_timeout=10 tests.scratch.crosshair_compare.tuple_gt_original tests.scratch.crosshair_compare.tuple_gt_inverted
```

Examples found:

```text
gt_original(-1.0, nan) -> False
gt_inverted(-1.0, nan) -> True

ge_original(7.4050824148778e-304, nan) -> False
ge_inverted(7.4050824148778e-304, nan) -> True

tuple_gt_original((1.184418094435452e-308,), (nan,)) -> False
tuple_gt_inverted((1.184418094435452e-308,), (nan,)) -> True
```

Hypothesis found minimized counterexamples, including the list case:

```bash
uv run python tests/scratch/hypothesis_compare.py
```

Minimized examples:

```text
float gt: (0.0, nan)
float ge: (0.0, nan)
tuple gt: ((0.0,), (nan,))
list gt: ([0.0], [nan])
```

Testing performed:

- `./bin/autoformat-nuitka-source --un-pushed`
- `./.venv/bin/python ./tests/basics/run_all.py only BuiltinsTest.py`
- `./bin/check-nuitka-with-pylint --un-pushed`
- `git diff --check`
- `NUITKA_CACHE_DIR=tests/scratch/nuitka-cache python3 tests/basics/run_all.py only BuiltinsTest`
- `NUITKA_CACHE_DIR=tests/scratch/nuitka-cache PYTHON=python NUITKA_EXTRA_OPTIONS="--lto=no --disable-cache=ccache" python tests/basics/run_all.py only BuiltinsTest`

Python 2.7.18 validation passed after disabling local LTO/ccache due to an unrelated LLVM bitcode linker mismatch with the default macOS toolchain.

## 🤖 AI Generated Code Policy

  - [ ] **Issue First**: I have created an issue explaining the problem *before* using AI to generate a fix, ensuring the direction is correct.

### DIFF
--- a/nuitka/build/include/nuitka/helper/comparisons_ge.h
+++ b/nuitka/build/include/nuitka/helper/comparisons_ge.h
@@ -157,16 +157,6 @@ extern PyObject *RICH_COMPARE_GE_OBJECT_LONG_INT(PyObject *operand1, PyObject *o
 extern bool RICH_COMPARE_GE_CBOOL_LONG_INT(PyObject *operand1, PyObject *operand2);
 #endif
 
-#if PYTHON_VERSION < 0x300
-/* Code referring to "INT" corresponds to Python2 'int' and "CLONG" to C platform long value. */
-extern PyObject *RICH_COMPARE_GE_OBJECT_INT_CLONG(PyObject *operand1, long operand2);
-#endif
-
-#if PYTHON_VERSION < 0x300
-/* Code referring to "INT" corresponds to Python2 'int' and "CLONG" to C platform long value. */
-extern bool RICH_COMPARE_GE_CBOOL_INT_CLONG(PyObject *operand1, long operand2);
-#endif
-
 /* Code referring to "LONG" corresponds to Python2 'long', Python3 'int' and "DIGIT" to C platform digit value for long
  * Python objects. */
 extern PyObject *RICH_COMPARE_GE_OBJECT_LONG_DIGIT(PyObject *operand1, long operand2);
--- a/nuitka/build/include/nuitka/helper/comparisons_gt.h
+++ b/nuitka/build/include/nuitka/helper/comparisons_gt.h
@@ -157,16 +157,6 @@ extern PyObject *RICH_COMPARE_GT_OBJECT_LONG_INT(PyObject *operand1, PyObject *o
 extern bool RICH_COMPARE_GT_CBOOL_LONG_INT(PyObject *operand1, PyObject *operand2);
 #endif
 
-#if PYTHON_VERSION < 0x300
-/* Code referring to "INT" corresponds to Python2 'int' and "CLONG" to C platform long value. */
-extern PyObject *RICH_COMPARE_GT_OBJECT_INT_CLONG(PyObject *operand1, long operand2);
-#endif
-
-#if PYTHON_VERSION < 0x300
-/* Code referring to "INT" corresponds to Python2 'int' and "CLONG" to C platform long value. */
-extern bool RICH_COMPARE_GT_CBOOL_INT_CLONG(PyObject *operand1, long operand2);
-#endif
-
 /* Code referring to "LONG" corresponds to Python2 'long', Python3 'int' and "DIGIT" to C platform digit value for long
  * Python objects. */
 extern PyObject *RICH_COMPARE_GT_OBJECT_LONG_DIGIT(PyObject *operand1, long operand2);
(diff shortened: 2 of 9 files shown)


kayhayen:
I was turned off by the logic that sourcery describes as badly written too. I do like the reduction. I am not sure if we have those with CLONG and Python longs too. I guess, step one is to isolate the bug report that's in here. Very unfortunate that this is a lengthy description and didn't follow the PR template to create an issue first, then it would be a lot easier to review.

kayhayen:
Ok, so inversion is not allowed with floats, but currently it's not possible to trigger the bug, so that concern is off the table, but it surely is worth addressing, we want to get there of course, where it will matter.

KRRT7:
>  but currently it's not possible to trigger the bug

what bug exactly is not possible to trigger?

PR #3890, opened 2026-05-21
Fusion: sum(x for x in range(...)) identity genexpr + any/all(range(n))

KRRT7:
# Description

## ❓ What does this PR do?

Fuses `sum(x for x in range(...))` into direct C arithmetic, extending the existing `sum(range(...))` fusion. Also adds identically-patterned fusions for `any(range(n))` and `all(range(n))`.

**Identity genexpr fusion** — The Nuitka optimizer previously fused `sum(range(stop))`, `sum(range(start,stop))`, `sum(range(start,stop,step))` into closed-form arithmetic via `BUILTIN_SUM_RANGE1/2/3` helpers. This PR extends that to recognize the identity generator-expression pattern `sum(x for x in range(...))` and lower it equivalently, eliminating the generator object, iterator, and per-element `next()` calls entirely.

**`any(range(n))` / `all(range(n))` fusion** — Transforms `any(range(n))` → `RICH_COMPARE_GE(n, 2)` (true if `n >= 2`, i.e. range has at least one non-zero element for `any`, or at least two elements since only 0 is falsy) and `all(range(n))` → `RICH_COMPARE_LE(n, 0)` (true if `n <= 0`, i.e. range is empty). No special code generation is needed — the existing `RICH_COMPARE_*` nodes handle it.

All fused forms match CPython semantics exactly for all integer inputs.

## 🧐 Why was it initiated?

These are natural optimization opportunities that users expect — generator expressions over `range` inside `sum`/`any`/`all` are common Python idioms. The `sum(range(...))` pattern was already optimized; extending to `sum(x for x in range(...))` makes the optimization apply to both styles.

## 🧱 Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

## ✅ PR Checklist

- [x] **Correct base branch selected**: `develop`
- [x] **Formatting**: Ran `./bin/autoformat-nuitka-source`
- [ ] **Tests**: All tests still pass.
  - [x] Correctness verified on 1/2/3-arg range for all three builtins
  - [x] Construct benchmarks added: `tests/benchmarks/constructs/BuiltinAnyWithRange.py`, `BuiltinAllWithRange.py`, `BuiltinSumWithRange2.py`, `BuiltinSumWithRange3.py`
  - [x] Crosshair contracts verify identity equivalence symbolically
  - [ ] Optimization XML tests (pending)
- [x] **New Coverage**: Identity genexpr fusion + any/all range fusion covered by construct benchmarks
- [x] **Documentation**: Inline comments document AST structure assumptions

### `_isIdentityGenexprBody()` (BuiltinSumNodes.py)

Checks that a generator body follows the identity pattern `yield x` where `x` is the loop variable. After Nuitka's early optimization passes, the generator body AST is:

```
StatementsSequence
  STATEMENT_TRY (outer cleanup)
    tried: STATEMENTS_FRAME_GENERATOR
      STATEMENT_TRY (StopIteration handler)
        tried: STATEMENT_LOOP
          loop_body: [try/next(.0), assign-to-var, yield var]
```

The function navigates this structure and returns the loop variable if the yield target matches the assigned loop variable.

### `_tryExtractRangeFromGenexprOutline()` (BuiltinSumNodes.py)

Navigates from an `ExpressionOutlineBody` — finds the `.0 = BuiltinIter1(Xrange1/2/3)` assignment and the `MakeGeneratorObject` reference, then calls `_isIdentityGenexprBody()`. If the generator is identity, extracts the raw Xrange node.

### `any(range(n))` / `all(range(n))` (BuiltinAnyNodes.py / BuiltinAllNodes.py)

- `any(range(n))`: range has a non-zero element iff `n >= 2` (range(0)=[], range(1)=[0], range(2)=[0,1])
- `all(range(n))`: all elements are truthy iff `n <= 0` (empty range → True)
- Handling in `computeExpression` via `RICH_COMPARE_GE/LE`

### Benchmark Results (compiled, aarch64 macOS, n=100000, 10000 reps)

| Variant | Time |
|---|---|
| `sum(range(n))` | 0.03 µs |
| `sum(x for x in range(n))` | 0.03 µs |
| `sum(range(start,stop))` | 0.03 µs |
| `sum(x for x in range(start,stop))` | 0.03 µs |
| `sum(range(start,stop,step))` | 0.03 µs |
| `sum(x for x in range(start,stop,step))` | 0.03 µs |

All six forms are indistinguishable — the genexpr identity fusion produces the same `BUILTIN_SUM_RANGE*` helpers as the plain range fusion.

## 🤖 AI Generated Code Policy

  - [x] **Issue First**: [Link to issue]
  - [x] **Verification**: Manually verified by compiling and running test binaries producing correct output for all range variants
  - [x] **Evidence**: Construct benchmarks + compiled C output shows `BUILTIN_SUM_RANGE*` with zero `MAKE_GENERATOR` calls

## Summary by Sourcery

Fuse common sum/any/all patterns over range into specialized arithmetic implementations and comparisons to eliminate iterator overhead while preserving CPython semantics.

New Features:
- Add fusion of sum(x for x in range(...)) into closed-form arithmetic using specialized sum(range) helper nodes and C APIs.
- Introduce builtin helper functions BUILTIN_SUM_RANGE1/2/3 and corresponding code generation for summing ranges without allocating range objects or iterators.
- Add optimizer transformations for all(range(n)) and any(range(n)) over 1-arg ranges into simple integer comparison expressions.

Enhancements:
- Extend builtin sum node analysis to detect outline generator expressions over range that are identity-yielding and lower them to the existing sum(range) fast paths.
- Leverage compile-time constants in new sum(range(...)) nodes to perform constant folding where possible while deferring to runtime helpers otherwise.

Tests:
- Add benchmark-style construct tests for sum(range(n)), sum(range(start, stop)), and sum(range(start, stop, step)) including identity generator-expression variants to validate fusion and performance.
- Add benchmark tests for any(range(n)) and all(range(n)) to verify the new comparison-based fusions and measure their runtime behavior.


KRRT7:
Screenshot 2026-05-21 at 5 21 27 AM
it'd be great if the tooling finally migrated to the 21st century.

kayhayen:
This is an interesting PoC indeed. I think this needs a generic matching engine driven by rules. I have adapted as discussed the planning documents to watch out for this. But I understand we are not going to use this code for real.

KRRT7:
actually I did want it to be used for real, but if it's not wanted, then cool.

kayhayen:
To be usable for real it lacks design. Which you would know if you had not have your AI fill in "Issue First: [Link to issue]" checked.

The actual design that would make this workable while scaling to many different optimization, has been discussed elsewhere already. I might open an issue for tracking it eventually or better yet, put it on the roadmap. In my mind it's always been there, but I couldn't implement it, but probably it's become very doable now.

Issue #3905, opened 2026-05-27
Type variables are missing the `__bound__` attribute


ZeroIntensity:
Discussed on Discord. In short:

```py
def a[T: str]():
    # CPython: <class 'str'>
    # Nuitka: None
    print(T.__bound__)
```

PR #3906, opened 2026-05-27
Implement `__bound__` for type variables

ZeroIntensity:
# Description

## ❓ What does this PR do?

Adds the proper value for the `__bound__` attribute in type variables.

- #3905 

## 🧱 Type of Change

  expected)
- [ ] 🏗️ Build / CI System
- [ ] 📚 Documentation Update

## ✅ PR Checklist

## 🤖 AI Generated Code Policy

## Summary by Sourcery

Implement support for type variable bounds so compiled generics correctly expose the __bound__ attribute.

Bug Fixes:
- Ensure ExpressionTypeVariable nodes carry and evaluate an optional bound so TypeVar.__bound__ is correctly populated at runtime.
- Fix handling of TypeVar AST nodes so their bound expressions are built and evaluated consistently with other type expressions.

Enhancements:
- Introduce a dedicated ExpressionTypeVariable base mixin with an optional bound child to support deferred evaluation and analysis of bounds.
- Split type variable code generation so plain TypeVar and exotic type constructs (TypeVarTuple, ParamSpec) use distinct helpers, clarifying responsibilities of each.
- Extend the C helper MAKE_TYPE_VAR to accept and store an optional bound factory callable for later evaluation.

Tests:
- Add generic function and class tests that exercise TypeVar bounds, including deferred bound evaluation with side effects, and verify the correct __bound__ values.


nuitka/code_generation/TypeAliasCodes.py
```diff
@@ -52,13 +52,45 @@ def generateTypeAliasCode(to_name, expression, emit, context):
 
 
 def generateTypeVarCode(to_name, expression, emit, context):
+    if expression.subnode_bound is not None:
+        compute_bound_name = generateChildExpressionCode(
+            expression=expression.subnode_bound,
+            emit=emit,
+            context=context,
+        )
+    else:
```
kayhayen: Sourcery is correct here, use "None" for the name and if a "if is not None else "NULL" in code template only.

nuitka/nodes/ExpressionBasesGenerated.py
```diff
@@ -738,6 +736,140 @@ def collectVariableAccesses(self, emit_variable):
...
+    def computeExpressionRaw(self, trace_collection):
+        """Compute an expression.
+
+        Default behavior is to just visit the child expressions first, and
+        then the node "computeExpression". For a few cases this needs to
+        be overloaded, e.g. conditional expressions.
+        """
+
+        # First apply the sub-expression, as they it's evaluated before.
+        expression = self.subnode_bound
```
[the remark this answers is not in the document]
kayhayen: The computation does that itself

tests/basics/Generics312.py
```diff
@@ -120,6 +120,30 @@ async def asyncGenTypeParams[T, *Ts, **P]():
 print(asyncFunctionTypeParams, asyncFunctionTypeParams.__type_params__)
 print(asyncGenTypeParams, asyncGenTypeParams.__type_params__)
 
+
+def sideEffect():
+    print("Side effect")
+    return 42
+
+
+def withBound[T: str, UnionT: str | int, Deferred: sideEffect()]():
```
kayhayen: We have a patter of context managers to use "def with" and for tests, something like "boundUsing" would be more appropiate.

tests/basics/Generics312.py
```diff
@@ -120,6 +120,30 @@ async def asyncGenTypeParams[T, *Ts, **P]():
...
+    print("Deferred.__bound__", Deferred.__bound__)
+
+
+print("Function with bound")
+withBound()
+
+print("Class with bound")
+
+
+class WithBound[T: str, UnionT: str | int, Deferred: sideEffect()]:
```
kayhayen: Same as above, I also like to see "BoundUsingClass" and "Function" used, rather than name guessing, just for the tests.

kayhayen (changes requested):
Good job, just a few minor notices.

Also, does `__bound__` really have to be deferred as well?

kayhayen:
Seems sourcery caught the issue of mismatching providers.

nuitka/build/static_src/HelpersTypes.c
```diff
@@ -1044,11 +1044,12 @@ static paramspecobject *_Nuitka_paramspec_alloc(PyThreadState *tstate, PyObject
     return ps;
 }
 
-PyObject *MAKE_TYPE_VAR(PyThreadState *tstate, PyObject *name) {
+PyObject *MAKE_TYPE_VAR(PyThreadState *tstate, PyObject *name, PyObject *bound) {
     // TODO: For Python 3.13 this would work.
     // return _PyIntrinsics_UnaryFunctions[INTRINSIC_TYPEVAR].func(tstate, name);
 
-    return (PyObject *)_Nuitka_typevar_alloc(tstate, name, NULL, NULL, NULL, NULL, false, false, true, NULL);
+    return (PyObject *)_Nuitka_typevar_alloc(tstate, name, NULL, /*evaluate_bound=*/bound, NULL, NULL, false, false,
```
kayhayen, line 1051: I was wondering about this stray inline comment before, is that to document a different name in CPython code?

nuitka/code_generation/TypeAliasCodes.py
kayhayen: This still stands, using "NULL" is not good. I prefer code like this to be used: 

```
    (compute_bound_name,) = generateChildExpressionsCode(
        expression=expression,
        emit=emit,
        context=context,
    )
```

This automatically breaks when new child nodes are added and is a simple pattern.

nuitka/code_generation/TypeAliasCodes.py
```diff
@@ -52,13 +52,48 @@ def generateTypeAliasCode(to_name, expression, emit, context):
...
+        emit(
+            "%s = MAKE_TYPE_VAR(tstate, %s, %s);"
+            % (
+                value_name,
+                context.getConstantCode(constant=expression.name),
+                compute_bound_name,
+            )
+        )
+
+        release_names = (
```
kayhayen: This then becomes easy, you can just pass the child value names and be good.

kayhayen (changes requested):
Looks good to me. This looks largely correct, but a few issues still, specifically the NULL manual stuff in the code names are not the right pattern. Good job.

nuitka/build/static_src/HelpersTypes.c
ZeroIntensity: No, it just makes it clearer as to what parameter is being used.

ZeroIntensity:
> Also, does __bound__ really have to be deferred as well?

Forgot to respond to this: yes.

nuitka/build/static_src/HelpersTypes.c
kayhayen: Ah, I see, I have split up lines and added parameter names in // style in that situation, but fine indeed.

nuitka/code_generation/TypeAliasCodes.py
ZeroIntensity: Switching to `generateChildExpressionsCode` gave me this weird error:

```
nuitka.Errors.NuitkaNodeDesignError: ('ExpressionFunctionRef', 'Need to provide code generation as well', 'EXPRESSION_FUNCTION_REF')
```

Is it crucial that we use it in this PR?

kayhayen: That's befudling to me. I am checking the code out, but that design error is grave and means something is very wrong.

kayhayen: It's also not true, works for me. Maybe you were talking about another place, if so let me know. I pushed for this function just now.

ZeroIntensity: I must have just messed something up then.

H0I00:
Thanks for working on this. The deferred-bound direction looks right to me.

One thing I wondered about while comparing this with a local patch is the constrained TypeVar case. In Python 3.12, `T: (str, bytes)` is represented in the AST through `TypeVar.bound`, but at runtime it is not exposed as a bound. CPython gives `T.__bound__ is None` and `T.__constraints__ == (str, bytes)`.

Would it make sense to include that split in this PR as well?

In my local patch I handled it by giving `ExpressionTypeVariable` two optional children, `bound` and `constraints`, and extending `MAKE_TYPE_VAR` to pass both `evaluate_bound` and `evaluate_constraints` into `_Nuitka_typevar_alloc`, which already has both fields.

I also added tests for function/class/type-alias type parameters, unbounded TypeVars, custom bounds, `list[int]`, tuple constraints, lazy side effects, and cached lazy evaluation. That seems closer to CPython's PEP 695 runtime behavior, but I may be missing an intended scope limitation for this PR.

ZeroIntensity:
The linked PR looks AI-generated. I'm not so sure that an extra subnode is the best option; as far as I can tell, `__contraints__` is mutually exclusive with `__bound__`, so we could just detect whether `bound` is a tuple during code generation and send it to the correct parameter accordingly.

Let's not touch `__constraints__` for now. I need to research how it works a little better.

PR #3926, opened 2026-06-09
Use setup-uv in testing workflow

KRRT7:
## Summary

- Testing workflow uses `actions/setup-python@v5` to select the requested Python version and `astral-sh/setup-uv` to provide `uv`.
- Dependency installation uses `uv pip` as the pip-compatible install layer.
- PyLint version detection reads `pylint.__version__` when available and falls back to `python -m pylint --version`.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -48,17 +48,8 @@ jobs:
             requirements-devel.txt
             requirements.txt
 
-      - name: 📦 Cache apt packages
-        uses: actions/cache@v4
-        with:
-          path: /var/cache/apt/archives
-          key: apt-${{ runner.os }}-patchelf-gdb-ccache-libfuse2
-          restore-keys: apt-${{ runner.os }}-
-
       - name: 🧳 Install Nuitka and dependencies
         run: |
-          sudo apt-get update
-          sudo apt-get install patchelf gdb ccache libfuse2
           python -m pip install --no-python-version-warning --disable-pip-version-check -r requirements-devel.txt
           python -m pip install --no-python-version-warning --disable-pip-version-check .
 
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -63,6 +63,7 @@ jobs:
         run: |
           python ./bin/check-nuitka-with-pylint
 
+
   check-format:
     runs-on: ubuntu-24.04
     name: Check / Format & Codegen
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -35,7 +35,6 @@ jobs:
           with: |
             fetch-depth: 1
             fetch-tags: false
-            filter: blob:none
           attempt_limit: 3
           attempt_delay: 120000
 
@@ -111,7 +110,6 @@ jobs:
           with: |
             fetch-depth: 1
             fetch-tags: false
-            filter: blob:none
           attempt_limit: 3
           attempt_delay: 120000
 
@@ -212,7 +210,6 @@ jobs:
           with: |
             fetch-depth: 1
             fetch-tags: false
-            filter: blob:none
           attempt_limit: 3
           attempt_delay: 120000
 
@@ -338,7 +335,6 @@ jobs:
           with: |
             fetch-depth: 1
             fetch-tags: false
-            filter: blob:none
           attempt_limit: 3
           attempt_delay: 120000
 
@@ -409,7 +405,6 @@ jobs:
           with: |
             fetch-depth: 1
             fetch-tags: false
-            filter: blob:none
           attempt_limit: 3
           attempt_delay: 120000
 
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -59,6 +59,13 @@ jobs:
           key: nuitka-downloads-${{ runner.os }}-${{ runner.arch }}
           restore-keys: nuitka-downloads-${{ runner.os }}-
 
+      - name: 🔧 Cache PyLint results
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/pylint
+          key: pylint-${{ runner.os }}-${{ hashFiles('requirements-devel.txt') }}
+          restore-keys: pylint-${{ runner.os }}-
+
       - name: 🔎 PyLint on Nuitka source code
         run: |
           python ./bin/check-nuitka-with-pylint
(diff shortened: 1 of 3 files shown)


kayhayen:
The pylint change is ineffective:

```
 P hayen@Palmas: ~/repos/Py2C$ time ./bin/check-nuitka-with-pylint
Using concrete python 3.12.3 on x86_64
Using PyLint version: 4.0.5
************* Module nuitka.code_generation.SpecialConstantData
nuitka/code_generation/SpecialConstantData.py:26 R6301 no-self-use SpecialConstantBase.getConstantDetails Method could be a function

real	1m10.708s
user	1m11.780s
sys	0m1.213s
 P hayen@Palmas: ~/repos/Py2C$ time ./bin/check-nuitka-with-pylint
Using concrete python 3.12.3 on x86_64
Using PyLint version: 4.0.5
************* Module nuitka.code_generation.SpecialConstantData
nuitka/code_generation/SpecialConstantData.py:26 R6301 no-self-use SpecialConstantBase.getConstantDetails Method could be a function

real	1m12.355s
user	1m14.290s
sys	0m1.141s
 P hayen@Palmas: ~/repos/Py2C$ time ./bin/check-nuitka-with-pylint
Using concrete python 3.12.3 on x86_64
Using PyLint version: 4.0.5

real	1m8.440s
user	1m10.422s
sys	0m1.078s
 P hayen@Palmas: ~/repos/Py2C$ time ./bin/check-nuitka-with-pylint
Using concrete python 3.12.3 on x86_64
Using PyLint version: 4.0.5

real	1m11.986s
user	1m13.927s
sys	0m1.131s
```

I won't tell you which of these runs are with a the `persistent=yes` on. For all I know, it can only be slower and cause crashes after refactorings.

That said, it's terribly slow indeed.

KRRT7:
@KRRT7 double check your cache is warm otherwise you probably won't see the effects, I had to trigger the PR a few times, if it's ineffective I'll revert

KRRT7:
dunno why it tagged me? I meant to tag you @kayhayen

kayhayen:
The point of fail-fast is to save useless minutes on GitHub, I won't be wasteful on purpose. I accept that overall build time is increased, by not running macOS and Windows builds on what's know to be not good anyway. Since GitHub gives us these resources for free, I don't feel entitled to do it otherwise.

What I wonder is if it would be possible to have some sort of assessment, which tests make sense running anyway, most of the time the other OS tests don't make sense, many times more than hello world style compilation neither, e.g. in case of only Yaml configuration changes.

KRRT7:
if you were to adopt more modern tooling like uv, we could simplify this quite heavily, I can demo this for you if you'd like.

it'd still use the GitHub action Python and everything else you want.

kayhayen:
Regarding the ccache changes, I am wondering, if within a compilation, where largely only "__main__" changes, is small, and other larger files are supposed to be cached already, if it really makes that much of a difference. We are not doing big standalone compilations at all there, and cache restoration is quickly just a burden in that case.

kayhayen:
As for using `uv`, the actions are kind of my only testing of GitHub Actions Python, but since it needs no installation, how could anything that needs a download be faster? Also we are not using `pip` at all outside of the distutils tests, where however, uv is not used a lot, and where poetry is to be used, well it's got to be used. Maybe there a bit of time could be saved by making Virtualenv work with it. I recall doing some work for that, not sure why anymore, maybe it is even used.

kayhayen:
The docs of PyLint suggest that it is not the case. I was using it thrice actually. I think that's just AI hallucination to think it's faster, when all it does is extra work. I specifically added it to be faster once upon a time. Docs say "Pickle collected data for later comparisons." https://docs.pylint.org/features.html

That was a waste of time, outside of realizing of course how slow it is on even a powerful machine.

Running more "checks" in paralell indeed wouldn't harm.

KRRT7:
can you tell me why you're using pylint for exactly?

KRRT7:
> how could anything that needs a download be faster? 

the uv pip compat layer is more performant whilst maintaining the same behavior.

> Since GitHub gives us these resources for free

from what I hear, that might change soon for projects like yours, either way, you shouldn't be wasteful of whatever they do give you for free and try to use the resources efficiently.

KRRT7:
@kayhayen I think I'll just leave it like this for now:

uv here is only for the install layer. It keeps actions/setup-python@v5 as the source of the Python version and uses uv pip as the pip-compatible installer, pip is the bottleneck here.

the pylint change removes a subprocess call that's not needed, reverted everything else

nuitka/tools/quality/pylint/PyLint.py
```diff
@@ -29,15 +29,24 @@ def checkVersion():
...
-        _pylint_version = _pylint_version.split("\n")[0].split()[-1].strip(",")
+        try:
+            import pylint
+        except ImportError:
+            _pylint_version = None
+        else:
+            _pylint_version = getattr(pylint, "__version__", None)
+
+        if _pylint_version is None:
+            with getNullOutput() as null_output:
```
kayhayen, line 40: This looks wrong, testing with the running version for one thing, and then using the other when there is nothing that will make sure it's the one that will be invoked.

.github/workflows/testing.yml
```diff
@@ -163,12 +169,17 @@ jobs:
...
+        with:
+          enable-cache: true
+
       - name: 🧳 Install Nuitka and dependencies
         run: |
           sudo apt-get update
           sudo apt-get install patchelf gdb ccache libfuse2
-          python -m pip install --no-python-version-warning --disable-pip-version-check -r requirements-devel.txt
-          python -m pip install --no-python-version-warning --disable-pip-version-check .
+          uv pip install --system -r requirements-devel.txt .
```
kayhayen, line 181: Actually, the private pip space could try and use ´uv` if available, but I consider that it's a "mistake" to install all things. At this point, most workflows use barely anything from it, and things listed there are nice to have, but no way essential. I can say with some certainty that Nuitka's private pip space, despite using pip will be faster, because it only installs the necessary things. Installing zig is terrible.

I will remove it, and see what surfaces in terms of errors. We strive to have Nuitka runnable out of the box with no environment changes, so why would we do this here, when we don't really want people to have to do it.

kayhayen (changes requested):
The pylint change makes no sense, and the uv change inspired me to drop that part. Not sure if I want to take on private pip space via "uv", because to be honest, I am glad once it becomes robust finally. Installing requirements from a requirement set lazily is nothing that is well supported by any tool really, but that's not what would cause a difference.

kayhayen:
I suffered through making downloads into private pip space the norm, making this very unnecessary.

PR #3931, opened 2026-06-15
Fix inefficient .index() call in dead statement elimination 

KRRT7:
## Summary by Sourcery

Enhancements:
- Replace a repeated index lookup in the dead statement elimination loop with direct use of the current position to improve performance.


kayhayen:
Lack of evidence this is even remote relevant enough to deterioate readability.

KRRT7:
@kayhayen maybe you should read over the code it touches fully and realize it's also a correctness issue.

PR #3933, opened 2026-06-15
iterative tree traversal in ops

KRRT7:
iterative tree traversal

## Summary by Sourcery

Replace recursive tree traversal with an iterative stack-based traversal in the visitor.

Bug Fixes:
- Strengthen validation for 'None' child nodes during tree traversal by checking each child node before visiting.

Enhancements:
- Use an explicit stack to traverse the tree iteratively instead of relying on recursion, preserving enter/leave node callbacks and improving robustness for deep trees.

### DIFF
--- a/nuitka/tree/Operations.py
+++ b/nuitka/tree/Operations.py
@@ -10,15 +10,19 @@ You can visit a scope, a tree (module), or every scope of a tree (module).
 
 
 def visitTree(tree, visitor):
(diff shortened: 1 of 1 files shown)


kayhayen:
Lack of evidence this is even remote relevant enough to deterioate readability. And for this one, I would even keep the old code, if that was the new readability unless it's massive, which it's not.

PR #3943, opened 2026-06-18
Cache normalized import directory entries

KRRT7:
Follow-up to https://github.com/Nuitka/Nuitka/pull/3942.

This adds a normalized-name cache for directory entries in `nuitka/importing/Importing.py` so case-insensitive module resolution can do constant-time lookups instead of rescanning the directory entries for each candidate.

## Summary by Sourcery

Cache normalized directory entries for case-insensitive module resolution and streamline parameter validation logic.

Enhancements:
- Add a cached, normalized directory listing to speed up case-insensitive module candidate selection during import resolution.
- Optimize function parameter validation by avoiding repeated linear scans for duplicate argument names and keyword argument indices.

### DIFF
--- a/nuitka/importing/Importing.py
+++ b/nuitka/importing/Importing.py
@@ -600,6 +600,7 @@ ImportScanFinding = collections.namedtuple(
 )
 
 _list_dir_cache = {}
+_normalized_list_dir_cache = {}
 
 
 def listDirCached(path):
@@ -611,6 +612,20 @@ def listDirCached(path):
     return _list_dir_cache[path]
 
 
+def normalizedListDirCached(path):
+    """Cached normalized listing of a directory."""
+
+    if path not in _normalized_list_dir_cache:
+        normalized_entries = {}
+
+        for fullname, _filename in listDirCached(path):
+            normalized_entries[os.path.normcase(fullname)] = fullname
+
+        _normalized_list_dir_cache[path] = normalized_entries
+
+    return _normalized_list_dir_cache[path]
+
+
 def flushImportCache():
     """Clear import related caches.
 
@@ -619,6 +634,7 @@ def flushImportCache():
     as it throws away so much.
     """
     _list_dir_cache.clear()
+    _normalized_list_dir_cache.clear()
     module_search_cache.clear()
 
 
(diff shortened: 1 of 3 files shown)


ZeroIntensity (commented):
Do you have benchmarks to back this up? I'm concerned that the overhead of comprehensions and construction will outweigh the speedup.

KRRT7:
@ZeroIntensity this is the script used to benchmark:

```python
from __future__ import print_function

import os
import shutil
import sys
import tempfile
from time import perf_counter

sys.path.insert(
    0,
    os.path.normpath(
        os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "..")
    ),
)

from nuitka.importing import Importing as ImportingModule
from nuitka.importing.Importing import flushImportCache, listDirCached

try:
    from nuitka.importing.Importing import normalizedListDirCached
except ImportError:
    normalizedListDirCached = None


LOOPS = 50000


def createBenchmarkTree():
    temp_dir = tempfile.mkdtemp()

    for index in range(128):
        with open(os.path.join(temp_dir, "distractor_%03d.py" % index), "w") as output:
            output.write("")

    return temp_dir


def _oldListDirMatch(search_path_entry):
    for fullname, _filename in listDirCached(search_path_entry):
        if fullname == os.path.join(search_path_entry, "distractor_000.py"):
            return fullname

    return None


def main():
    search_path_entry = createBenchmarkTree()
    original_normcase = ImportingModule.os.path.normcase

    def fakeNormcase(path):
        return path.lower()

    ImportingModule.os.path.normcase = fakeNormcase
    flushImportCache()

    if normalizedListDirCached is not None:
        normalizedListDirCached(search_path_entry)

    print("START:")
    start_time = perf_counter()

    try:
        for _index in range(LOOPS):
            if normalizedListDirCached is not None:
                result = normalizedListDirCached(search_path_entry)
                assert result
            else:
                result = _oldListDirMatch(search_path_entry)
                assert result is not None
    finally:
        ImportingModule.os.path.normcase = original_normcase
        flushImportCache()
        shutil.rmtree(search_path_entry)

    print(perf_counter() - start_time)
    print("STOP:")


if __name__ == "__main__":
    main()
```

KRRT7:
Also, this benchmarking work ended up finding a correctness issue in the original cache implementation on case-sensitive filesystems with case-insensitive normalization, so the final implementation preserves all exact-case variants per normalized key.

kayhayen (commented):
This is AI slop unfortunately, just consider the test changes.

kayhayen:
No issue was created, AI usage does not follow the lenient rules set up. I want an issue first to even know what something aims at achieving. I too am not convinced this is actually achieving real life performance improvements.

KRRT7:
> No issue was created, AI usage does not follow the lenient rules set up

AI didn't follow the "lenient rules" set up because an AI wasn't working on this. I was.

see: https://github.com/Nuitka/Nuitka/issues/3944

PR #3962, opened 2026-07-10
UI: Python flag values and verbosity levels for "trace_imports"

### LINKED ISSUE #2159: FATAL: Unsupported python flag '-vvv'.
* Please check whether the bug was already reported or fixed.

It's not.

* Please check out if the develop version of Nuitka works better for you.

Same thing.

* Nuitka version, full Python version, flavor, OS, etc. as output by *this*
  command.

```
1.5.5
Commercial: None
Python: 3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)]
Flavor: Unknown
Executable: C:\Users\bers\.pyenv-win-venv\envs\project_3.11\Scripts\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 10
Version C compiler: C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.35.32215\bin\HostX64\x64\cl.exe (cl 14.3).
```

* How did you install Nuitka and Python

pip

* Also supply a Short, Self Contained, Correct, Example

`nuitka anyfile.py --python-flag=-vvv`

Note that `python -vvv` is valid.

NevermindNilas:
# Description

## ❓ What does this PR do?

Makes `--python-flag` values carry an optional integer value, and uses that to support the repeated verbose forms (`-vv`, `-vvv`) with an actual verbosity level, rather than collapsing them onto a single boolean.

Following the review feedback, this no longer just silences the error:

* Flag values are now generic: `--python-flag=name=value`, e.g. `--python-flag=trace_imports=2`. A flag given without a value has the value `1`, a flag not given at all has the value `0`. `_getPythonFlags()` therefore returns a mapping of flag name to integer value, and `_getPythonFlagValue()` always returns an integer, so this can be reused for other flags that want levels.

* `-v` may be repeated like with the standard Python executable, and the number of `v` characters is the value, i.e. `-vv` is the same as `trace_imports=2`.

* `python_sysflag_verbose` became a number and is passed through as such, so `SYSFLAG_VERBOSE` and therefore `Py_VerboseFlag` receive the actual level. The Python importer, which is used for some things, honors the higher levels, so the behavior is now compatible with what standard Python does, and not just an accepted-and-ignored flag.

```
$ nuitka anyfile.py --python-flag=-vv     # sys.flags.verbose == 2
$ nuitka anyfile.py --python-flag=trace_imports=2  # same thing
```

Booleans helpers like `hasPythonFlagTraceImports()` remain, they are now `value > 0`, and `getPythonFlagTraceImportsValue()` gives the level.

Fixes #2159.

## 🧱 Type of Change

- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🏗️ Build / CI System
- [ ] 📚 Documentation Update

## ✅ PR Checklist

- [x] **Correct base branch selected**: `develop`.
- [x] **Formatting**: Ran `./bin/autoformat-nuitka-source` (no changes needed) and `./bin/check-nuitka-with-pylint` (clean).
- [x] **Tests**: Verified by compilation and by running the result, see evidence below.
- [ ] **New Coverage**: n/a.
- [x] **Documentation**: `--python-flag` help text now mentions `-v` / `trace_imports` and the value form.

## 🤖 AI Generated Code Policy

  - [x] **Issue First**: Issue #2159 already existed (labelled `help wanted`).
  - [x] **Documentation**: Prompts used (paraphrased):
    - "Look at open issues, check for easy code fixes, reproduce them and if possible fix them."
    - "Look for other issues you can easily fix; it shouldn't have been attempted already."
    - "Widen the search for something small and untaken."
    - "Look at the review comments on the PR and address them."
  - [x] **Verification**: Manually verified, see evidence.
  - [x] **Evidence**:

    Option parsing:

    ```
    -v                -> {'trace_imports': 1}
    -vv               -> {'trace_imports': 2}
    -vvv              -> {'trace_imports': 3}
    trace_imports     -> {'trace_imports': 1}
    trace_imports=2   -> {'trace_imports': 2}
    -vx               -> FATAL: Unsupported python flag '-vx'.
    trace_imports=x   -> FATAL: Python flag 'trace_imports' needs an integer value, not 'x'.
    ```

    Compiling `print(sys.flags.verbose)` with `--python-flag=-vv` gives `#define SYSFLAG_VERBOSE 2` in `build_definitions.h`, and the compiled binary prints `2`, with the import tracing of the Python importer at that level, matching uncompiled `python -vv`.

### DIFF
--- a/nuitka/MainControl.py
+++ b/nuitka/MainControl.py
@@ -73,6 +73,7 @@ from nuitka.options.Options import (
     getOutputDir,
     getPgoArgs,
     getPositionalArgs,
+    getPythonFlagTraceImportsValue,
     getPythonPgoInput,
     getShallFollowExtra,
     getShallFollowExtraFilePatterns,
@@ -834,7 +835,7 @@ def runSconsBackend():
         scons_options["python_sysflag_no_site"] = asBoolStr(True)
 
     if hasPythonFlagTraceImports():
-        scons_options["python_sysflag_verbose"] = asBoolStr(True)
+        scons_options["python_sysflag_verbose"] = str(getPythonFlagTraceImportsValue())
 
     if hasPythonFlagNoRandomization():
         scons_options["python_sysflag_no_randomization"] = asBoolStr(True)
--- a/nuitka/build/Backend.scons
+++ b/nuitka/build/Backend.scons
@@ -105,7 +105,7 @@ python_sysflag_bytes_warning = getArgumentBool("python_sysflag_bytes_warning", F
 # python_sysflag_no_site
 python_sysflag_no_site = getArgumentBool("python_sysflag_no_site", False)
 # python_sysflag_verbose
-python_sysflag_verbose = getArgumentBool("python_sysflag_verbose", False)
+python_sysflag_verbose = getArgumentInt("python_sysflag_verbose", 0)
 # python_sysflag_unicode
 python_sysflag_unicode = getArgumentBool("python_sysflag_unicode", False)
 # python_sysflag_utf8
@@ -446,7 +446,7 @@ def createBuildDefinitionsFile():
 
     env.build_definitions["SYSFLAG_NO_SITE"] = 1 if python_sysflag_no_site else 0
 
-    env.build_definitions["SYSFLAG_VERBOSE"] = 1 if python_sysflag_verbose else 0
+    env.build_definitions["SYSFLAG_VERBOSE"] = python_sysflag_verbose
 
     env.build_definitions["SYSFLAG_UTF8"] = 1 if python_sysflag_utf8 else 0
 
--- a/nuitka/options/OptionParsing.py
+++ b/nuitka/options/OptionParsing.py
@@ -187,9 +187,11 @@ enforces a specific mode. These are options that also exist to standard
 Python executable. Currently supported: "-S" (alias "no_site"),
 "static_hashes" (do not use hash randomization), "no_warnings" (do not
 give Python run time warnings), "-O" (alias "no_asserts"), "no_docstrings"
-(do not use doc strings), "-u" (alias "unbuffered"), "isolated" (do not
-load outside code), "-P" (alias "safe_path", do not used current directory
-in module search) and "-m" (package mode, compile as "package.__main__").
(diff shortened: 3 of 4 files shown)


NevermindNilas:
Addressed the review feedback: instead of hardcoding `-vv`/`-vvv`, the match now normalizes any repeated `-v` sequence (`-v`, `-vv`, `-vvv`, ...) to `trace_imports`, so the comment is accurate and no future edits are needed for more `v`s. Non-`-v` inputs (`-vx`, `-x`, `-`) still error as before. Re-verified compilation for each and ran autoformat + pylint.

kayhayen:
I don't like to add support for flags that are not compatible in behaviour. Actually we do use the Python importer for some things, and maybe "python_sysflag_verbose" should become a number instead of a boolean.

kayhayen (changes requested):
Just silencing the "error" is not a good thing by itself.

kayhayen:
That integer should be passed like so: `--python-flag=trace_imports=2` and then map `-vv` to that. We probably should make that splitting of a value a generic thing, and return an integer always from the function, where not given is 0, in that way we can use it for more things. The default of course is 1 then.

NevermindNilas:
Will take a look a bit later, Ty for the feedback

NevermindNilas:
Reworked as suggested. `--python-flag` now takes an optional integer value generically (`--python-flag=trace_imports=2`), with 1 as the default when a flag is given without a value and 0 when it is not given at all, so `_getPythonFlags()` maps flag names to integers and the accessor always returns an integer, usable for other flags later.

`python_sysflag_verbose` is a number now and is passed through to `SYSFLAG_VERBOSE`/`Py_VerboseFlag`, so `-vv` really is verbosity level 2 and the Python importer honors it, instead of the error being silenced. Verified `-v`/`-vv`/`-vvv`/`trace_imports=2` parsing, invalid values still error, and a compiled binary reports `sys.flags.verbose == 2` for `-vv`. Autoformat and pylint are clean.

PR #3974, opened 2026-07-20
Fix: only defer outer-iterator await for async-for genexprs

### LINKED ISSUE #3973: async for over a generator expression with await in the filter miscompiles: RuntimeError: Task got bad yield
Hey, thanks for Nuitka, it's been great for shipping a Python app onto an ancient ARM Kindle.

Ran into a codegen bug with async generator expressions. When you `async for` over a genexpr that has an `await` in its `if` filter, the compiled binary leaks the yielded element straight into the asyncio Task instead of feeding it back through `__anext__`, and asyncio bails with `Task got bad yield`. The interpreter runs the exact same code fine.

Minimal repro, 8 lines:

```python
import asyncio

async def val(x):
    return x

async def main():
    async for x in (x for x in [1, 2, 3] if (await val(x)) == 2):
        print("got", x)

asyncio.run(main())
```

```
# interpreted
$ python minrepro.py
got 2

# compiled
$ python -m nuitka minrepro.py && ./minrepro.bin
RuntimeError: Task got bad yield: 1
```

The traceback points at the genexpr line, and the "bad yield" value is the genexpr's element (`1`), which is the tell that the async generator's yield is escaping into the Task.

Rewriting it as a plain `for` loop with the `await` in the body compiles and runs correctly, so it's specific to the async generator expression:

```python
for x in [1, 2, 3]:
    if (await val(x)) == 2:
        print("got", x)
```

**Versions / platforms**, reproduces on all of these:
- Nuitka 4.1.3
- Python 3.13.5 (x86_64) and Python 3.11 (armv7, in a Docker build)
- standalone and accelerated mode both affected

Where I hit it in the wild: the [google/bumble](https://github.com/google/bumble) BLE stack has exactly this pattern in its GATT server, so any inbound ATT request crashed the handler once compiled. Writeup and the fix is in this PR: https://github.com/zampierilucas/kindle-hid-passthrough/pull/111, and the plain-loop workaround I'm carrying is here: https://github.com/zampierilucas/kindle-hid-passthrough/blob/1938d9041f70ab6625d465dfd909205e9b91d134/patches/bumble-gatt-server-async-genexpr.patch

Happy to test a fix against the real thing if that helps.

typenoob:
# Description

## ❓ What does this PR do?

Restrict deferred `await` of the outermost iterator in async generator expressions to clauses that are actually `async for`.
In `_buildContractionBodyNode`, Python 3.7+ previously did `YieldFromAwaitable` on the outer iterator for **every** async genexpr (`for_asyncgen`). That is only correct when the outermost clause is `async for` (aiter must be awaited after construction, including when the genexpr is created in a sync `def`).

For sync `for` genexprs that are async only because of an inner `await` (in the element or `if`), the outer value is a normal iterator (e.g. `list_iterator`). Awaiting / yield-from'ing it leaked the element into the driving `Task` and raised:
```text
RuntimeError: Task got bad yield: 1
```

The guard is now also `getattr(qual, "is_async", 0)`, matching the intent of 70549ba and the existing `_makeIteratorCreation` path.

Adds `tests/basics/AsyncGenexprAwaitTest37.py` covering:
- sync `for` + `await` driven via `__anext__` (must not bad-yield)
- `async for` genexpr created in a normal `def` (deferred aiter await must still work)

Fixes #3973

## 🧱 Type of Change

      expected)
- [ ] 🏗️ Build / CI System
- [ ] 📚 Documentation Update

## ✅ PR Checklist

## 🤖 AI Generated Code Policy

  - [x] **Issue First**: I have created an issue explaining the problem *before* using AI to
  - [x] **Verification**: I have **manually verified** the AI generated code.
  - [x] **Evidence**: I have included test evidence that the change is effective.

## Summary by Sourcery

Fix handling of async generator expressions so that deferred awaiting of the outer iterator only occurs for truly async-for comprehensions, and add coverage for these async genexpr behaviors.

Bug Fixes:
- Prevent async generator expressions with a synchronous outer for and inner await from incorrectly treating the outer iterator as an awaitable, avoiding bad Task yields.
- Correct error message construction in YAML scalar node handling to avoid reliance on deprecated ruamel.yaml internals.

Enhancements:
- Align async generator expression iterator handling with existing iterator creation logic by guarding await behavior on the generator qualifier's async flag.

Tests:
- Add an async genexpr test verifying correct behavior for sync-for with inner await and async-for genexprs created in a synchronous function.

### DIFF
--- a/nuitka/tree/ReformulationContractionExpressions.py
+++ b/nuitka/tree/ReformulationContractionExpressions.py
@@ -463,10 +463,11 @@ def _buildContractionBodyNode(
         if qual is node.generators[0]:
             iterator_ref = makeVariableRefNode(variable=iter_tmp, source_ref=source_ref)
 
-            if for_asyncgen and python_version >= 0x370:
-                iterator_ref = ExpressionYieldFromAwaitable(
-                    expression=iterator_ref, source_ref=source_ref
-                )
+            if getattr(qual, "is_async", 0):
+                if for_asyncgen and python_version >= 0x370:
+                    iterator_ref = ExpressionYieldFromAwaitable(
+                        expression=iterator_ref, source_ref=source_ref
+                    )
 
             tmp_iter_variable = None
 
--- a/tests/basics/AsyncGenexprAwaitTest37.py
+++ b/tests/basics/AsyncGenexprAwaitTest37.py
@@ -1,0 +1,52 @@
+#     Copyright 2026, Kay Hayen, mailto:kay.hayen@gmail.com find license text at end of file
+
+
+"""Async genexpr: sync-for with await vs async-for.
+
+* Sync ``for`` plus ``await`` in the element must not treat the outer
+  list iterator as awaitable (would surface as Task got bad yield).
+* ``async for`` genexpr must still work when created in a normal ``def``,
+  where getting the async iterator is deferred until the asyncgen runs
+  (cannot await at construction time in a sync function).
+"""
+
+import asyncio
+
+
+async def val(x):
+    return x
+
+
+async def arange(n):
+    for i in range(n):
+        yield i
+
+
+def make_arange(n):
+    return (i async for i in arange(n))
+
+
+async def main():
+    agen = (await val(x) for x in [1])
+    print(await agen.__anext__())
+
+    agen2 = make_arange(3)
+    print(await agen2.__anext__())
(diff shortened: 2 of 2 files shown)


kayhayen:
Perfect job, thanks for taking this on.

PR #4002, opened 2026-08-14
Fix, prefer source package init over C extension when recompiling

### LINKED ISSUE #3999: invalid or missing encoding declaration with prefer-source-code option
# Bug Report

## 🐛 Bug Description

Build executable with dependency that come with source and compiled file failed with --prefer-source-code option because nuitka try to process pyd as source file.

```sh
>poetry run python -m nuitka --version
4.1.3
Commercial: None
Python: 3.12.1 (tags/v3.12.1:2305ca5, Dec  7 2023, 22:03:25) [MSC v.1937 64 bit (AMD64)]
Flavor: CPython Official
Executable: C:\Projects\tests\zeroconf-playground\.venv\Scripts\python.exe
OS: Windows
Arch: x86_64
WindowsRelease: 11
Version C compiler: cl (cl 14.3).```


Python was installed from the exe downloaded from python website.
Then nuitka was installed using poetry in the local venv.

```sh
> poetry run python -m pip list -v
Package             Version Editable project location                       Location                                                                Installer
------------------- ------- ----------------------------------------------- ----------------------------------------------------------------------- ------------
ifaddr              0.2.0                                                   C:\Projects\ests\zeroconf-playground\.venv\Lib\site-packages Poetry 2.4.1
Nuitka              4.1.3                                                   C:\Projects\tests\zeroconf-playground\.venv\Lib\site-packages Poetry 2.4.1
pip                 25.3                                                    C:\Projects\tests\zeroconf-playground\.venv\Lib\site-packages pip
zeroconf            0.150.0                                                 C:\Projects\tests\zeroconf-playground\.venv\Lib\site-packages Poetry 2.4.1
zeroconf-playground 0.1.0   C:\Projects\tests\zeroconf-playground C:\Projects\tests\zeroconf-playground\.venv\Lib\site-packages poetry
```

```sh
> poetry run python -m nuitka --list-distribution-metadata
Nuitka 4.1.3 poetry 2.4.1
ifaddr 0.2.0 poetry 2.4.1
pip 25.3 pip
zeroconf 0.150.0 poetry 2.4.1
zeroconf-playground 0.1.0 poetry
```

### 1. "Hello World" Test (if applicable)

I create an empty project like this:
```powershell
poetry new zeroconf-playground
cd zeroconf-playground
poetry add zeroconf
poetry add nuitka -G build
@'
import zeroconf
import zeroconf._services
print(zeroconf.__version__)
'@ | Set-Content src/zeroconf_playground/app.py
poetry run python -m nuitka `
    --standalone `
    --prefer-source-code `
    --output-dir=build-test-source `
    src/zeroconf_playground/app.py
```

Output will be :
`PASS 1:   0.0%|                         | 0/1 modules - __main__SyntaxError: invalid or missing encoding declaration for 'C:\\Projects\\tests\\zeroconf-playground\\.venv\\Lib\\site-packages\\zeroconf\\_services\\__init__.cp312-win_amd64.pyd'`

Provide a minimal code example that reproduces the issue. It should have as few external
dependencies as possible. *(Issues without a clear, runnable example are likely to be rejected. For
packaging issues, a simple import might suffice. Repositories with `pipenv` or archives with small
examples are very welcome, but must contain only source code. Ensure usage is obvious; a compile
script is ideal. For non-commercial users, keep the code brief.)*

```python
import zeroconf
import zeroconf._services
print(zeroconf.__version__)
```

### 3. Nuitka Command Line Options

State the Nuitka options used. 

```powershell
# Your Nuitka command here
poetry run python -m nuitka `
    --standalone `
    --prefer-source-code `
    --output-dir=build-test-source `
    src/zeroconf_playground/app.py
```

## 📉 Expected Behavior

Generation must succeed

## 📄 Actual Behavior & Output

Build failded with log:

```
Nuitka-Options: Used command line options:
Nuitka-Options:   --standalone --prefer-source-code --output-dir=build-test-source src\zeroconf_playground\app.py
Nuitka: Starting Python compilation with:
Nuitka:   Version '4.1.3' on Python 3.12 (flavor 'CPython Official') commercial grade 'not installed'.
PASS 1:   0.0%|                         | 0/1 modules - __main__SyntaxError: invalid or missing encoding declaration for 'C:\\Projects\\tests\\zeroconf-playground\\.venv\\Lib\\site-packages\\zeroconf\\_services\\__init__.cp312-win_amd64.pyd'

Nuitka is very syntax compatible with standard Python. It is currently running
with Python version '3.12', you might want to specify more clearly with the use
of the precise Python interpreter binary and '-m nuitka', e.g. use this
'python3.13 -m nuitka' option, if that's not the one the program expects.
```

## ↩️ Regression (if applicable)

I tested with the version 2.8.10 and it work with this old version

typenoob:
## ❓ What does this PR do?

Fixes an issue where `getPackageDirFilename()` would incorrectly select the compiled `__init__.so` instead of the source `__init__.py` when both exist in a package directory and extension module recompilation is enabled.

- Added `prefer_source` parameter to `getPackageDirFilename()`, which skips `C_EXTENSION` type suffixes when set to `True`
- Pass `decideRecompileExtensionModules(module_name)` as the value when calling from `decideModuleSourceRef()`

## 🧐 Why was it initiated? Any Relevant Issues?

Close #3999

When a package has both `.py` source and compiled `.so`/`.pyd` extension modules, and `--recompile-extension-modules` is enabled, Nuitka should prefer the source file for compilation rather than directly using the compiled C extension as the package's `__init__` entry point.

## 🧱 Type of Change

- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 🧹 Refactoring
- [ ] 🏗️ Build / CI System
- [ ] 📚 Documentation Update

## ✅ PR Checklist

- [ ] **Tests**: All tests still pass. (See

## Summary by Sourcery

Bug Fixes:
- Ensure getPackageDirFilename can prefer source __init__.py over C extension __init__.so/__init__.pyd in package directories when requested.

## Summary by Sourcery

Fix package initialization selection so recompilation uses the source initializer instead of an existing compiled extension.

Bug Fixes:
- Prefer a package's Python source initializer over compiled C-extension initializers when extension-module recompilation is enabled.

Enhancements:
- Update package initializer resolution across importing, module construction, scanning, and module-finding paths to use the relevant package context when selecting candidates.


nuitka/utils/Importing.py
```diff
@@ -303,10 +303,12 @@ def hasPackageDirFilename(path):
     return False
 
 
-def getPackageDirFilename(path):
+def getPackageDirFilename(path, prefer_source=False):
     assert os.path.isdir(path)
 
-    for suffix, _module_type in getModuleFilenameSuffixes():
+    for suffix, module_type in getModuleFilenameSuffixes():
```
kayhayen, line 335: That would break, if there is nothing existing, making the decision moot, aber it would stop being a package directory. Instead I guess we need to re-order the `getModuleFilenameSuffixes` applied. Also I don't like the default argument usage, it might be hiding the fact that this has more uses.

kayhayen (changes requested):
Good approach, but I think not yet robust enough, thanks for tackling this one.

nuitka/utils/Importing.py
typenoob: Do you mean re-ordering the suffixes in getModuleFilenameSuffixes() so that `PY_SOURCE` comes before `C_EXTENSION`?

kayhayen: Yes, exactly that is what preference is. Prefering is not preventing the other thing from usage unless the the preferred thing exists. By re-ordering it should magically work fine with "*" matching all extension modules, but not all having source code.

typenoob: Another question: Should all other callers of `getPackageDirFilename` also respect the re-compile decision?

I'm considering passing `module_name` into `getPackageDirFilename`, and have the function determine the re-compile decision internally.

kayhayen: if all users can come by with the module_name, then deciding inside of there would be good, but what would isPackageDir() pass (it doesn't care about precedence). One thing I wondered btw, if getPackageDirFilename is called a lot, and should cache its decision, calls to "os.path.isfile(candidate)" could be costly in terms of locking/file system access. 

So, I think you should get away with indeed passing module_name, but allow for "None" in the I don't care category.

Ignore my concern about caching, I think I will add something like `CachedFileOperations` in the future to address that, it's out of the scope of this kind of change. Just something to keep track of. Doing IO repeatedly on the same things is very much avoid-worthy. And caching results is bad, because here e.g. we might call it with None, and its module name at times.

The decision ought to defalse to False in module_name (package_name probably more like it) is None. And then call the decision on the inside. That avoids pollution of the code. Thanks for pushing towards quality there.

typenoob:
Basically done. I'm just not sure whether the call sites care about the recompilation decision, so you may need to check that.

PR #4042, opened 2026-09-19
Fix temporary variable collisions in cloned comprehension outlines

### LINKED ISSUE #4041: AssertionError when cloning comprehensions in a finally block
# Bug Report

## ⚠️ Pre-Submission Checklist

- [x] **Search Existing Issues:** Searched open and closed issues and PRs for the assertion names and related cloning/finally terms. No matching report found. #847 concerns a node constructor argument error; #3689 concerns trace merging during optimization.
- [x] **Python vs. Nuitka:** The example runs successfully with CPython.
- [x] **Test `develop` Branch:** Reproduced on `957a4006f` (4.3rc3).

## 🐛 Bug Description

A comprehension in a `finally` block can crash Nuitka during tree building:

```text
AssertionError: listcomp_1__.0_clone
```

This happens when the final block is cloned for several exit paths. The example below needs no third-party packages, and the function does not need to be called to trigger the compiler crash.

Reproduced on `develop` at `957a4006ff060feced71f2a60938e4970377d2a4`:

```text
4.3rc3
Update status: unavailable.
Commercial: None
Python: 3.13.5 (main, Aug 10 2026, 12:06:59) [GCC 14.2.0]
Flavor: Debian Python
GIL: yes
Executable: /usr/bin/python3
OS: Linux
Arch: x86_64
Distribution: Debian 13
Version C compiler: gcc (gcc 14).
```

Nuitka from a Git checkout, run directly from source. System Debian Python, no virtual environment.

None; only Python built-ins are used.

### 1. "Hello World" Test (if applicable)

Compiling and importing a module containing `print("hello world")` works with the same checkout and interpreter.

Save as `repro.py`:

```python
def run(flag):
    try:
        if flag:
            return "returned"
        print("body")
    finally:
        values = [x for x in range(3)]
        print("finally", values)


print(run(False))
print(run(True))
```

### 3. Nuitka Command Line Options

With the source checkout on `PYTHONPATH`:

```sh
python3 -m nuitka --mode=module repro.py
```

## 📉 Expected Behavior

Compilation succeeds. Importing the module produces the same output as CPython:

```text
body
finally [0, 1, 2]
None
finally [0, 1, 2]
returned
```

## 📄 Actual Behavior & Output

Compilation fails before C code generation. Traceback excerpt (intermediate frames omitted):

```text
  File ".../nuitka/tree/ReformulationTryFinallyStatements.py", line 186, in makeTryFinallyStatement
    statements=(result, getFinal()), source_ref=source_ref
  File ".../nuitka/tree/ReformulationTryFinallyStatements.py", line 107, in getFinal
    return final.makeClone()
  ...
  File ".../nuitka/nodes/OutlineNodes.py", line 256, in makeClone
    assert new_temp_name not in entry_point.temp_variables, new_temp_name
AssertionError: listcomp_1__.0_clone
```

## ↩️ Regression (if applicable)

Also reproduced on 4.3rc2 (`58e718e95`). I have not established the first affected release. Testing was done with Python 3.13.5.

## 💡 Additional Context (Optional)

Set and dictionary comprehensions also fail with this shape on the previously tested 4.3rc2 checkout, producing `setcontraction_1__.0_clone` and `dictcontraction_1__.0_clone`. Replacing the comprehension with a generator expression compiled successfully there.

`makeTryFinallyStatement` uses the original final block once, then clones it for the remaining exit paths. Here, exception, return, and normal fall-through handling require the original plus two clones. `ExpressionOutlineFunctionBase.makeClone` appends the fixed suffix `_clone` to each temporary variable name, so the second clone of the same original requests a name that is already registered.

Proposed fix: #4042.

sealpp:
# Description

## ❓ What does this PR do?

Fixes the `listcomp_1__.0_clone` assertion when a `finally` block containing a comprehension is cloned more than once.

Keep `_clone` for the first available name, then try `_clone_2`, `_clone_3`, etc. Each outline still owns separate temporary variables. The new basic test covers a final block shared by exception, return, and normal fall-through paths.

Fixes #4041.

## 🧱 Type of Change

- [ ] 🏗️ Build / CI System
- [ ] 📚 Documentation Update

## ✅ PR Checklist

- [x] **Correct base branch selected:** `develop`.
- [x] **Formatting:** `./bin/autoformat-nuitka-source --un-pushed --assume-yes-for-downloads`.
- [x] **Tests:** Focused checks passed on Python 3.13.5:
  - `./bin/check-nuitka-with-pylint --un-pushed`
  - `python3 tests/basics/run_all.py only TryFinallyListContractionTest`
  - Module compilation and CPython output comparison for the minimal example, list/set/dict comprehensions, and the nested async-generator example.
- [x] **New Coverage:** `tests/basics/TryFinallyListContractionTest.py`.
- [ ] **Documentation:** No user-facing behavior change beyond accepting valid code.

## Summary by Sourcery

Ensure cloned comprehension outlines receive unique temporary variable names across repeated cloning.

Bug Fixes:
- Prevent temporary-variable name collisions when cloned comprehension outlines are created multiple times.

Tests:
- Add coverage for comprehensions in try/finally blocks reached through return and normal execution paths.


kayhayen (approved):
Looks very good to me, good job.

kayhayen:
The usuall good stuf. I cleaned it up and added an accessor for checking variable presence, and added tests for set and dict contraction. Not that it matters much, but just as I think it would be good to have those generally.

Issue #1777, opened 2022-09-06
huge packages with long module names problem (sh: Argument list too long)


lepy:
Hi,

huge packages with long module names can not linked correctly.  

## System

* Nuitka version, full Python version, flavor, OS, etc. as output by *this*
  command.

  > python -m nuitka --version

1.0.6
Commercial: None
Python: 3.9.0 (default, Nov 15 2020, 14:28:56) 
Flavor: Anaconda Python
Executable: /home/user/anaconda39/envs/py39/bin/python
OS: Linux
Arch: x86_64
Distribution: Ubuntu (based on Debian) 20.04.4

* How did you install Nuitka and Python

 pip in ananconda env

  > python -m pip freeze

Nuitka==1.0.6

```import os

## Example
### create mymodule

"""
python -m nuitka --module m.py --include-package=mymodule
"""
def create_mymodule(n=2000):
    if not os.path.exists("mymodule"):
        os.mkdir("mymodule")

    with open(os.path.join("mymodule", "__init__.py"), "w") as fh:
        i = 1
        fh.write(f"""
import module{i:0200d}

print(module{i:0200d}.f())
""")

    for i in range(n):
        with open(os.path.join("mymodule", f"module{i:0200d}.py"), "[Shell Command Line Length Limitations](https://www.cyberciti.biz/faq/argument-list-too-long-error-solution/)w") as fh:
            fh.write(f"def f():return f'module{i:0200d}'")



if __name__ == '__main__':
    create_mymodule(2000)
```

### create module `m.py`

```import mymodule

import mymodule.module000001

print(mymodule.module000001.f())
```

### nuitka command

`python -m nuitka --module m.py --include-package=mymodule`

### Error

```$ python -m nuitka --module m.py --include-package=mymodule
Nuitka-Options:INFO: Used command line options: --module m.py --include-package=mymodule
Nuitka:INFO: Starting Python compilation with Nuitka '1.0.6' on Python '3.9' commercial grade 'not installed'.
Nuitka:INFO: Completed Python level compilation and optimization.                                                                                                                                                                       
Nuitka:INFO: Generating source code for C backend compiler.
Nuitka:INFO: Running data composer tool for optimal constant value handling.                                                                                                                                                            
Nuitka:INFO: Running C compilation via Scons.
Nuitka-Scons:INFO: Backend C compiler: gcc (gcc).
Nuitka-Scons:INFO: Backend linking program with 2010 files (no progress information available).
[...]

PR #4053, opened 2026-09-21
Fix: use linker response file for clang gcc_mode builds

### LINKED ISSUE #4039: macOS clang standalone build fails with "Argument list too long" — linker response-file fix from #1777 excludes clang_mode
## 🐛 Bug Description                                                                                                        
                                                                                                                               
  The linker response-file workaround that avoids `Argument list too long`                                                     
  failures on the final link step is skipped for plain clang. It was added                                                     
  for gcc/mingw in #1777 ("huge packages with long module names problem (sh:                                                   
  Argument list too long)"), fixed in Nuitka 1.1, but the fix's condition                                                      
  excludes clang unless `zig_mode` is also set:                                                                                
                                                                                                                               
  `nuitka/build/Backend.scons`, line ~526:                                                                                     
                                                                                                                               
  ```python                                                                                                                    
  # Hack to make Scons use tempfile for gcc linking, to avoid line length limits,                                              
  # which can make linking fail with many modules otherwise. Most needed on Windows,                                           
  # but useful on other platforms too.                                                                                         
  if env.gcc_mode and (not env.clang_mode or env.zig_mode):                                                                    
      makeGccUseLinkerFile(                                                                                                    
          env=env,                                                                                                             
          source_files=source_files,                                                                                           
      ) 
  ```                                                                                                                      
                                                                                                                               
  As a result, a standalone macOS build (Apple clang, not gcc, not zig) with                                                   
  enough object files at the final link step hits the same                                                                     
  "Argument list too long" failure that #1777 was meant to fix.

🖥 Environment                                                                                                               
                                                                                                                               
  1. Nuitka Version, Python Version, OS, and Platform                                                                          
                                                                                                                               
  Nuitka 4.1.2                                                                                                                 
  Python: 3.12.10 (v3.12.10:0cc81280367, Apr 8 2025, 08:46:59) [Clang 13.0.0 (clang-1300.0.29.30)]                             
  Flavor: CPython Official                                                                                                     
  OS: Darwin                                                                                                                   
  Arch: arm64                                                                                                                  
  macOSRelease: 26.3.1                                                                                                         
  Version C compiler: clang (clang 21.0.0).                                                                                    
                                                                                                                               
  2. How Nuitka and Python were Installed                                                                                      
                                                                                                                               
  pip, inside a project virtualenv (Python.org 3.12 build).                                                                    
                                                                                                                               
  🛠 To Reproduce                                                                                                              
                                                                                                                               
  1. "Hello World" Test                                                                                                        
                                                                                                                               
  Not applicable — the failure only manifests once the final link step's                                                       
  object file count is large enough to exceed ARG_MAX on the platform. A                                                       
  trivial "hello world" compiles fine. 

 2. SSCCE                                                                                                                     
                                                                                                                               
  No minimal reproducer — the failure only shows up once the standalone                                                        
  build's object file count is large enough to blow past ARG_MAX on the                                                        
  final executable link (~11k .o/.os files in our real application).                                                           
                                                                                                                               
  3. Nuitka Command Line Options                                                                                               
                                                                                                                               
  python -m nuitka \                                                                                                           
    --standalone \                                                                                                             
    --macos-create-app-bundle \                                                                                                
    --output-dir=dist \                                                                                                        
    --macos-target-arch=arm64 \                                                                                                
    --assume-yes-for-downloads \                                                                                               
    .                                                                                                                          
                                                                                                                               
  (trimmed to the flags relevant to the failure; the full command also                                                         
  includes several --include-package=/--nofollow-import-to= entries for                                                        
  our own application code, omitted here as not relevant to the bug)

📉 Expected Behavior                                                                                                         
                                                                                                                               
  The final backend C link step succeeds regardless of the number of object                                                    
  files, the same way it does for gcc/mingw builds since #1777.                                                                
                                                                                                                               
  📄 Actual Behavior & Output                                                                                                  
                                                                                                                               
  Nuitka-Scons: Backend C linking with 11277 files (no progress information available for this stage).                         
  scons: *** [.../AjaxDesktop.app/Contents/MacOS/AjaxDesktop] /usr/bin/clang: Argument list too long                           
  FATAL: Failed unexpectedly in Scons C backend compilation.                                                                   
                                                                                                                               
  followed by Nuitka exiting non-zero after writing a crash report.                                                            
                                                                                                                               
  ↩️ Regression                                                                                                                
                                                                                                                               
  Not a regression — the clang_mode exclusion in Backend.scons appears to                                                      
  have been there since the response-file mechanism was introduced for                                                         
  #1777 (Nuitka 1.1), so clang was never covered by that fix in the first                                                      
  place.

20000419:
## Summary
The linker response-file workaround from #1777 avoided `Argument list too long` for gcc/mingw, but the condition skipped plain `clang_mode` unless `zig_mode` was also set. macOS clang standalone builds with many objects still hit ARG_MAX (#4039).

Enable `makeGccUseLinkerFile` for all `gcc_mode` toolchains (gcc, mingw, clang, zig).

## Why
Leftover of #1777: clang was never covered by that fix.

## Test plan
- [x] Confirmed single call site in `Backend.scons`
- [x] `makeGccUseLinkerFile` uses `@link_input.txt` (clang/ld accept `@file`)
- [x] 2026-09-21 box smoke (Linux): Backend.scons condition is `if env.gcc_mode:` (clang no longer excluded); makeGccUseLinkerFile smoke + clang @file link of 80 objs OK; `PYTHONPATH=fork python3 -m nuitka --module hello.py --clang` → built hello.so AND wrote hello.build/@link_input.txt (clang 19.1.7)
- [ ] Prefer maintainer/CI clang standalone smoke if available
- [ ] Full macOS ARG_MAX 10k-obj standalone not reproduced here; clang @file path verified on Linux

Fixes #4039

## Summary by Sourcery

Enable linker response files across gcc-mode toolchains to support large clang and other builds without exceeding command-line length limits.

Bug Fixes:
- Use linker response files for all gcc-mode toolchains, including standalone clang builds, preventing failures caused by oversized linker command lines.

Enhancements:
- Refine linker response-file generation to select the appropriate object-file suffix and include only the resolved object filenames.


kayhayen:
I did a bunch of cleanups on top, and checked that zig seems to be happy with it on Linux at least. I will merge into integrating branch, so my CI can prove that zig on Windows is not negatively affected and such.

kayhayen:
I did some more cleanups while reviewing the old code. I tested zig to also accept the linker response file, using gcc in the name feels wrong now. And there was a bit of dead code in there too with $SOURCES never having any contribution. Also I put source_files to source_filenames, something I will eventually propagate to the outside, I like to not have to guess if it's a file or a filename, all old sins of mine.

Very good job thank you for the contribution.

